PR #4, opened 2020-05-06
Coil + Compose integration module

chrisbanes:
This isn't designed to expose everything in Coil via Compose, rather just the simple case. Although this could be expanded later.

/cc @colinrtwhite


colinrtwhite:
This is awesome! Thanks for creating this. 🎉

Also let me know if there's anything that could be improved with Coil's interop with Compose. I'm still a Compose novice 😅 .

coil/src/main/java/dev/chrisbanes/accompanist/coil/Coil.kt
```diff
@@ -0,0 +1,264 @@
...
+        }
+    }
+}
+
+/**
+ * A simple composable which loads an image using [Coil] into a [Box]
+ */
+@Composable
+fun NetworkImage(
+    data: Any,
```
colinrtwhite, line 158: Since this takes `data`, it could technically work with any supported Coil [data type](https://coil-kt.github.io/coil/getting_started/#supported-data-types). What do you think about renaming it to `CoilImage` or just make it an overload of `Image`?

You could also make `data` a `Uri` or `HttpUrl`, however that's more restrictive.

chrisbanes:
No problem! 😃

I think the integration is great, especially after https://github.com/coil-kt/coil/issues/278

As to upstreaming, I'm totally up for that eventually (maybe around beta time). As you say, Compose is still changing APIs so I wouldn't want to move the burden onto you to keep up. I need to do it anyway to keep Tivi up to date, so I think it makes sense to keep it here for now. Sound good?

colinrtwhite:
@chrisbanes :+1: Sounds good

coil/src/main/java/dev/chrisbanes/accompanist/coil/Coil.kt
chrisbanes: > What do you think about renaming it to CoilImage

Yep, done! I personally like the `data: Any` thing, because I use mapped types extensively.

chrisbanes:
PTAL at the new changes. Still needs docs + tests but I think the API is looking good.

coil/src/main/java/dev/chrisbanes/accompanist/coil/Coil.kt
```diff
@@ -0,0 +1,257 @@
...
+                        duration = (crossfadeDuration * 0.75f).roundToInt()
+                    }
+                    saturation using tween {
+                        duration = crossfadeDuration
+                    }
+                }
+            }
+        }
+
+        val requestWidth = when {
```
nickbutcher, line 112: What about making this an ext fun on `Constraints`?

coil/src/main/java/dev/chrisbanes/accompanist/coil/Coil.kt
```diff
@@ -0,0 +1,257 @@
...
+        val result = GetRequest.Builder(ContextAmbient.current)
+                .data(data)
+                .size(requestWidth.value, requestHeight.value)
+                .apply {
+                    if (requestBuilder != null) {
+                        requestBuilder(this)
+                    }
+                }
+                .build()
+                .executeAsComposable()
```
nickbutcher, line 133: Won't this re-execute on every recompose? Should this be guarded by `onActive(requestWidth, requestHeight)`?

coil/src/main/java/dev/chrisbanes/accompanist/coil/Coil.kt
```diff
@@ -0,0 +1,257 @@
...
+    Loaded,
+    Empty
+}
+
+private val alpha = FloatPropKey()
+private val brightness = FloatPropKey()
+private val saturation = FloatPropKey()
+
+/**
+ * A composable which loads an image using [Coil] into a [Box], using a crossfade when first
```
nickbutcher, line 70:
```suggestion
 * A composable which loads an image using [Coil] into an [Image], using a crossfade when first
```

coil/src/main/java/dev/chrisbanes/accompanist/coil/Coil.kt
```diff
@@ -0,0 +1,257 @@
...
+                    contentScale = contentScale,
+                    alignment = alignment
+                )
+            }
+        }
+    }
+}
+
+/**
+ * A simple composable which loads an image using [Coil] into a [Box]
```
nickbutcher, line 163:
```suggestion
 * A simple composable which loads an image using [Coil] into an [Image]
```

coil/src/main/java/dev/chrisbanes/accompanist/coil/Coil.kt
```diff
@@ -0,0 +1,257 @@
...
+    }
+
+    /**
+     * Return the dimension of the underlying [Image] as it's intrinsic width and height
+     */
+    override val intrinsicSize: PxSize get() = size
+}
+
+/**
+ * A simple [loadImage] composable, which loads [data].
```
nickbutcher, line 237: What is [loadImage]?

coil/src/main/java/dev/chrisbanes/accompanist/coil/Coil.kt
nickbutcher: Oh wait, `executeAsComposable` guard this. nm

coil/src/main/java/dev/chrisbanes/accompanist/coil/Coil.kt
```diff
@@ -0,0 +1,257 @@
...
+                val matrix = remember(asset) { ImageLoadingColorMatrix() }
+                matrix.saturationFraction = transitionState[saturation]
+                matrix.alphaFraction = transitionState[alpha]
+                matrix.brightnessFraction = transitionState[brightness]
+
+                // Unfortunately ColorMatrixColorFilter is not mutable so we have to create a new
+                // instance every time
+                val cf = ColorMatrixColorFilter(matrix)
+                Image(
+                    painter = AndroidColorMatrixImagePainter(asset, cf),
```
nickbutcher, line 153: `remember` the painter?

coil/src/main/java/dev/chrisbanes/accompanist/coil/Coil.kt
chrisbanes: Yep, the request will be created every time (there's no way around that while keeping the `requestBuilder` lambda), but the actual execution is guarded. I'll add a comment for this.

coil/src/main/java/dev/chrisbanes/accompanist/coil/Coil.kt
chrisbanes: Yep, need to fix the docs. Done.

coil/src/main/java/dev/chrisbanes/accompanist/coil/Coil.kt
```diff
@@ -0,0 +1,276 @@
...
+ * given bounds defined by the width and height.
+ * @param contentScale Optional scale parameter used to determine the aspect ratio scaling to be
+ * used if the bounds are a different size from the intrinsic size of the loaded [ImageAsset].
+ * @param modifier Modifier used to adjust the layout algorithm or draw decoration content (ex.
+ * background)
+ */
+@Composable
+fun CoilImage(
+    data: Any,
+    requestBuilder: ((GetRequestBuilder) -> Unit)? = null,
```
colinrtwhite, line 189: Instead of having a `requestBuilder` lambda param what do you think about having `CoilImage` variants? One that accepts just `data: Any` for simple use cases and one that accepts a full `request: GetRequest`? Internally, you could use `newBuilder()` to modify the request instance. [Something like this](https://gist.github.com/colinrtwhite/05205d6c7fc0782883f9854c1ca290c3).

coil/src/main/java/dev/chrisbanes/accompanist/coil/Coil.kt
```diff
@@ -0,0 +1,276 @@
...
+@Composable
+fun GetRequest.executeAsComposable(): RequestResult? {
+    var result by stateFor<RequestResult?>(this) { null }
+    val context = ContextAmbient.current
+
+    // Launch and execute a new request when it changes
+    onCommit(this) {
+        val job = CoroutineScope(Dispatchers.Main).launch {
+            // Start loading the image and await the result
+            result = Coil.imageLoader(context).execute(this@executeAsComposable)
```
colinrtwhite, line 267: In the future it might make sense to offer separate singleton and non-singleton variants so users using `coil-base` can pass their own `ImageLoader` instances. Maybe `accompanist-coil-base` to follow the same naming scheme? The singleton variant could depend on the base artifact and offer `CoilImage` overloads that pass the singleton `ImageLoader` by default.

coil/src/main/java/dev/chrisbanes/accompanist/coil/Coil.kt
chrisbanes: SGTM, adding now.

coil/src/main/java/dev/chrisbanes/accompanist/coil/Coil.kt
chrisbanes: SGTM, but I'll leave this as a task for when we upstream back to the Coil library.

PR #7, opened 2020-05-14
Update to Compose v0.1.0-dev11

chrisbanes:
Also added some caching of `Paint` instances as they're expensive to create.

### DIFF
--- a/coil/src/main/java/dev/chrisbanes/accompanist/coil/Coil.kt
+++ b/coil/src/main/java/dev/chrisbanes/accompanist/coil/Coil.kt
@@ -16,6 +16,7 @@
 
 package dev.chrisbanes.accompanist.coil
 
+import android.graphics.ColorMatrix
 import android.graphics.ColorMatrixColorFilter
 import android.graphics.drawable.Drawable
 import androidx.animation.FloatPropKey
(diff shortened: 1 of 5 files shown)


coil/src/main/java/dev/chrisbanes/accompanist/coil/Coil.kt
```diff
@@ -133,8 +137,10 @@ fun CoilImageWithCrossfade(
     crossfadeDuration: Int = defaultTransitionDuration,
     modifier: Modifier = Modifier
 ) {
-    WithConstraints(modifier) { constraints, _ ->
-        var imgLoadState by stateFor(request) { ImageLoadState.Empty }
+    WithConstraints(modifier) {
+        // We key off the data, for the same reasons as executeAsComposable()
+        // below
```
manuelvicnt, line 142: I'd inline this with the line above

chrisbanes: As in expression style?

coil/src/main/java/dev/chrisbanes/accompanist/coil/Coil.kt
```diff
@@ -284,25 +287,29 @@ fun CoilImage(
     }
 }
 
+private val paintPool = Pools.SimplePool<Paint>(2)
```
nickbutcher, line 290: Comment why using a pool here?

PR #30, opened 2020-06-17
Move away from using `WithConstraints`

chrisbanes:
`WithConstraints` is not suitable for how we're using it, so we're moving to `Modifier.onSizeChanged()` (which is based on `Modifier.onPositioned`). This required some changes to how CoilImage is implemented. While here, I also migrated us to the new `launchInComposition()` to launch our Coil request.

### DIFF
--- a/coil/src/main/java/dev/chrisbanes/accompanist/coil/Coil.kt
+++ b/coil/src/main/java/dev/chrisbanes/accompanist/coil/Coil.kt
@@ -19,21 +19,16 @@ package dev.chrisbanes.accompanist.coil
 import android.graphics.drawable.Drawable
 import androidx.compose.Composable
 import androidx.compose.getValue
+import androidx.compose.launchInComposition
 import androidx.compose.onCommit
 import androidx.compose.remember
 import androidx.compose.setValue
-import androidx.compose.stateFor
+import androidx.compose.state
 import androidx.core.graphics.drawable.toBitmap
 import androidx.ui.core.Alignment
-import androidx.ui.core.Constraints
 import androidx.ui.core.ContentScale
 import androidx.ui.core.ContextAmbient
 import androidx.ui.core.Modifier
-import androidx.ui.core.WithConstraints
-import androidx.ui.core.hasBoundedHeight
-import androidx.ui.core.hasBoundedWidth
-import androidx.ui.core.hasFixedHeight
-import androidx.ui.core.hasFixedWidth
 import androidx.ui.foundation.Box
 import androidx.ui.foundation.Image
 import androidx.ui.graphics.ColorFilter
@@ -41,19 +36,11 @@ import androidx.ui.graphics.ImageAsset
 import androidx.ui.graphics.asImageAsset
 import androidx.ui.graphics.painter.ImagePainter
 import androidx.ui.graphics.painter.Painter
+import androidx.ui.unit.IntPxSize
 import coil.Coil
 import coil.decode.DataSource
 import coil.request.GetRequest
 import coil.request.GetRequestBuilder
-import kotlinx.coroutines.CoroutineScope
-import kotlinx.coroutines.Dispatchers
-import kotlinx.coroutines.launch
-
-private val Constraints.requestWidth
-    get() = if (hasFixedWidth || hasBoundedWidth) maxWidth else minWidth
-
-private val Constraints.requestHeight
-    get() = if (hasFixedHeight || hasBoundedHeight) maxHeight else minHeight
 
 /**
  * Creates a composable that will attempt to load the given [data] using [Coil], and then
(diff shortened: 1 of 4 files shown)


coil/src/main/java/dev/chrisbanes/accompanist/coil/Coil.kt
```diff
@@ -135,65 +122,72 @@ fun CoilImage(
     loading: @Composable (() -> Unit)? = null,
     onRequestCompleted: (RequestResult) -> Unit = emptySuccessLambda
 ) {
-    WithConstraints(modifier) {
-        val requestWidth = constraints.requestWidth.value
-        val requestHeight = constraints.requestHeight.value
+    var requestSize by state { IntPxSize.Zero }
+    var result by state<RequestResult?> { null }
+
+    launchInComposition(requestSize) {
```
adamp, line 128: Try to avoid looping back into composition when requestSize changes

coil/src/main/java/dev/chrisbanes/accompanist/coil/Coil.kt
```diff
@@ -135,65 +122,72 @@ fun CoilImage(
...
+    }
 
-        val painter = when (result) {
-            is SuccessResult -> {
-                if (getSuccessPainter != null) {
-                    getSuccessPainter(result)
-                } else {
-                    defaultSuccessPainterGetter(result)
-                }
+    val painter = when (val r = result) {
```
adamp, line 154: Try to avoid doing this work in composition

chrisbanes: Yeah, this is necessary currently because `getSuccessPainter` is `@Composable`. I'll remove this eventually once I think of a good way to implement transitions.

coil/src/main/java/dev/chrisbanes/accompanist/coil/Coil.kt
```diff
@@ -135,67 +123,95 @@ fun CoilImage(
...
+    }
+
+    onDispose {
+        // When we leave the composition, close the Actor channel
+        requestActor.close()
+    }
+
+    onCommit(result) {
+        // Execute the onRequestCompleted callback if we have a new result
+        result?.also { onRequestCompleted(it) }
```
adamp, line 144: Consider running this in the same block as the `result = it` in the CoilRequestActor callback instead of returning to composition with an onCommit; store the received callback in a MutableState in case it changes and call it from the state value

coil/src/main/java/dev/chrisbanes/accompanist/coil/Coil.kt
```diff
@@ -135,67 +123,95 @@ fun CoilImage(
...
+    }
+
+    launchInComposition(requestActor) {
+        // Launch the Actor
+        requestActor.run()
+    }
+
+    onDispose {
+        // When we leave the composition, close the Actor channel
+        requestActor.close()
```
adamp, line 139: you shouldn't need this since the requestActor uses consumeAsFlow - cancelling the call to run() will cancel the channel. Even that is probably unnecessary.

chrisbanes: Agreed, I also don't think it's safe to `consumeAsFlow()` because the Actor lifetime could (maybe) be longer than the composition. Moved to use `receiveAsFlow()` instead.

coil/src/main/java/dev/chrisbanes/accompanist/coil/Coil.kt
chrisbanes: Done, I'm liking where this has landed 👍. Thanks for the help!

coil/src/main/java/dev/chrisbanes/accompanist/coil/Coil.kt
```diff
@@ -135,65 +122,90 @@ fun CoilImage(
     loading: @Composable (() -> Unit)? = null,
     onRequestCompleted: (RequestResult) -> Unit = emptySuccessLambda
 ) {
-    WithConstraints(modifier) {
-        val requestWidth = constraints.requestWidth.value
-        val requestHeight = constraints.requestHeight.value
+    var result by state<RequestResult?> { null }
+    val callback = mutableStateOf(onRequestCompleted)
```
manuelvicnt, line 126: Why is this a state? this value is never changed and `onRequestCompleted` is defined as a param

coil/src/main/java/dev/chrisbanes/accompanist/coil/Coil.kt
```diff
@@ -135,65 +122,90 @@ fun CoilImage(
...
-            }
-            else -> {
-                // Otherwise we have a zero size, so no point executing a request
-                null
+    val requestActor = remember(request) { CoilRequestActor(request) }
+
+    launchInComposition(requestActor) {
+        // Launch the Actor
+        requestActor.run { actorResult ->
+            // Store the result
```
manuelvicnt, line 133: Feels like this comment should reflect that we're updating the state (instead of storing)

coil/src/main/java/dev/chrisbanes/accompanist/coil/Coil.kt
```diff
@@ -135,65 +122,90 @@ fun CoilImage(
...
+            result = actorResult
+
+            if (actorResult != null) {
+                // Execute the onRequestCompleted callback if we have a new result
+                callback.value(actorResult)
```
manuelvicnt, line 138: Isn't this `onRequestCompleted(actorResult)`?

coil/src/main/java/dev/chrisbanes/accompanist/coil/Modifier.kt
```diff
@@ -0,0 +1,41 @@
...
+import androidx.compose.getValue
+import androidx.compose.setValue
+import androidx.compose.state
+import androidx.ui.core.LayoutCoordinates
+import androidx.ui.core.Modifier
+import androidx.ui.core.OnPositionedModifier
+import androidx.ui.core.composed
+import androidx.ui.unit.IntPxSize
+
+internal fun Modifier.onSizeChanged(
```
manuelvicnt, line 28: Document and add an example

coil/src/main/java/dev/chrisbanes/accompanist/coil/RequestActor.kt
```diff
@@ -0,0 +1,43 @@
...
+
+package dev.chrisbanes.accompanist.coil
+
+import kotlinx.coroutines.channels.Channel
+import kotlinx.coroutines.flow.collect
+import kotlinx.coroutines.flow.consumeAsFlow
+import kotlinx.coroutines.flow.distinctUntilChanged
+import kotlinx.coroutines.flow.receiveAsFlow
+
+internal class RequestActor<Param, Result>(
```
manuelvicnt, line 25: Document and add an example, please

coil/src/main/java/dev/chrisbanes/accompanist/coil/Coil.kt
chrisbanes: Nope, the callback has a `RequestResult` type, actorResult can `RequestResult?`

coil/src/main/java/dev/chrisbanes/accompanist/coil/Coil.kt
chrisbanes: This was something @adamp mentioned in https://github.com/chrisbanes/accompanist/pull/30#discussion_r441850532

coil/src/main/java/dev/chrisbanes/accompanist/coil/Coil.kt
```diff
@@ -135,65 +123,106 @@ fun CoilImage(
...
+    // invoke the last provided [onRequestCompleted].
+    //
+    // If a composition happens *after* launchInComposition has launched, the given
+    // [onRequestCompleted] might have changed. If the actor lambda below directly referenced
+    // [onRequestCompleted] it would have captured access to the initial onRequestCompleted
+    // value, not the latest.
+    //
+    // This `callback` state enables the actor lambda to only capture the remembered state
+    // reference, which we can update on each composition.
+    val callback = state { onRequestCompleted }
```
chrisbanes, line 138: @adamp is OK to use `state` here? I was thinking of using `remember { AtomicReference() }` (or some other mutable ref holder) to avoid the unnecessary re-composition.

adamp: the state will make sure that the change isn't seen until the composition commits.

chrisbanes: 👍

PR #33, opened 2020-06-22
Improve MDC theme assertion

ricknout:
This improves the check that requires the context is using a `Theme.MaterialComponents.*` theme. MDC introduced a particular attr for this: [`isMaterialTheme`](https://github.com/material-components/material-components-android/blob/e2eec4aca1795c2795f52098e391c85ccc95a1a4/lib/java/com/google/android/material/color/res/values/attrs.xml#L55).

I considered using [`ThemeEnforcement#obtainStyledAttributes`](https://github.com/material-components/material-components-android/blob/master/lib/java/com/google/android/material/internal/ThemeEnforcement.java#L67) (which makes use of this attr) but it's quite verbose and is currently restricted to the library group.

A basic test for this was added.

Note: AGP was also updated to 4.2.0-alpha02.

### DIFF
--- a/buildSrc/src/main/java/dev/chrisbanes/accompanist/buildsrc/dependencies.kt
+++ b/buildSrc/src/main/java/dev/chrisbanes/accompanist/buildsrc/dependencies.kt
@@ -21,7 +21,7 @@ object Versions {
 }
 
 object Libs {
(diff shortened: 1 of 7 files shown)


ricknout:
Is something up with the CI? Seems to be stuck

chrisbanes:
Weird, just tweaked the protected branch setup. Looks like it's working now!

PR #40, opened 2020-07-09
Stop refetching on every size change by default

chrisbanes:
Currently we invoke Coil on every size change, which is correct but a lot of wasted work most of the time. We no longer re-fetch on every size change. If developers want to customize this, they can use the new `shouldRefetchOnSizeChange` lambda parameter to allow re-fetching as they see fit.

### DIFF
--- a/coil/src/androidTest/java/dev/chrisbanes/accompanist/coil/CoilTest.kt
+++ b/coil/src/androidTest/java/dev/chrisbanes/accompanist/coil/CoilTest.kt
@@ -48,6 +48,7 @@ import kotlinx.coroutines.flow.MutableStateFlow
 import kotlinx.coroutines.runBlocking
 import kotlinx.coroutines.test.TestCoroutineDispatcher
 import kotlinx.coroutines.test.runBlockingTest
+import kotlinx.coroutines.withTimeoutOrNull
 import org.junit.Rule
 import org.junit.Test
 import org.junit.runner.RunWith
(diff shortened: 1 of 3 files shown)


coil/src/androidTest/java/dev/chrisbanes/accompanist/coil/CoilTest.kt
```diff
@@ -175,6 +176,39 @@ class CoilTest {
         loadCompleteSignal.close()
     }
 
+    @OptIn(ExperimentalCoroutinesApi::class)
+    @Test
+    fun basicLoad_changeSize() {
+        val loadCompleteSignal = Channel<Unit>(Channel.UNLIMITED)
+        val sizeFlow = MutableStateFlow(128.dp)
```
manuelvicnt, line 183: What's the lifecycle of this `MutableStateFlow` you cannot cancel / stop / cancel it manually. What happens when the test finishes?

coil/src/androidTest/java/dev/chrisbanes/accompanist/coil/CoilTest.kt
```diff
@@ -175,6 +176,39 @@ class CoilTest {
...
+        runBlocking {
+            loadCompleteSignal.receive()
+        }
+
+        // Now change the size
+        sizeFlow.value = 256.dp
+
+        // Await the potential second load (which shouldn't come)
+        runBlocking {
+            val result = withTimeoutOrNull(3000) { loadCompleteSignal.receive() }
```
manuelvicnt, line 204: This test will take 3 seconds to run every time... can you decrease this timeout?

coil/src/androidTest/java/dev/chrisbanes/accompanist/coil/CoilTest.kt
chrisbanes: `StateFlow` doesn't have a lifecycle, it's just an observable value store. After the test finishes it just gets garbage collected I guess.

coil/src/androidTest/java/dev/chrisbanes/accompanist/coil/CoilTest.kt
chrisbanes: The problem is that these tests aren't very fast, so it could timeout and be a false positive. 3 seconds is good enough for me, the build is far slower than that anyway 🤣

coil/src/androidTest/java/dev/chrisbanes/accompanist/coil/CoilTest.kt
manuelvicnt: SG, was just checking that you thought about this

coil/src/androidTest/java/dev/chrisbanes/accompanist/coil/CoilTest.kt
manuelvicnt: Fair enough 😄

PR #49, opened 2020-07-28
Update to SNAPSHOT 6713401

chrisbanes:


JoseAlcerreca:
Do you still need `freeCompilerArgs += "-Xskip-prerelease-check"`?

chrisbanes:
@JoseAlcerreca doesn't look like it

PR #55, opened 2020-08-06
Compose v0.1.0-dev16

giorgosneokleous93:
- Updated Compose dependencies to `0.1.0-dev16`
- Updated Kotlin to `1.4.0-rc`
- Migrated to the new namespace where necessary according to the [docs](https://developer.android.com/jetpack/androidx/releases/compose#package-refactor)
- Replace deprecated API with new API where needed.

Tested: 
- Run sample project
- Run android tests


giorgosneokleous93:
@googlebot I fixed it.

chrisbanes:
Hi,

First, I'm sorry for the wasted work here. 

We actually already have updates in the `snapshot` branch (which targets Compose snapshot builds). I've just merged that branch in via #56, which bumps us to dev16.

Thanks though for the PR though!
Chris

PR #61, opened 2020-08-14
Fix using CoilImage with an implicit size

chrisbanes:
Currently `CoilImage` only works when you specific a size Modifier, since it waits for `onPositioned` to be called, and only starts a Coil request when it has a non-zero size. If no Modifier is provided, the `CoilImage` will have a zero size and we never execute a Coil request.

Fixed by moving back to `WithConstraints` to move the request size logic to be based on the incoming constraints, rather than the final size, which is probably more correct anyway.


chrisbanes:
> Can you create a new constant called `UNSPECIFIED` which can be `Int.MAX_VALUE` behind the scenes? 
Sounds good, done.

coil/src/main/java/dev/chrisbanes/accompanist/coil/Coil.kt
```diff
@@ -224,6 +224,11 @@ fun CoilImage(
     }
 }
 
+/**
+ * Value for a [IntSize] dimension, where the dimension is not specified or is unknown.s
```
handstandsam, line 228: Thanks for making the update!

Minor typo in the javadoc `.s` at the end.

chrisbanes: Woops, thanks!

coil/src/main/java/dev/chrisbanes/accompanist/coil/Coil.kt
```diff
@@ -186,36 +188,50 @@ fun CoilImage(
...
 
-    if (painter == null) {
-        // If we don't have a result painter, we add a Box with our modifier
-        Box(mod) {
-            // If we don't have a result yet, we can show the loading content
-            // (if not null)
-            if (result == null && loading != null) {
-                loading()
+        if (painter == null) {
+            // If we don't have a result painter, we add a Box with our modifier
```
manuelvicnt, line 208: Nit: The modifier is applied to the WithConstraints above now

```suggestion
            // If we don't have a result painter, we add a Box
```

coil/src/main/java/dev/chrisbanes/accompanist/coil/Coil.kt
```diff
@@ -186,36 +188,50 @@ fun CoilImage(
...
     }
 }
 
+/**
+ * Value for a [IntSize] dimension, where the dimension is not specified or is unknown.
+ */
+private const val UNSPECIFIED = -1
+
+@Stable
+private data class MutableRef<T>(var value: T)
```
manuelvicnt, line 233: Smart! Nit: Add some comments about why you need this?

coil/src/main/java/dev/chrisbanes/accompanist/coil/Coil.kt
chrisbanes: Thanks, done in 05f1927

coil/src/main/java/dev/chrisbanes/accompanist/coil/Coil.kt
chrisbanes: Thanks, done in 05f1927

PR #63, opened 2020-08-19
Upgrade to Coil v0.12.0

chrisbanes:


manuelvicnt (approved):

coil/src/main/java/dev/chrisbanes/accompanist/coil/Coil.kt
```diff
@@ -236,11 +231,12 @@ private const val UNSPECIFIED = -1
 private data class MutableRef<T>(var value: T)
 
 private fun CoilRequestActor(
-    request: GetRequest
+    request: ImageRequest
 ) = RequestActor<IntSize, RequestResult?> { size ->
     when {
-        request.sizeResolver != null -> {
-            // If the request has a sizeResolver set, we just execute the request as-is
+        request.sizeResolver !is DisplaySizeResolver -> {
```
colinrtwhite, line 237: You can actually check `request.defined.sizeResolver != null`. `request.sizeResolver` could be set explicitly to `DisplaySizeResolver`: `ImageRequest.Builder(context).data(...).sizeResolver(DisplaySizeResolver(context)).build()`. `request.defined` is always null unless it's attributes are set explicitly when building the request.

chrisbanes: Thanks! Fixed in #64

PR #72, opened 2020-09-07
Updates Coil docs to match latest public api

JorgeCastilloPrz:
GetRequest and ImageRequest are a single thing now https://github.com/coil-kt/coil/pull/424

Adding this change just because I was about to use it and couldn't copy the snippet on this docs right away.

The only dependency I have is `implementation "dev.chrisbanes.accompanist:accompanist-coil:0.2.1"` (not depending on Coil separately or anything).


chrisbanes:
Thanks for the PR. I fixed this while updating the docs in #89

PR #73, opened 2020-09-11
Snapshot 6824694

nickbutcher:


.idea/codeStyles/Project.xml
```diff
@@ -1,5 +1,23 @@
 <component name="ProjectCodeStyleConfiguration">
   <code_scheme name="Project" version="173">
+    <JetCodeStyleSettings>
```
JoseAlcerreca, line 3: rm (why is this file checked in?)

gradle.properties
```diff
@@ -21,7 +21,7 @@ android.useAndroidX=true
 systemProp.org.gradle.internal.publish.checksums.insecure=true
 
 GROUP=dev.chrisbanes.accompanist
-VERSION_NAME=0.1.10.ui-6765009-SNAPSHOT
+VERSION_NAME=0.2.1
```
chrisbanes, line 24: This needs bumping to 0.2.2.ui-XXXX-SNAPSHOT

.idea/codeStyles/Project.xml
chrisbanes: The file is fine, but these changes should be reverted.

chrisbanes (changes requested):

chrisbanes (approved):

PR #88, opened 2020-09-18
API updates

### LINKED ISSUE #86: Add @Composable placeholder for errors
Something like:
`error: @Composable ((ErrorResult) -> Unit)? = null`

The reason is that you might be needed to display a custom placeholder if the load fails. The might not be a regular image as you might be needed to use some logic there (showing differect vectors assets, drawing checkmarks etc).

chrisbanes:
- getSuccessPainter and getFailurePainter are now gone, replaced
  with `error` and `image` slots.

These slots allow you to provide your own custom content to display the result as required. A simplified overload is provided, for the default case of displaying the result in an `Image`.

This opens up the possibility of removing `CoilWithCrossfade`.

Closes #86


coil/src/main/java/dev/chrisbanes/accompanist/coil/Coil.kt
```diff
@@ -215,26 +173,123 @@ fun CoilImage(
...
-                // If we don't have a result yet, we can show the loading content
-                // (if not null)
-                if (result == null && loading != null) {
-                    loading()
+        when (r) {
+            is SuccessResult -> image(r)
+            is ErrorResult -> if (error != null) error(r)
+            null -> {
+                // If we don't have a result painter, we add a Box...
+                Box {
```
nickbutcher, line 181: Do you need to add the `Box` still now that loading is a slot?

coil/src/main/java/dev/chrisbanes/accompanist/coil/Coil.kt
```diff
@@ -215,26 +173,123 @@ fun CoilImage(
...
         }
 
-        if (painter == null) {
-            // If we don't have a result painter, we add a Box...
-            Box {
-                // If we don't have a result yet, we can show the loading content
-                // (if not null)
-                if (result == null && loading != null) {
-                    loading()
+        when (r) {
```
yrezgui, line 176: Why can't you use `result` instead of `r`?

coil/src/main/java/dev/chrisbanes/accompanist/coil/Coil.kt
```diff
@@ -215,26 +173,123 @@ fun CoilImage(
...
+ * @param loading Content to be displayed when the request is in progress.
+ * @param imageLoader The [ImageLoader] to use when requesting the image. Defaults to [Coil]'s
+ * default image loader.
+ * @param shouldRefetchOnSizeChange Lambda which will be invoked when the size changes, allowing
+ * optional re-fetching of the image. Return true to re-fetch the image.
+ * @param onRequestCompleted Listener which will be called when the loading request has finished.
+ */
+@Composable
+fun CoilImage(
+    data: Any,
```
yrezgui, line 213: You didn't comment `data`

coil/src/main/java/dev/chrisbanes/accompanist/coil/Coil.kt
chrisbanes: Erm, probably not actually.

coil/src/main/java/dev/chrisbanes/accompanist/coil/Coil.kt
chrisbanes: It's a delegate property so the compiler can't guarantee the value won't change.

coil/src/main/java/dev/chrisbanes/accompanist/coil/Coil.kt
chrisbanes: Good catch, added.

yrezgui (approved):

PR #93, opened 2020-09-22
More API changes

### LINKED ISSUE #92: API update proposal
With the recent update we have 3 callback for each situation, such as:

- Image is loading
- Image is loaded
- Image has failed to load.

We also have an internal sealed class `RequestResult` which represents the state of the loader. When the image is loading, the internal code uses `null` instead.

My suggestion is:
- Make it public and rename to something like `CoilState` or `CoilImage`.
- Move all the children classes inside parent for more code clarity, rename it to `Success` and `Fail`.
- Add third state, `object Loading: CoilState()`, and use it instead of nulls, in a similar way Compose does for a lot of components, for example `Modifier`.
- Move nullable `image` property from sealed class to success state, and make it not null.
- Add a single trailing lambda for all the states, e.g. `content: @Composable (CoilState) -> Unit`. This will allow to use syntax like `CoilImage(url){ it: CoilState -> }` with `when` expression inside – it's more easy and consistent rather than using 3 lambdas.
- Add an overload with the lambdas for other cases.

It'd be great to see `CoilImage()` as a state managing function, completely without rendering logic. Overloads with the current functionality may still be available as well.

It'd be interesting to hear more opinions from the community as well.

chrisbanes:
As we're on a roll of API changes, here's one final change:

- `SuccessResult` and `ErrorResult` have now been expanded in the sealed `CoilImageState`, which now contains: `Success`, `Error`, `Loading` & `Empty`
- The recent `CoilImage` with the `image` composable has been expanded and made more flexible:

``` kotlin
CoilImage(
  data = "https://www.image.url",
) { state ->
  when (state) {
    is CoilImageState.Success -> // TODO
    is CoilImageState.Error -> // TODO
    CoilImageState.Loading -> // TODO
    CoilImageState.Empty -> // TODO
  }
}
```

The more opinionated (and easier to use) version which hosts content slots still exists, allowing:

``` kotlin
CoilImage(
  data = "https://www.image.url",
  fadeIn = true,
  loading = { /* content to show while loading */ },
  error = { /* content to show when an error has occurred */ }
)
```

- The tests have been revamped, using [MockWebServer](https://square.github.io/okhttp/4.x/mockwebserver/okhttp3.mockwebserver/-mock-web-server/) to serve the images in a more realistic way.

Closes #92

### DIFF
--- a/coil/src/androidTest/java/dev/chrisbanes/accompanist/coil/CoilTest.kt
+++ b/coil/src/androidTest/java/dev/chrisbanes/accompanist/coil/CoilTest.kt
@@ -78,7 +78,8 @@ class CoilTest {
     @get:Rule
     val composeTestRule = createComposeRule()
 
-    private val server = coilTestWebServer()
+    // Our MockWebServer. We use a response delay to simulate real-world conditions
+    private val server = coilTestWebServer(responseDelayMs = 200)
 
     @Before
     fun setup() {
@@ -485,7 +486,7 @@ private fun resourceUri(id: Int): Uri {
  * [MockWebServer] which returns a valid at the path `/image` and a 404 for anything else.
  * We add a small delay to simulate 'real-world' network conditions.
  */
-private fun coilTestWebServer(): MockWebServer {
+private fun coilTestWebServer(responseDelayMs: Long = 0): MockWebServer {
     val dispatcher = object : Dispatcher() {
         override fun dispatch(request: RecordedRequest): MockResponse = when (request.path) {
             "/image" -> {
@@ -497,10 +498,14 @@ private fun coilTestWebServer(): MockWebServer {
                 }
 
                 MockResponse()
+                    .setHeadersDelay(responseDelayMs, TimeUnit.MILLISECONDS)
                     .addHeader("Content-Type", "image/jpeg")
                     .setBody(imageBuffer)
             }
-            else -> MockResponse().setResponseCode(404)
+            else ->
+                MockResponse()
+                    .setHeadersDelay(responseDelayMs, TimeUnit.MILLISECONDS)
+                    .setResponseCode(404)
         }
     }
 
--- a/coil/src/androidTest/java/dev/chrisbanes/accompanist/coil/CoilTest.kt
+++ b/coil/src/androidTest/java/dev/chrisbanes/accompanist/coil/CoilTest.kt
@@ -21,7 +21,6 @@ import android.net.Uri
 import androidx.compose.foundation.Image
 import androidx.compose.foundation.Text
 import androidx.compose.foundation.layout.preferredSize
-import androidx.compose.runtime.Composable
 import androidx.compose.runtime.collectAsState
 import androidx.compose.ui.Modifier
 import androidx.compose.ui.graphics.Color
@@ -59,6 +58,13 @@ import kotlinx.coroutines.runBlocking
 import kotlinx.coroutines.test.TestCoroutineDispatcher
 import kotlinx.coroutines.test.runBlockingTest
 import kotlinx.coroutines.withTimeoutOrNull
+import okhttp3.mockwebserver.Dispatcher
+import okhttp3.mockwebserver.MockResponse
+import okhttp3.mockwebserver.MockWebServer
+import okhttp3.mockwebserver.RecordedRequest
+import okio.Buffer
+import org.junit.After
+import org.junit.Before
 import org.junit.Rule
 import org.junit.Test
 import org.junit.runner.RunWith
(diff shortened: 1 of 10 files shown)


coil/src/androidTest/java/dev/chrisbanes/accompanist/coil/CoilTest.kt
```diff
@@ -388,7 +466,50 @@ class CoilTest {
...
+        .build()
+}
+
+private fun resourceUri(id: Int): Uri {
+    val packageName = InstrumentationRegistry.getInstrumentation().targetContext.packageName
+    return "${ContentResolver.SCHEME_ANDROID_RESOURCE}://$packageName/$id".toUri()
+}
+
+/**
+ * [MockWebServer] which returns a valid at the path `/image` and a 404 for anything else.
```
nickbutcher, line 486: "A valid __ at path" response?

coil/src/main/java/dev/chrisbanes/accompanist/coil/Coil.kt
```diff
@@ -48,17 +48,19 @@ import coil.request.ImageRequest
 import coil.request.ImageResult
 
 /**
- * Creates a composable that will attempt to load the given [data] using [Coil], and then
- * display the result in the provided [image] content.
- *
- * This version of the function allows complete control over how the loaded image is displayed,
- * by being able to provide custom layout:
+ * Creates a composable that will attempt to load the given [data] using [Coil], and provides
+ * complete content of how the current state is displayed:
```
nickbutcher, line 52: s/content/control?

coil/src/main/java/dev/chrisbanes/accompanist/coil/Coil.kt
```diff
@@ -69,54 +71,41 @@ import coil.request.ImageResult
...
     modifier: Modifier = Modifier,
     imageLoader: ImageLoader = Coil.imageLoader(ContextAmbient.current),
-    shouldRefetchOnSizeChange: (currentResult: RequestResult, size: IntSize) -> Boolean = defaultRefetchOnSizeChangeLambda,
-    onRequestCompleted: (RequestResult) -> Unit = emptySuccessLambda,
-    error: @Composable ((ErrorResult) -> Unit)? = null,
-    loading: @Composable (() -> Unit)? = null,
-    image: @Composable (SuccessResult) -> Unit
+    shouldRefetchOnSizeChange: (currentResult: CoilImageState, size: IntSize) -> Boolean = defaultRefetchOnSizeChangeLambda,
+    onRequestCompleted: (CoilImageState) -> Unit = emptySuccessLambda,
+    content: @Composable (state: CoilImageState) -> Unit
```
nickbutcher, line 83: what about naming the param `imageState` — state is a pretty overloaded term.

coil/src/main/java/dev/chrisbanes/accompanist/coil/Coil.kt
```diff
@@ -69,54 +71,41 @@ import coil.request.ImageResult
...
 }
 
 /**
- * Creates a composable that will attempt to load the given [request] using [Coil], and then
- * display the result in the provided [image] content.
- *
- * This version of the function allows complete control over how the loaded image is displayed,
- * by being able to provide custom layout:
+ * Creates a composable that will attempt to load the given [request] using [Coil], and provides
+ * complete content of how the current state is displayed:
```
nickbutcher, line 97: ditto

coil/src/main/java/dev/chrisbanes/accompanist/coil/Coil.kt
```diff
@@ -179,34 +163,49 @@ fun CoilImage(
         // We remember the last size in a MutableRef (below) rather than a MutableState.
         // This is because we don't need value changes to trigger a re-composition, we are only
         // using it to store the last value.
-        val lastRequestedSize = remember(requestActor) { MutableRef(IntSize.Zero) }
+        val lastRequestedSize = remember(requestActor) { MutableRef<IntSize?>(null) }
```
nickbutcher, line 166: Use delegate?

nickbutcher (approved):
A few small doc suggestions but changes LG. I like the easy default but ability to customize everything if you want.

coil/src/androidTest/java/dev/chrisbanes/accompanist/coil/CoilTest.kt
chrisbanes: Done

coil/src/main/java/dev/chrisbanes/accompanist/coil/Coil.kt
chrisbanes: Done

coil/src/main/java/dev/chrisbanes/accompanist/coil/Coil.kt
chrisbanes: Done

coil/src/main/java/dev/chrisbanes/accompanist/coil/Coil.kt
chrisbanes: Can't, this is a custom `MutableRef` rather than `MutableState` (so no `getValue`, etc)

PR #95, opened 2020-09-22
Update to Coil 1.0.0-rc3

chrisbanes:
Turned off `-Xjvm-default=all` as per Coil [release notes](https://github.com/coil-kt/coil/releases/tag/1.0.0-rc3), and started using new `imageLoader` extension property.

/cc @colinrtwhite


colinrtwhite (approved):
🙌

PR #128, opened 2020-10-23
Add PicassoAmbient for Picasso

chrisbanes:


picasso/src/main/java/dev/chrisbanes/accompanist/picasso/Picasso.kt
```diff
@@ -47,6 +48,11 @@ import kotlinx.coroutines.suspendCancellableCoroutine
 import okhttp3.HttpUrl
 import java.io.File
 
+/**
+ * Ambient containing the preferred [Picasso] to use in [PicassoImage]. Defaults to [Picasso.get].
+ */
+val AmbientPicasso = staticAmbientOf { Picasso.get() }
```
manuelvicnt, line 54: I was a bit sceptical at the beginning about this implementation as it's pretty similar to the `Coil` one but very different in reality. There you have something like `PicassoConstants.defaultPicasso()` and here, an Ambient with a default value. 

After thinking more about it, I think that's fine as the image loader is a param that you pass to make the call and this Picasso is with what you make the call.

manuelvicnt (approved):

picasso/src/main/java/dev/chrisbanes/accompanist/picasso/Picasso.kt
chrisbanes: Yeah, this is the ideal implementation. We can't do that in Coil though because it needs a `Context` 😞

PR #134, opened 2020-11-05
Update to snapshot 6953474.

nickbutcher:


coil/build.gradle
```diff
@@ -82,6 +82,7 @@ dependencies {
 
     implementation Libs.AndroidX.coreKtx
     implementation Libs.AndroidX.Compose.runtime
+    implementation Libs.AndroidX.Compose.material
```
chrisbanes, line 85: I'd rather not add this as a library dependency. Can we move this to be a test dependency instead?

nickbutcher: Done.

chrisbanes (approved):

PR #142, opened 2020-11-13
Keyboard (IME) animation support

chrisbanes:
This brings initial support for `WindowInsetsAnimation` to Accompanist Insets. The API is marked as experimental, since it's not final and will likely be refined.

- [x] Refine the API?
- [x] Add parameter to enable/disable `WindowInsetsAnimation`support

### DIFF
--- a/insets/src/main/java/dev/chrisbanes/accompanist/insets/Insets.kt
+++ b/insets/src/main/java/dev/chrisbanes/accompanist/insets/Insets.kt
@@ -370,12 +370,18 @@ private fun Insets.updateFrom(windowInsets: WindowInsetsPlatform, type: Int) {
     isVisible = windowInsets.isVisible(type)
 }
 
-internal fun Insets.coerceEachDimensionAtLeast(other: Insets): Insets = copy(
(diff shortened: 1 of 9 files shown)


insets/src/main/java/dev/chrisbanes/accompanist/insets/Insets.kt
```diff
@@ -167,5 +255,12 @@ private fun Insets.updateFrom(windowInsets: WindowInsetsCompat, type: Int) {
     isVisible = windowInsets.isVisible(type)
 }
 
+internal fun Insets.coerceAtLeastEachDimension(other: Insets): Insets = copy(
+    left = left.coerceAtLeast(other.left),
+    top = top.coerceAtLeast(other.top),
+    right = right.coerceAtLeast(other.top),
```
consp1racy, line 261: copypasta error here

Zhuinden: `top` is `right`, `right` is `top` :thinking:

sjmueller:
Excited for this!

PR #149, opened 2020-11-18
[Image Loading] Use rememberUpdatedState

chrisbanes:
This replaces some existing code which achieves the same pattern.

### DIFF
--- a/imageloading-core/src/main/java/dev/chrisbanes/accompanist/imageloading/ImageLoad.kt
+++ b/imageloading-core/src/main/java/dev/chrisbanes/accompanist/imageloading/ImageLoad.kt
@@ -25,8 +25,8 @@ import androidx.compose.runtime.MutableState
 import androidx.compose.runtime.Stable
 import androidx.compose.runtime.getValue
 import androidx.compose.runtime.mutableStateOf
-import androidx.compose.runtime.referentialEqualityPolicy
 import androidx.compose.runtime.remember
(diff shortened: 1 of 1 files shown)


imageloading-core/src/main/java/dev/chrisbanes/accompanist/imageloading/ImageLoad.kt
```diff
@@ -72,18 +72,7 @@ fun <R : Any, TR : Any> ImageLoad(
...
-    // invoke the last provided [onRequestCompleted].
-    //
-    // If a composition happens *after* LaunchedEffect has launched, the given
-    // [onRequestCompleted] might have changed. If the actor lambda below directly referenced
-    // [onRequestCompleted] it would have captured access to the initial onRequestCompleted
-    // value, not the latest.
-    //
-    // This `callback` state enables the actor lambda to only capture the remembered state
-    // reference, which we can update on each composition.
-    val callback = remember { mutableStateOf(onRequestCompleted, referentialEqualityPolicy()) }
```
nickbutcher, line 85: Ok to drop the `referentialEqualityPolicy()`?

chrisbanes: Yep. The only thing which reads it isn't composable so there's nothing to recompose.

nickbutcher (approved):

PR #158, opened 2020-11-27
Add ViewWindowInsetObserver

### LINKED ISSUE #155: Initial flicker when using ProvideWindowInsets
I think the issue is that we only request insets once a composition has happened. The starting values are all `0` the initial layout of your composables will be as if there were no insets.

On the first composition we request insets, then need to wait for the next view layout pass (for dispatch), then the next composition to update the right values, then the new Compose layout.

AFAICT This is mostly an issue for apps which use `Fragment`s and then `ProvideWindowInsets` in each, since all the above happens on each fragment view creation. We should investigate if there is a way for us to provide an API which allows fragments to 'prime' the values earlier.

Raised by @apkelly

chrisbanes:
This enables fragments to manage their own `WindowInsets` and setup the necessary listeners earlier than `ProvideWindowInsets`. See the related issue for more information on why this is needed.

Fixes #155

### DIFF
--- a/insets/src/main/java/dev/chrisbanes/accompanist/insets/Insets.kt
+++ b/insets/src/main/java/dev/chrisbanes/accompanist/insets/Insets.kt
@@ -130,7 +130,7 @@ val AmbientWindowInsets = staticAmbientOf<WindowInsets> {
  *
  *     // Call start() to start listening now.
  *     // The WindowInsets instance is returned to us.
- *     val windowInsets = observer.observe()
+ *     val windowInsets = observer.start()
  *
  *     setContent {
  *         // Instead of calling ProvideWindowInsets, we use Providers to provide
--- a/insets/src/main/java/dev/chrisbanes/accompanist/insets/Insets.kt
+++ b/insets/src/main/java/dev/chrisbanes/accompanist/insets/Insets.kt
@@ -105,8 +105,87 @@ val AmbientWindowInsets = staticAmbientOf<WindowInsets> {
 }
 
 /**
+ * This function sets up the necessary listeners on the given [view] to be able to observe
+ * [WindowInsetsCompat].
+ *
+ * This function is useful for when you prefer to handle the ownership of the [WindowInsets]
+ * yourself. One example of this is if you find yourself using [ProvideWindowInsets] in multiple
+ * fragments.
+ *
+ * It is convenient to use [ProvideWindowInsets] in each fragment, but that result in a delay in
+ * the insets being updated on startup, which visually results in a flicker.
+ * See [this issue](https://github.com/chrisbanes/accompanist/issues/155) for more information.
+ *
+ * The alternative is for fragments to manage the [WindowInsets] themselves, and call this function
+ * once in `onCreateView()`:
+ *
+ * ```
+ * override fun onCreateView(
+ *     inflater: LayoutInflater,
+ *     container: ViewGroup?,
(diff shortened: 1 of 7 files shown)


insets/src/main/java/dev/chrisbanes/accompanist/insets/Insets.kt
```diff
@@ -104,10 +104,89 @@ val AmbientWindowInsets = staticAmbientOf<WindowInsets> {
     error("AmbientInsets value not available. Are you using ProvideWindowInsets?")
 }
 
+/**
+ * This function sets up the necessary listeners on the given [view] to be able to observe
+ * [WindowInsetsCompat] instance dispatched by the system.
+ *
+ * This function is useful for when you prefer to handle the ownership of the [WindowInsets]
+ * yourself. One example of this is if you find yourself using [ProvideWindowInsets] in multiple
```
nickbutcher, line 112: I don't think that the "multiple" part is helpful here as the issue exhibits with a single fragment right? It's a timing issue.

insets/src/main/java/dev/chrisbanes/accompanist/insets/Insets.kt
```diff
@@ -105,22 +105,74 @@ val AmbientWindowInsets = staticAmbientOf<WindowInsets> {
...
+ *     savedInstanceState: Bundle?
+ * ): View = ComposeView(requireContext()).apply {
+ *     layoutParams = LayoutParams(MATCH_PARENT, MATCH_PARENT)
+ *
+ *     // Create an ViewWindowInsetObserver using this view
+ *     val observer = ViewWindowInsetObserver(this)
+ *
+ *     // Call start() to start listening now.
+ *     // The WindowInsets instance is returned to us.
+ *     val windowInsets = observer.observe()
```
nickbutcher, line 133: This should be `start` right?

```suggestion
 *     val windowInsets = observer.start()
```

chrisbanes: Yep, woops!

insets/src/main/java/dev/chrisbanes/accompanist/insets/Insets.kt
```diff
@@ -105,22 +105,74 @@ val AmbientWindowInsets = staticAmbientOf<WindowInsets> {
...
+ *     savedInstanceState: Bundle?
+ * ): View = ComposeView(requireContext()).apply {
+ *     layoutParams = LayoutParams(MATCH_PARENT, MATCH_PARENT)
+ *
+ *     // Create an ViewWindowInsetObserver using this view
+ *     val observer = ViewWindowInsetObserver(this)
+ *
+ *     // Call start() to start listening now.
+ *     // The WindowInsets instance is returned to us.
+ *     val windowInsets = observer.start()
```
nickbutcher, line 133: Don't think that this will `stop` itself right? What about using `LifecycleObserver` and `ViewTreeLifecycleOwner` to make this unregister itself as appropriate?

chrisbanes: It doesn't need to unregister as everything is driven from the view. If the view is removed, nothing happens.

nickbutcher: ack

nickbutcher (approved):
Thanks for the changes!

PR #164, opened 2020-12-07
[Image] Throw exception if incorrect data type is passed in

chrisbanes:
GlideImage, CoilImage and PicassoImage currently all accept an `Any` param for data. There have been examples of developers passing in things like Drawables, ImageBitmaps, and others, which are not supported. To aid developers, we now throw an exception when any `Drawable`, `ImageBitmap`, `ImageVector` or `Painter` is passed in.


nickbutcher (approved):
Changes LG but rather than throwing for incorrect usage, is accepting `Any` the right API? Would more restrictive types help prevent these errors?

chrisbanes:
I don't really want to maintain more overloads for each type which each library supports. 

We could wrap the data param in a sealed class to enforce the type, but that adds friction for usage.

PR #177, opened 2021-01-07
[Insets] Use WindowInsetsAnimationCompat

chrisbanes:
Doing so allows IME animations to work on devices before API 30. 🎉

_Video incoming_


manuelvicnt (approved):
LGTM

PR #188, opened 2021-01-23
Update to the Transition v2 APIs

### LINKED ISSUE #187: Migrate away from transition V1 API
Suppresed the deprecation in #184, but we should migrate to the new APIs.

chrisbanes:
Required API change, since the `clock` param is no longer necessary.

Fixes #187


manuelvicnt (approved):

chrisbanes:
Got a verbal LGTM from Doris too.

PR #204, opened 2021-02-19
Migrate to using produceState

chrisbanes:
Our custom RequestActor was working fine, but using the built-in side effect means less code for us to maintain.

### DIFF
--- a/imageloading-core/src/main/java/dev/chrisbanes/accompanist/imageloading/ImageLoad.kt
+++ b/imageloading-core/src/main/java/dev/chrisbanes/accompanist/imageloading/ImageLoad.kt
@@ -74,7 +74,6 @@ fun <R : Any, TR : Any> ImageLoad(
         initialValue = ImageLoadState.Loading,
         key1 = requestKey,
         key2 = requestSize,
-        key3 = updatedTransformRequestForSize,
     ) {
         value = requestSize?.let { updatedTransformRequestForSize(request, it) }
             ?.let { transformedRequest ->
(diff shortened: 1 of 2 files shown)


imageloading-core/src/main/java/dev/chrisbanes/accompanist/imageloading/ImageLoad.kt
```diff
@@ -68,78 +64,43 @@ fun <R : Any, TR : Any> ImageLoad(
...
 
-            if (newState is ImageLoadState.Success || newState is ImageLoadState.Error) {
-                callback(newState)
-            }
-        }
+    val loadState by produceState<ImageLoadState>(
+        initialValue = ImageLoadState.Loading,
+        key1 = requestKey,
+        key2 = requestSize,
+        key3 = updatedTransformRequestForSize,
```
manuelvicnt, line 77: I don't think you need `key3`. Since `updatedTransformRequestForSize` is backed by `rememberUpdatedState`, any update to `transformRequestForSize` won't cause a recomposition here

imageloading-core/src/main/java/dev/chrisbanes/accompanist/imageloading/ImageLoad.kt
```diff
@@ -68,78 +64,43 @@ fun <R : Any, TR : Any> ImageLoad(
...
-        ) {
-            val transformedRequest = transformRequestForSize(request, requestSize)
-            if (transformedRequest != null) {
-                requestActor.send(transformedRequest)
-                lastRequestedSize.value = requestSize
-            } else {
-                // If the transform request is null, set our state to empty
-                state = ImageLoadState.Empty
-            }
+        val lastSize = requestSize
```
manuelvicnt, line 98: Why do you need `lastSize` as a backing property?

imageloading-core/src/main/java/dev/chrisbanes/accompanist/imageloading/ImageLoad.kt
chrisbanes: Good point, removed.

imageloading-core/src/main/java/dev/chrisbanes/accompanist/imageloading/ImageLoad.kt
chrisbanes: No reason, other than being able to keep the next line on 1 line. Removed.

manuelvicnt (approved):

Issue #141, opened 2020-11-12
CoilImage: ImageState is Empty (briefly) before the image loads.


Bradleycorn:
When using the "content version" of the `CoilImage` composable, it is composed in the "empty" state briefly before the request starts (and it then gets recomposed in the "loading" state). This **could** cause a brief "flash" of the empty state, if the device is slow. I haven't seen it on a physical device, but if you add a log message in the content composable for the Empty state, you will see it. One way that I HAVE been able to see it, is to create a `@Preview` for the below composable, and then in Android Studio click the "Deploy Preview" button to deploy the preview to an emulator. When the preview app loads on the emulator, you can see the brief flash of the Empty State before the Loading state. For example:

```
@Composable
fun MyImage() {
    CoilImage(
        data = "https://picsum.photos/300/300",
    ) { imageState ->
        when (imageState) {
            is ImageLoadState.Success -> {
                MaterialLoadingImage(
                    result = imageState,
                    fadeInEnabled = true,
                    fadeInDurationMs = 600,
                )
            }
            is ImageLoadState.Error -> /* TODO */
            ImageLoadState.Loading -> Text("Loading...")
            ImageLoadState.Empty -> {
                  Log.d(TAG, "COIL IMAGE IS EMPTY")
                  Text("⚠️ EMPTY! ⚠️")
            }
        }
    }
}
```

When the `MyImage` composable is called the UI will go through the following states:

1. ⚠️ EMPTY! ⚠️ (You won't typically see this on a device, but if you check the LogCat, you will find the "COIL IMAGE IS EMPTY" message logged there).
2. Loading...
3. ![image](https://user-images.githubusercontent.com/4521694/99001584-1c05af00-2509-11eb-9f1d-41be541d17ed.png)

It seems like it should go directly to the "Loading" state, and then to the Success (show the image) state, and skip the Empty state all together in this case.

chrisbanes:
I was initially thinking that we could change the internal loading logic to add a debounce on the `Loading` state, but the more I think about this, the more than feels like a UI thing.

The way I'm thinking about this is that we wrap the `loading()` content with something like this:

``` kotlin
@Composable
fun WithDelay(
    delayMs: Long,
    content: @Composable () -> Unit
) {
    var visible by remember { mutableStateOf(false) }

    LaunchedEffect(Unit) {
        delay(delayMs)
        visible = true
    }

    if (visible) content()
}
```

You can then do:

``` kotlin
CoilImage(
    // ...
[...]

PR #206, opened 2021-02-19
Update image fade

chrisbanes:
- We now fade in each painter (if it changes)
- Updated to new transition best practices

### DIFF
--- a/imageloading-core/src/main/java/dev/chrisbanes/accompanist/imageloading/MaterialLoadingImage.kt
+++ b/imageloading-core/src/main/java/dev/chrisbanes/accompanist/imageloading/MaterialLoadingImage.kt
@@ -24,6 +24,10 @@ import androidx.compose.animation.core.tween
 import androidx.compose.animation.core.updateTransition
 import androidx.compose.foundation.Image
 import androidx.compose.runtime.Composable
+import androidx.compose.runtime.Stable
+import androidx.compose.runtime.State
+import androidx.compose.runtime.getValue
+import androidx.compose.runtime.mutableStateOf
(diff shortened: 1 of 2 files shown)


imageloading-core/src/main/java/dev/chrisbanes/accompanist/imageloading/MaterialLoadingImage.kt
```diff
@@ -173,40 +190,47 @@ fun MaterialLoadingImage(
 }
 
 @Composable
-private fun fadeInColorFilter(durationMs: Int): ColorFilter {
+private fun updateFadeInTransition(key: Any, durationMs: Int): FadeInTransition {
     // Create our transition state, which allow us to control the state and target states
-    val transitionState = remember { MutableTransitionState(ImageLoadTransitionState.Empty) }
-    transitionState.targetState = ImageLoadTransitionState.Loaded
+    val transitionState = remember(key) {
+        MutableTransitionState(ImageLoadTransitionState.Empty).apply {
```
nickbutcher, line 196: Add a comment here for others to learn from e.g. 

`// Start with Empty state then immediately transition to Loaded`?

nickbutcher (approved):
One request for a comment, otherwise LGTM

imageloading-core/src/main/java/dev/chrisbanes/accompanist/imageloading/MaterialLoadingImage.kt
chrisbanes: That is actually done on purpose due to #141

Issue #207, opened 2021-02-19
[Idea] Provide inset aware layouts


chrisbanes:
We could make it easier to support insets, through a custom `Scaffold` implementation, as well making `InsetAwareTopAppBar` a real thing ™️ 

Candidates:

- `Scaffold`
- `TopAppBar`
- `BottomNavigation`
- `ModalDrawerLayout`
- `BottomAppBar` (less priority)

dovahkiin98:
I don't know if adding custom composables of already existing ones is a good idea. That would require alot of maintanance in the future. If `Scaffold` is changed, so will `InsetAwareScaffold` have to change.

Wouldn't a simple modifier that applies the insets to the any Composable? Something like `.padding(WindowInsets)` or `applyWindowInsets()`

chrisbanes:
> If `Scaffold` is changed, so will `InsetAwareScaffold` have to change.

Compose’s API is now API stable, so there won’t be any API changes. This library isn’t designed to live forever either.

> Wouldn't a simple modifier that applies the insets to the any Composable? Something like `.padding(WindowInsets)` or `applyWindowInsets()`

We already have that: https://google.github.io/accompanist/insets/

dovahkiin98:
From what I remember in Flutter, only the `Scaffold` was insets aware, and it passed those insets to its children. So maybe that's how it should be done in here as well?

chrisbanes:
Sure, but we're not copying Flutter, or being limited to what Flutter provides.

Omico:
Hi, I made a very early version of InsetAwareScaffold.
```
class InsetAwareState(
    var topBarSize: IntSize = IntSize.Zero,
    var bottomBarSize: IntSize = IntSize.Zero,
)

@Composable
fun rememberInsetAwareState(): InsetAwareState = remember { InsetAwareState() }

fun Modifier.padding(
    insetAwareState: InsetAwareState,
    additionalTop: Dp = 0.dp,
    additionalBottom: Dp = 0.dp,
) = composed {
    padding(
        LocalWindowInsets.current.systemBars.toPaddingValues(
            top = false,
            additionalTop = with(LocalDensity.current) { insetAwareState.topBarSize.height.toDp() + additionalTop },
            additionalBottom = with(LocalDensity.current) { insetAwareState.bottomBarSize.height.toDp() + additionalBottom },
        )
    )
}

@Composable
fun InsetAwareScaffold(
    modifier: Modifier = Modifier,
[...]

PR #208, opened 2021-02-21
Fix actions or navigationIcon covered by navigation bar in landscape

AfzalivE:
The sample for `InsetAwareTopAppBar` currently doesn't take into account the navigation bar in landscape mode. When in landscape, either the `navigationIcon` or an `action` are overlapped by the Navigation bar. This PR fixes this so that all icons are visible in portrait as well as landscape mode.

Ideally, I think this might be better off applied in something like an `InsetAwareScaffold` since the `bodyContent` will also need to have this modifier.


chrisbanes:
Thanks! You might be interested in #207

chrisbanes (approved):

PR #211, opened 2021-02-25
Add contentDescription argument to readme

hitherejoe:
Add contentDescription argument to CoilImage declarations, as this is a required argument when using the composable.

I'm not sure if this was intentionally left out to keep things simple, but I feel having this here removes the need for users to question why they are seeing errors after copying any samples from here. Feel free to disregard this if it was intentional :)

### DIFF
--- a/coil/README.md
+++ b/coil/README.md
@@ -14,7 +14,8 @@ The simplest usage is like so:
 
 ```kotlin 
 CoilImage(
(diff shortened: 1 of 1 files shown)


hitherejoe:
@googlebot I signed it!

chrisbanes (approved):
Thanks!

PR #213, opened 2021-02-25
Fixed ColorMatrix brightness adjustment

kosyak:
I'm not a ColorMatrix expert but it seems to me the correct way to set brightness is to use 5th columns of color matrix - not 4th.

Here's another implementation of Material image loading that uses the proposed way of setting brightness http://blog.neteril.org/blog/2014/11/23/android-material-image-loading/

Feel free to close this if I'm wrong.


kosyak:
@googlebot I signed it!

chrisbanes:
Woops, great catch! I converted this code to use the new `ColorMatrix` support in #195, and obviously typed the wrong number in.

chrisbanes (approved):

PR #236, opened 2021-03-11
[Pager] Migrate to Modifier.scrollable

### LINKED ISSUE #247: [Pager] Migrate over to Modifier.scrollable()

chrisbanes:
Fixes #247


pager/src/main/java/com/google/accompanist/pager/PagerState.kt
```diff
@@ -85,12 +88,26 @@ class PagerState(
...
     @IntRange(from = 0) currentPage: Int = 0,
     @FloatRange(from = 0.0, to = 1.0) currentPageOffset: Float = 0f,
-) {
+) : ScrollableState {
     private var _pageCount by mutableStateOf(pageCount)
     private var _currentPage by mutableStateOf(currentPage)
     private val _currentPageOffset = mutableStateOf(currentPageOffset)
     internal var pageSize by mutableStateOf(0)
 
+    private val globalPosition: Float
```
nickbutcher, line 97: Comment

pager/src/main/java/com/google/accompanist/pager/PagerState.kt
```diff
@@ -322,39 +335,60 @@ class PagerState(
...
             animate(
-                initialValue = currentPageOffset * pageSize,
-                targetValue = pageSize * determineSpringBackOffset(
-                    velocity = initialVelocity * -1,
-                    offset = targetOffset
-                ),
+                initialValue = globalPosition * pageSize,
+                targetValue = targetPosition * pageSize,
                 initialVelocity = initialVelocity,
                 animationSpec = spring()
```
nickbutcher, line 368: I know not touched in this PR but is the `spring` what we want for settling and/or should it be configurable? I feel like ViewPager had a specific interpolator, do we want to approximate that?

nickbutcher (approved):
LGTM, a few things that might be clarified/commented.

pager/src/main/java/com/google/accompanist/pager/PagerState.kt
```diff
@@ -160,9 +177,9 @@ class PagerState(
 
         if (page == currentPage) return
 
-        // We don't specifically use the DragScope's dragBy, but
+        // We don't specifically use the ScrollScope's scrollBy, but
```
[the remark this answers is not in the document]
chrisbanes: So `scroll {}` gives you a `ScrollScope`. The normal usage is:

```
scroll {
    // scrollBy is provided by the scope
    scrollBy(someDelta)
}
```

but we don't need the `scrollBy` here, only the fact that the block is run in a lock.

pager/src/main/java/com/google/accompanist/pager/PagerState.kt
chrisbanes: done

pager/src/main/java/com/google/accompanist/pager/PagerState.kt
chrisbanes: Good call, done. I don't really want to reimplement VP's scroll interpolator, but I increased the spring stiffness which feels close.

pager/src/main/java/com/google/accompanist/pager/Pager.kt
```diff
@@ -97,51 +108,62 @@ private val Measurable.page: Int
 fun HorizontalPager(
     state: PagerState,
     modifier: Modifier = Modifier,
+    reverseLayout: Boolean = false,
     @IntRange(from = 1) offscreenLimit: Int = 1,
-    content: @Composable PagerScope.(page: Int) -> Unit
+    decayAnimationSpec: DecayAnimationSpec<Float> = defaultDecayAnimationSpec(),
+    snapAnimationSpec: AnimationSpec<Float> = spring(stiffness = SnapSpringStiffness),
```
matvei-g, line 114: Here's is an idea: since your internal PagerState.fling uses only public API available (if not, it's still a good idea :) ) what you can do here is to replace two animation specs with one FlingBehaviour, e.g.
```
object PagerDefaults {
     @Composable
     fun defaultPagerFlingConfig(state: PagerState, decayAnimationSpec: DecayAnimationSpec<Float>, snapAnimationSpec: AnimationSpec<Float>): FlingConfig { ... }
}
```

Or similar, then you can just open the whole flingbehaviour for overriding (make FlingBehaviour a param in HorizontalPager), so people can:
1. Use default behavior by providing nothing
2. Use default with tweaked params
3. Use their own sophisticated or slightly different fling behaviour by providing their own instance.
4. The API for customization will be consistent with LazyColumn and friends.

What do you think?

pager/src/main/java/com/google/accompanist/pager/Pager.kt
```diff
@@ -97,51 +108,62 @@ private val Measurable.page: Int
...
+    content: @Composable PagerScope.(page: Int) -> Unit,
 ) {
     require(offscreenLimit >= 1) { "offscreenLimit is required to be >= 1" }
 
-    val reverseScroll = LocalLayoutDirection.current == LayoutDirection.Rtl
-
-    val density = LocalDensity.current
-    val decay = remember(density) { splineBasedDecay<Float>(density) }
+    val isRtl = LocalLayoutDirection.current == LayoutDirection.Rtl
+    val reverseDirection = if (isRtl) !reverseLayout else reverseLayout
```
matvei-g, line 120: If you do `if (isRtl) reverseLayout else !reverseLayout` to "reverse the reverse", do you still need the shenanigans with minuses in the rest of the logic?

matvei-g (commented):
Thanks Chris, just two smallish comments!

pager/src/main/java/com/google/accompanist/pager/Pager.kt
chrisbanes: Yeah I like it, I'll follow up after #254

PR #243, opened 2021-03-15
Fix crash when one constraint is set before the other

troyhen:
Fix crash when Coil is used with ConstraintLayout and dimensions are set independently


chrisbanes (approved):

chrisbanes:
Thanks!

PR #264, opened 2021-03-18
New SwipeRefresh composable

### LINKED ISSUE #192: [Idea] New SwipeRefresh library
This would take the [SwipeRefreshLayout](https://github.com/android/compose-samples/blob/main/JetNews/app/src/main/java/com/example/jetnews/ui/SwipeToRefresh.kt) component from JetNews, and fix the issues around state->event trampoline, nested scrolling, etc.

Things to note:
- There's an [issue](https://issuetracker.google.com/issues/160576011) for this to be added to Compose Material but there's currently no progress
- Technically swipe refresh isn't in the current Material Design spec, hence why we have [`androidx.swiperefreshlayout`](https://developer.android.com/jetpack/androidx/releases/swiperefreshlayout)

chrisbanes:
Fixes #192 

## Tasks

- [x] Working indicator! Thanks @ricknout 
- [x] kdoc
- [x] API review (by @matvei-g)
- [ ] Doc page (following up in #324)

## Future things

- [ ] Add further indicator customization (switching colors?)


sample/src/main/java/com/google/accompanist/sample/swiperefresh/DocsSamples.kt
```diff
@@ -0,0 +1,24 @@
...
+ * limitations under the License.
+ */
+
+package com.google.accompanist.sample.swiperefresh
+
+import androidx.compose.runtime.Composable
+
+@Composable
+fun SwipeRefreshSample() {
+    // TODO
```
ricknout, line 23: Should we add this in? Could just copy `SwipeRefreshBasicSample`

swiperefresh/src/main/java/com/google/accompanist/swiperefresh/SwipeRefresh.kt
```diff
@@ -0,0 +1,327 @@
...
+import androidx.compose.ui.input.nestedscroll.nestedScroll
+import androidx.compose.ui.layout.onSizeChanged
+import androidx.compose.ui.unit.IntOffset
+import androidx.compose.ui.unit.Velocity
+import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.launch
+import kotlin.math.absoluteValue
+import kotlin.math.max
+
+private const val DragMultiplier = 0.5f
```
ricknout, line 50: Bit of a nit but wondering if we should align on const name casing (eg. `DragMultiplier` vs. `MAX_PROGRESS_ARC`)

sample/src/main/java/com/google/accompanist/sample/swiperefresh/DocsSamples.kt
chrisbanes: Good catch, done.

swiperefresh/src/main/java/com/google/accompanist/swiperefresh/SwipeRefresh.kt
chrisbanes: Yep, Compose uses pascal case so aligning on that

swiperefresh/src/main/java/com/google/accompanist/swiperefresh/Slingshot.kt
```diff
@@ -0,0 +1,80 @@
...
+
+internal data class Slingshot(
+    val offset: Int,
+    val startTrim: Float,
+    val endTrim: Float,
+    val rotation: Float,
+    val arrowScale: Float
+)
+
+internal const val MAX_PROGRESS_ARC = 0.8f
```
ricknout, line 80: nit: make this pascal case

chrisbanes: Done

docs/swiperefresh.md
```diff
@@ -0,0 +1,30 @@
+# Swipe Refresh layout for Jetpack Compose
+
+[![Maven Central](https://img.shields.io/maven-central/v/com.google.accompanist/accompanist-swiperefresh)](https://search.maven.org/search?q=g:com.google.accompanist)
+
+TODO
+
+## Usage
+
+``` kotlin
+// TODO
```
ricknout, line 10: We should add this in, right?

swiperefresh/src/main/java/com/google/accompanist/swiperefresh/SwipeRefresh.kt
```diff
@@ -0,0 +1,336 @@
...
+        mutatorMutex.mutate {
+            _indicatorOffset.animateTo(if (isRefreshing) refreshTrigger else 0f)
+        }
+    }
+
+    /**
+     * Dispatch scroll delta in pixels.
+     */
+    suspend fun dispatchRawDelta(delta: Float) {
+        mutatorMutex.mutate {
```
matvei-g, line 115: I guess this method should either not lock the mutex at all (since it's a "raw" dispatch), or lock the mutex with the `MutationPriority.UserInput` since we're reacting on the user input while dispatching raw deltas.

swiperefresh/src/main/java/com/google/accompanist/swiperefresh/SwipeRefresh.kt
```diff
@@ -0,0 +1,336 @@
...
+ * will be a [SwipeRefreshIndicator].
+ * @param content The content containing a scroll composable.
+ */
+@Composable
+fun SwipeRefresh(
+    state: SwipeRefreshState,
+    onRefresh: () -> Unit,
+    modifier: Modifier = Modifier,
+    swipeEnabled: Boolean = true,
+    indicator: @Composable SwipeRefreshIndicatorScope.() -> Unit = {
```
matvei-g, line 225: What do you think if you just pass State as a parameter for this slot? Or, alternatively, just do nothing and make a SwipeRefreshIndicator to accept a state from the SwipeRefresh itself?

I don't see a big usability reason to have this scope in general, but maybe I'm missing smth?

swiperefresh/src/main/java/com/google/accompanist/swiperefresh/SwipeRefresh.kt
```diff
@@ -0,0 +1,336 @@
...
+    val indicatorContentScope = remember(state) { SwipeRefreshIndicatorScopeImpl(state) }
+
+    Layout(
+        content = {
+            Box(Modifier.layoutId(LayoutContentTag)) { content() }
+            Box(Modifier.layoutId(LayoutIndicatorTag)) { indicatorContentScope.indicator() }
+        },
+        modifier = modifier
+            .nestedScroll(connection = nestedScrollConnection)
+            .clipToBounds()
```
matvei-g, line 264: Question: Do you really need to clip? It's ok I guess, but you might want to leave this for a scroll-container implementations (Modifier.verticalScroll, LazyColumn) and they already do it by default.

You might save a render node/ a view here + allow developers to not to clip if they have their own components that does the job slightly different.

Up to you though.

swiperefresh/src/main/java/com/google/accompanist/swiperefresh/SwipeRefreshIndicator.kt
```diff
@@ -0,0 +1,199 @@
...
+fun SwipeRefreshIndicator(
+    isRefreshing: Boolean,
+    offset: Float,
+    triggerOffset: Float,
+    modifier: Modifier = Modifier,
+    scale: Boolean = false,
+    arrowEnabled: Boolean = true,
+    backgroundColor: Color = MaterialTheme.colors.surface,
+    contentColor: Color = contentColorFor(backgroundColor),
+    shape: CornerBasedShape = CircleShape,
```
matvei-g, line 117: Consider using `MaterialTheme.shapes.small.copy(CornerSize(percent = 50))` here. Similarly to FAB, you want this to have cut corners if the theme is set this way

swiperefresh/src/main/java/com/google/accompanist/swiperefresh/SwipeRefreshIndicator.kt
```diff
@@ -0,0 +1,199 @@
...
+    sizes: SwipeRefreshIndicatorSizes = SwipeRefreshIndicatorDefaults.DefaultSizes,
```
matvei-g, line 118: I don't mind having this data class, but I'm curious how good for UX to allow such customisations? It might be that just a `largeIndication: Boolean = false` is enough and if people really want a fancier arc -- they could make their own indicator. 

The reason I suggest that is that I want those indicators to feel/look the same in other apps, so we could make it a but more complex to go out of the box, but still support large/small spec distinctions.

Curious what you think about it

matvei-g (commented):
Thanks Chris, looks amazing already.

A few high-level comments

docs/swiperefresh.md
chrisbanes: Yep! I was going to take a pass through docs at the end.

swiperefresh/src/main/java/com/google/accompanist/swiperefresh/SwipeRefresh.kt
chrisbanes: I think we need to allow apps to clip, as without it the indicator will drawn when it's 'off-screen'. Apps supplying their own `clipToBounds()` modifier is fine though I think. I'll add a comment for this.

swiperefresh/src/main/java/com/google/accompanist/swiperefresh/SwipeRefreshIndicator.kt
chrisbanes: Yeah agreed. Added a `largeIndication` param, and hid the data class for now.

swiperefresh/src/main/java/com/google/accompanist/swiperefresh/SwipeRefreshIndicator.kt
chrisbanes: Good call, done.

swiperefresh/src/main/java/com/google/accompanist/swiperefresh/SwipeRefresh.kt
chrisbanes: As discussed, done.

swiperefresh/src/main/java/com/google/accompanist/swiperefresh/SwipeRefresh.kt
chrisbanes: Went with `MutationPriority.UserInput`. Maybe I should rename this to `onDragDelta()` or something?

swiperefresh/src/main/java/com/google/accompanist/swiperefresh/SwipeRefresh.kt
```diff
@@ -0,0 +1,336 @@
...
+    }
+
+    private fun onScroll(available: Offset): Offset {
+        state.isSwipeInProgress = true
+
+        val drag = available.y * DragMultiplier
+        val distanceAvailable = when {
+            // If we're refreshing, we don't want the indicator to scroll below the refresh
+            // trigger
+            state.isRefreshing -> max(drag, -(state.indicatorOffset - state.refreshTrigger))
```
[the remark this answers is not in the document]
chrisbanes: Done. Completely re-wrote this, as I couldn't work out why I wrote it this way.

swiperefresh/src/main/java/com/google/accompanist/swiperefresh/SwipeRefresh.kt
matvei-g: yeah, might be worth it. It's not so raw anymore since you are acquiring mutex here :)

On a related note: do you really want it to be public?

matvei-g (approved):

ricknout (approved):

swiperefresh/src/main/java/com/google/accompanist/swiperefresh/SwipeRefresh.kt
chrisbanes: Done, and made `internal`

sample/src/main/java/com/google/accompanist/sample/swiperefresh/SwipeRefreshCustomSample.kt
```diff
@@ -0,0 +1,125 @@
...
+                backgroundColor = MaterialTheme.colors.surface,
+            )
+        },
+        modifier = Modifier.fillMaxSize()
+    ) {
+        var refreshing by remember { mutableStateOf(false) }
+
+        LaunchedEffect(refreshing) {
+            if (refreshing) {
+                delay(2000)
```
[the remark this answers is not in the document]
chrisbanes: Yeah totally agreed, added a comment.

docs/swiperefresh.md
chrisbanes: Will follow up on this in #324

PR #272, opened 2021-03-21
[Pager] Added a PagerIndicator composable

CaptnBlubber:
Adding a PagerIndicator Composable with flexible configuration.

I was not shure if the pager module itself is the right place or if it should be an individual module. 

Demo:

https://user-images.githubusercontent.com/878307/111903106-db7eea80-8a40-11eb-8216-759b482283d6.mp4

### DIFF
--- a/pager/src/main/java/com/google/accompanist/pager/PagerIndicator.kt
+++ b/pager/src/main/java/com/google/accompanist/pager/PagerIndicator.kt
@@ -1,0 +1,74 @@
+/*
+ * Copyright 2021 The Android Open Source Project
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.google.accompanist.pager
+
+import androidx.compose.foundation.background
+import androidx.compose.foundation.layout.Arrangement
+import androidx.compose.foundation.layout.Box
+import androidx.compose.foundation.layout.Row
+import androidx.compose.foundation.layout.offset
+import androidx.compose.foundation.layout.size
+import androidx.compose.foundation.shape.CircleShape
+import androidx.compose.runtime.Composable
+import androidx.compose.ui.Alignment
+import androidx.compose.ui.Modifier
+import androidx.compose.ui.graphics.Color
+import androidx.compose.ui.graphics.Shape
+import androidx.compose.ui.unit.Dp
+import androidx.compose.ui.unit.dp
+
+@ExperimentalPagerApi
+@Composable
+fun PagerIndicator(
+    modifier: Modifier = Modifier,
+    pagerState: PagerState,
+    color: Color,
+    indicatorColor: Color,
+    size: Dp = 8.dp,
+    spacing: Dp = size,
+    shape: Shape = CircleShape,
+    indicatorShape: Shape = shape
+) {
+    Box(modifier = modifier, contentAlignment = Alignment.CenterStart) {
+        Row(
+            horizontalArrangement = Arrangement.spacedBy(spacing),
+            verticalAlignment = Alignment.CenterVertically
+        ) {
+            repeat(pagerState.pageCount) {
+                Box(
+                    modifier = Modifier
+                        .size(size)
+                        .background(
+                            color = color,
+                            shape = shape,
+                        )
+                )
+            }
+        }
+
+        Box(
+            modifier = Modifier
+                .offset(x = ((spacing + size) * pagerState.currentPage + (spacing + size) * pagerState.currentPageOffset))
+                .size(size)
+                .background(
+                    color = indicatorColor,
+                    shape = indicatorShape,
+                )
+
+        )
+    }
+}
--- a/pager/src/main/java/com/google/accompanist/pager/PagerIndicator.kt
+++ b/pager/src/main/java/com/google/accompanist/pager/PagerIndicator.kt
@@ -31,16 +31,66 @@ import androidx.compose.ui.graphics.Shape
 import androidx.compose.ui.unit.Dp
 import androidx.compose.ui.unit.dp
 
+/**
+ * Contains default values used for [PagerIndicator].
+ */
+object PagerDefaults {
+    /**
+     * Default spacing used for [PagerIndicator].
+     */
+    val Spacing = 8.dp
+
+    /**
+     * Default indicator size used for [PagerIndicator].
+     */
+    val Size = 8.dp
+
+    /**
+     * Default shape used for [PagerIndicator].
+     */
+    val Shape = CircleShape
+
+    /**
+     * Default Indicator color value used for [PagerIndicator].
+     */
+    val IndicatorColor = Color.White
+
+    /**
+     * Default alpha value used for generating the inactive indicators color for [PagerIndicator].
+     */
+    val InactiveIndicatorColorAlpha = 0.6f
+}
+
+/**
+ * A Indicator for a [Pager] representing the currently active page and total pages using a [Shape].
+ *
+ * This Element allows the setting of both the [shape] and [indicatorShape], which defines how the
+ * Indicator is visually represented. By default Indicators are represented as [CircleShape].
+ * When changing the [shape] by default the [indicatorShape] adjusts accordingly. If you want the
+ * active page indicator to have a different appearance override [indicatorShape].
+ *
+ * @sample com.google.accompanist.sample.pager.PagerIndicatorSample
+ *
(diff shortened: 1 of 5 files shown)


sample/src/main/AndroidManifest.xml
```diff
@@ -1,5 +1,4 @@
-<?xml version="1.0" encoding="utf-8"?>
```
chrisbanes, line 1: Revert this diff please

pager/src/main/java/com/google/accompanist/pager/PagerIndicator.kt
```diff
@@ -0,0 +1,74 @@
...
+import androidx.compose.ui.Alignment
+import androidx.compose.ui.Modifier
+import androidx.compose.ui.graphics.Color
+import androidx.compose.ui.graphics.Shape
+import androidx.compose.ui.unit.Dp
+import androidx.compose.ui.unit.dp
+
+@ExperimentalPagerApi
+@Composable
+fun PagerIndicator(
```
chrisbanes, line 36: Needs some kdoc

pager/src/main/java/com/google/accompanist/pager/PagerIndicator.kt
```diff
@@ -0,0 +1,74 @@
...
+    modifier: Modifier = Modifier,
+    pagerState: PagerState,
+    color: Color,
```
chrisbanes, line 39: Maybe default `color` to `indicatorColor.copy(alpha = 0.6f)` ? Extract the `0.6f` to a const too

pager/src/main/java/com/google/accompanist/pager/PagerIndicator.kt
```diff
@@ -0,0 +1,74 @@
...
+import androidx.compose.ui.graphics.Color
+import androidx.compose.ui.graphics.Shape
+import androidx.compose.ui.unit.Dp
+import androidx.compose.ui.unit.dp
+
+@ExperimentalPagerApi
+@Composable
+fun PagerIndicator(
+    modifier: Modifier = Modifier,
+    pagerState: PagerState,
```
chrisbanes, line 38: Should be the first parm

pager/src/main/java/com/google/accompanist/pager/PagerIndicator.kt
```diff
@@ -0,0 +1,74 @@
...
+import androidx.compose.ui.Modifier
+import androidx.compose.ui.graphics.Color
+import androidx.compose.ui.graphics.Shape
+import androidx.compose.ui.unit.Dp
+import androidx.compose.ui.unit.dp
+
+@ExperimentalPagerApi
+@Composable
+fun PagerIndicator(
+    modifier: Modifier = Modifier,
```
chrisbanes, line 37: Move this down, it should be the first parameter without a default value

chrisbanes (commented):
Looks great, thanks!

I had the same thought about whether this should be in a separate module. I think this is fine for now, and we can extract later. R8 will strip this if it's not used anyway.

pager/src/main/java/com/google/accompanist/pager/PagerIndicator.kt
CaptnBlubber: I Indeed had this previously but scraped it. Now I have added it to a new Object called PagerDefaults alongside with some other Defaults.

CaptnBlubber:
Adjusted all requested changes.

chrisbanes (approved):

chrisbanes:
Perfect, thanks for the contribution!

pager/src/main/java/com/google/accompanist/pager/PagerIndicator.kt
```diff
@@ -0,0 +1,124 @@
...
+ * This Element allows the setting of both the [shape] and [indicatorShape], which defines how the
+ * Indicator is visually represented. By default Indicators are represented as [CircleShape].
+ * When changing the [shape] by default the [indicatorShape] adjusts accordingly. If you want the
+ * active page indicator to have a different appearance override [indicatorShape].
+ *
+ * @sample com.google.accompanist.sample.pager.PagerIndicatorSample
+ *
+ * @param pagerState the state object of your [Pager] to be used to observe the list's state.
+ * @param modifier the modifier to apply to this layout.
+ * @param indicatorColor the color of the active Page indicator
```
RinOrz, line 76: Would names like activeColor, disabledColor be better?

pager/src/main/java/com/google/accompanist/pager/PagerIndicator.kt
```diff
@@ -0,0 +1,124 @@
...
+
+    /**
+     * Default shape used for [PagerIndicator].
+     */
+    val Shape = CircleShape
+
+    /**
+     * Default Indicator color value used for [PagerIndicator].
+     */
+    val IndicatorColor = Color.White
```
RinOrz, line 56: Should it be replaced with `LocalContentColor.current`? White is not friendly under the light theme

pager/src/main/java/com/google/accompanist/pager/PagerIndicator.kt
```diff
@@ -0,0 +1,124 @@
...
+
+    /**
+     * Default alpha value used for generating the inactive indicators color for [PagerIndicator].
+     */
+    val InactiveIndicatorColorAlpha = 0.6f
```
RinOrz, line 61: Maybe use `ContentAlpha.disabled`

RinOrz (commented):
Great contribution, but I have a few questions during use

pager/src/main/java/com/google/accompanist/pager/PagerIndicator.kt
CaptnBlubber: I originally Planned something like this and thought of both LocalContentColor or directly MaterialTheme. The issue is that both solutions rely on the material dependency wich is currently not part of the pager module and i did not want to pull that dependency in. I think there is a discussion needed for that and maybe the indicator needs to become a seperate module.

pager/src/main/java/com/google/accompanist/pager/PagerIndicator.kt
CaptnBlubber: Same as with the color, this is part of material wich is currently not pulled in

pager/src/main/java/com/google/accompanist/pager/PagerIndicator.kt
CaptnBlubber: I had similar thoughts already although "disabled" might not be correct. Also not sure if "inactive" is a better replacement

CaptnBlubber:
> Great contribution, but I have a few questions during use.
> 
> Or even change the `size` to `width` and `height`? Because sometimes we need to display it as a rectangle
> 

Do you have a example of such an rectangle as design so i can get a better picture of the requirement.

It should be easy to adapt that concept however

pager/src/main/java/com/google/accompanist/pager/PagerIndicator.kt
chrisbanes: Yep, this is one of the reasons why I was thinking of extracting the indicator to a different module `pager-indicators` or something. We could make that depend on `compose-material`. I'll do it quickly now.

RinOrz:
> Do you have a example of such an rectangle as design so I can get a better picture of the requirement?
Yes this is the indicator i need
image

pager/src/main/java/com/google/accompanist/pager/PagerIndicator.kt
CaptnBlubber: Do you think that it might make sense to have a generic pager indicator like the current ones and preconfigured ones like a DotPagerIndicator ? I could create another pr for that.

I would in this case remove the default shape from the defaults so that indicator always requires a shape with size and DotPagerIndicator just accepts the size and calls PagerIndicator with the CircleShape

CaptnBlubber:
I will file another PR that works with width and height so your requirements should work

RinOrz:
@CaptnBlubber In addition, I even think that the indicator cannot limit the default switching animation, because some people may prefer some cool animations. Would it be a good choice to provide a transform interface? Just like https://github.com/ogaclejapan/SmartTabLayout/blob/master/library/src/main/java/com/ogaclejapan/smarttablayout/SmartTabIndicationInterpolator.java#L23, but this may also lead to too "flexible"

chrisbanes:
Update PR: #279

I'm not sure that I want to make the the indicator composables more complex. Indicators are not that difficult to write, so I think it's OK to expect devs to write whatever they need.

We could offer different indicator composables in the new module, but I don't think we should be creating a single generic solution which tries to work for everything.

RinOrz:
> I'm not sure that I want to make the current indicator composables any more complex. Indicators are not that difficult to write, so I think it's OK to expect devs to write whatever they need.
> 
> We could offer different indicator composables in the new module, but I don't think we should be creating a single generic solution which tries to work for everything.

You are right, but is it better to avoid more wheels?

chrisbanes:
> but is it better to avoid more wheels?

I don't believe so with Compose. Most things are easier to write, so there's less need to create super generic solutions.

RinOrz:
> I don't believe so with Compose. Most things are easier to write, so there's less need to create super generic solutions.

Okay, BTW, here is a question, when are **Pager** and **Indicator** available in maven?

chrisbanes:
When the next release happens, which should be tomorrow.

PR #277, opened 2021-03-22
[Pager] Add PagerState.calculateCurrentOffsetForPage() ext-fun 

chrisbanes:
Also added a sample which adds item transitions, such as a fade-in/out and light parallax effect.

### DIFF
--- a/pager/src/main/java/com/google/accompanist/pager/Pager.kt
+++ b/pager/src/main/java/com/google/accompanist/pager/Pager.kt
@@ -349,6 +349,11 @@ interface PagerScope : BoxScope {
      * Returns the current selected page offset
      */
     val currentPageOffset: Float
+
+    /**
+     * The current scroll position, as a float value between `0 until pageSize`
+     */
+    val scrollPosition: Float
 }
 
 @ExperimentalPagerApi
@@ -361,4 +366,7 @@ private class PagerScopeImpl(
 
     override val currentPageOffset: Float
         get() = state.currentPageOffset
+
+    override val scrollPosition: Float
+        get() = state.scrollPosition
 }
--- a/pager/src/main/java/com/google/accompanist/pager/Pager.kt
+++ b/pager/src/main/java/com/google/accompanist/pager/Pager.kt
@@ -349,11 +349,6 @@ interface PagerScope : BoxScope {
      * Returns the current selected page offset
      */
     val currentPageOffset: Float
-
-    /**
-     * The current scroll position, as a float value between `0 until pageSize`
-     */
-    val scrollPosition: Float
 }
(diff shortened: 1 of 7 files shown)


pager/src/main/java/com/google/accompanist/pager/PagerState.kt
```diff
@@ -95,9 +95,9 @@ class PagerState(
     internal var pageSize by mutableStateOf(0)
 
     /**
-     * The page position, as a float value between `0 until pageSize`
+     * The current scroll position, as a float value between `0 until pageSize`
      */
-    private val globalPosition: Float
+    val scrollPosition: Float
```
nickbutcher, line 100: `scrollPosition` seems to imply a px value to me but Im struggling to come up with better alternatives. `currentPosition`? `absolutePosition`?

chrisbanes: `scrollPosition` isn't great, but I can't think of anything better either. `scrollPosition` works ok, it's just the assumption of the unit. Considering now other part of the API uses `px`, I think the API name is ok

nickbutcher: SG.

nickbutcher (approved):

pager/src/main/java/com/google/accompanist/pager/Pager.kt
```diff
@@ -362,3 +362,17 @@ private class PagerScopeImpl(
     override val currentPageOffset: Float
         get() = state.currentPageOffset
 }
+
+/**
+ * Calculate the offset for the given [page] from the current scroll position. This is useful
+ * when using the scroll position to apply effects or animations to items.
+ *
+ * The returned offset can positive or negative, depending on whether which direction the [page] is
```
nickbutcher, line 370: Document the range `[-1f .. 1f]` right?

nickbutcher: Oh wait no, for a 'distant' page this could be outside this range. Are consumers expected to clamp this to be useful for effects?

Is this a smell that `currentPageOffset` doesn't do what consumers expect?

chrisbanes: I can't make a call on clamping, because we don't know what size the items are (there could be 1, but there could be 5). Each effect is different too, as it could span across all items, or just the center one. 

> Is this a smell that currentPageOffset doesn't do what consumers expect?

Maybe. I don't want to make too many drastic changes now so close to release. This is why I've done this as an ext-fun which can be easily deprecated and removed.

PR #278, opened 2021-03-22
[Pager] Add PagerState.pageChanges

chrisbanes:
Allow apps to collect any current page changes via a `Flow`. Typical usage looks like the following:

``` kotlin
val pagerState = rememberPagerState(pageCount = 10)

LaunchedEffect(pagerState) {
    // Collect from the PageState's pageChanges, which emits when the
    // current page has changed
    pagerState.pageChanges.collect { page ->
        AnalyticsService.sendPageSelectedEvent(page)
    }
}
```

### DIFF
--- a/pager/src/main/java/com/google/accompanist/pager/PagerState.kt
+++ b/pager/src/main/java/com/google/accompanist/pager/PagerState.kt
@@ -40,6 +40,11 @@ import androidx.compose.runtime.saveable.Saver
 import androidx.compose.runtime.saveable.listSaver
 import androidx.compose.runtime.saveable.rememberSaveable
 import androidx.compose.runtime.setValue
+import androidx.compose.runtime.snapshotFlow
+import kotlinx.coroutines.flow.Flow
+import kotlinx.coroutines.flow.distinctUntilChanged
+import kotlinx.coroutines.flow.filter
+import kotlinx.coroutines.flow.map
 import kotlin.math.absoluteValue
 import kotlin.math.floor
 import kotlin.math.roundToInt
@@ -436,3 +441,19 @@ class PagerState(
         )
     }
 }
+
+/**
+ * A flow which emits the [PagerState.currentPage] if it changes due a scroll
+ * has completed.
+ *
+ * This flow is not meant to be used for updating any UI within the attached [HorizontalPager]
+ * or [VerticalPager]. For that use-case, you should read
+ * [PagerScope.currentPage] and [PagerScope.currentPageOffset] from the content scope.
+ */
+@ExperimentalPagerApi
+inline val PagerState.pageChangedFlow: Flow<Int>
+    get() = snapshotFlow { isScrollInProgress }
+        // Only emit when the scroll has finished
+        .filter { isScrollInProgress -> !isScrollInProgress }
+        .map { currentPage }
+        .distinctUntilChanged()
--- a/pager/src/main/java/com/google/accompanist/pager/PagerState.kt
+++ b/pager/src/main/java/com/google/accompanist/pager/PagerState.kt
@@ -443,7 +443,7 @@ class PagerState(
 }
 
 /**
- * A flow which emits the [PagerState.currentPage] as it changes due after completed scrolls.
+ * A flow which emits the [PagerState.currentPage] as it changes due to a scroll completing.
  *
  * This flow is not meant to be used for updating any UI within the attached [HorizontalPager]
  * or [VerticalPager]. For that use-case, you should read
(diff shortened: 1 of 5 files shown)


pager/src/main/java/com/google/accompanist/pager/PagerState.kt
```diff
@@ -436,3 +441,21 @@ class PagerState(
...
+ * has completed.
+ *
+ * This flow is not meant to be used for updating any UI within the attached [HorizontalPager]
+ * or [VerticalPager]. For that use-case, you should read
+ * [PagerScope.currentPage] and [PagerScope.currentPageOffset] from the content scope.
+ *
+ * @sample com.google.accompanist.sample.pager.PageChangedFlowSample
+ */
+@ExperimentalPagerApi
+inline val PagerState.pageChangedFlow: Flow<Int>
```
nickbutcher, line 456: Don't love encoding the type in the name. What about `pageChanges`? Looking at [`InteractionSource.interactions`](https://cs.android.com/androidx/platform/frameworks/support/+/androidx-main:compose/foundation/foundation/src/commonMain/kotlin/androidx/compose/foundation/interaction/InteractionSource.kt;l=74) for inspiration.

pager/src/main/java/com/google/accompanist/pager/PagerState.kt
```diff
@@ -436,3 +441,21 @@ class PagerState(
         )
     }
 }
+
+/**
+ * A flow which emits the [PagerState.currentPage] if it changes due a scroll
+ * has completed.
```
JoseAlcerreca, line 447: "due to scroll completing" or similar

pager/src/main/java/com/google/accompanist/pager/PagerState.kt
```diff
@@ -436,3 +441,21 @@ class PagerState(
...
+ *
+ * @sample com.google.accompanist.sample.pager.PageChangedFlowSample
+ */
+@ExperimentalPagerApi
+inline val PagerState.pageChangedFlow: Flow<Int>
+    get() = snapshotFlow { isScrollInProgress }
+        // Only emit when the scroll has finished
+        .filter { isScrollInProgress -> !isScrollInProgress }
+        .map { currentPage }
+        .distinctUntilChanged()
```
nickbutcher, line 461: Is this needed? Shouldn't the filter enforce that emissions only get here when scrolling finishes?

pager/src/main/java/com/google/accompanist/pager/PagerState.kt
chrisbanes: Good call, done.

pager/src/main/java/com/google/accompanist/pager/PagerState.kt
chrisbanes: Not necessarily. A scroll could finish and the pager snaps back to the original page.

pager/src/main/java/com/google/accompanist/pager/PagerState.kt
chrisbanes: Thanks, done.

pager/src/main/java/com/google/accompanist/pager/PagerState.kt
```diff
@@ -436,3 +441,20 @@ class PagerState(
...
+ * or [VerticalPager]. For that use-case, you should read
+ * [PagerScope.currentPage] and [PagerScope.currentPageOffset] from the content scope.
+ *
+ * @sample com.google.accompanist.sample.pager.PageChangesSample
+ */
+@ExperimentalPagerApi
+inline val PagerState.pageChanges: Flow<Int>
+    get() = snapshotFlow { isScrollInProgress }
+        // Only emit when the scroll has finished
+        .filter { isScrollInProgress -> !isScrollInProgress }
```
nickbutcher, line 458: would `filterNot { it } // only emit…` be clearer?

pager/src/main/java/com/google/accompanist/pager/PagerState.kt
nickbutcher: True, true.

nickbutcher (approved):
Thanks for the name change. Added one small suggestion, otherwise LGTM.

pager/src/main/java/com/google/accompanist/pager/PagerState.kt
chrisbanes: Ah good call, always forget about `filterNot()`

pager/src/main/java/com/google/accompanist/pager/PagerState.kt
JoseAlcerreca: How about making this a StateFlow with stateIn? You wouldn't need the distinctUntilChanged and since this is a PagerState it sounds appropriate (but I don't know the details of how it's consumed)

chrisbanes: Happy to change it, but not sure what the benefit would be? Would the type be `StateFlow` or `Flow`?

PR #279, opened 2021-03-22
[Pager] Move indicators to pager-indicators module

chrisbanes:
This allows us to make the indicator module depend on Compose Material, and provide better default values for the indicators. Also simplified the parameters to make them more consistent with other Compose layouts.


pager-indicators/src/main/java/com/google/accompanist/pager/PagerIndicator.kt
```diff
@@ -0,0 +1,169 @@
...
+ */
+@ExperimentalPagerApi
+@Composable
+fun HorizontalPagerIndicator(
+    pagerState: PagerState,
+    modifier: Modifier = Modifier,
+    activeColor: Color = LocalContentColor.current.copy(alpha = LocalContentAlpha.current),
+    inactiveColor: Color = activeColor.copy(ContentAlpha.disabled),
+    indicatorWidth: Dp = 8.dp,
+    indicatorHeight: Dp = 8.dp,
```
CaptnBlubber, line 66: Maybe use `indicatorWidth` here as default value so its symetric by default?

CaptnBlubber (commented):
Looks good to me thanks for migrating it to a seperate Module and applying the default color changes

pager-indicators/src/main/java/com/google/accompanist/pager/PagerIndicator.kt
chrisbanes: Good call, done.

FilippoVigani:
Sorry to commend on an old issue, but I think it would make sense to call the library `accompanist-pager-indicators-material` since it depends on Compose Material. The same way the placeholder has the library `accompanist-placeholder-material`.

PR #287, opened 2021-03-25
Pager parameter to disable scrolling by hand

ln-12:
This commit adds a parameter to the Pager composable to disable scrolling by hand. If disabled, PagerState.animateScrollToPage() can be used to programmatically switch to a different page.

### DIFF
--- a/pager/src/main/java/com/google/accompanist/pager/Pager.kt
+++ b/pager/src/main/java/com/google/accompanist/pager/Pager.kt
@@ -138,6 +138,7 @@ fun HorizontalPager(
     state: PagerState,
     modifier: Modifier = Modifier,
     reverseLayout: Boolean = false,
+    isDraggable: Boolean = true,
     @IntRange(from = 1) offscreenLimit: Int = 1,
     flingBehavior: FlingBehavior = PagerDefaults.defaultPagerFlingConfig(state),
     verticalAlignment: Alignment.Vertical = Alignment.CenterVertically,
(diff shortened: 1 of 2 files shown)


pager/src/main/java/com/google/accompanist/pager/Pager.kt
```diff
@@ -245,12 +250,16 @@ internal fun Pager(
         selectableGroup()
     }
 
-    val scrollable = Modifier.scrollable(
-        orientation = if (isVertical) Orientation.Vertical else Orientation.Horizontal,
-        flingBehavior = flingBehavior,
-        reverseDirection = reverseDirection,
-        state = state,
-    )
+    val scrollable = if(isDraggable) {
```
chrisbanes, line 253: No need to do this: `scrollable()` has an `enabled` parameter.

ln-12:
@googlebot I signed it!

pager/src/main/java/com/google/accompanist/pager/Pager.kt
```diff
@@ -138,6 +138,7 @@ fun HorizontalPager(
     state: PagerState,
     modifier: Modifier = Modifier,
     reverseLayout: Boolean = false,
+    isDraggable: Boolean = true,
```
chrisbanes, line 141: I think this should be called `dragEnabled`. Same for the others.

chrisbanes (commented):
Thanks!

Once you've done the changes, you'll need to run `./gradlew spotlessApply apiDump`. This should also really have a test.

ln-12:
Thanks for your suggestions, I performed the changes and ran the command.

chrisbanes (approved):

chrisbanes:
Thanks for the update!

PR #299, opened 2021-03-31
[Pager] Update currentPage to act as the selected page

### LINKED ISSUE #280: [Pager] Dragging backwards instantly updates the state's current page
## Describe the bug
The current page of the pager is updated incorrectly while there's an active drag to the previous page, when dragging to the next page it works fine.

## Expected behavior
The currentPage should only be updated when the finger is released/the drag is settling.

## Environment:
 - Android OS version: 10
 - Device: OnePlus 5
 - Accompanist version: 0.6.3-SNAPSHOT (gradle sync done a few minutes ago)

## Additional context

Watch the tab indicator being updated when dragging backwards:

https://user-images.githubusercontent.com/3521738/112030009-1749ab00-8b3a-11eb-8c01-8ab0f8e7468f.mp4

This is the code used in the video above:

```kt
@OptIn(ExperimentalPagerApi::class)
@Composable
fun PagerDragBug() {
  val state = rememberPagerState(pageCount = 3, initialPage = 1)
  val pages = listOf(
    "A" to Color.Gray,
    "B" to Color.Black,
    "C" to Color.Gray
  )

  Column {
    TabRow(selectedTabIndex = state.currentPage) {
      pages.forEachIndexed { idx, page ->
        Tab(
          selected = state.currentPage == idx,
          text = { Text(page.first) },
          onClick = {}
        )
      }
    }
    HorizontalPager(state = state) { page ->
      Box(modifier = Modifier.fillMaxSize().background(pages[page].second))
    }
  }
}
```

I'm using your `Modifier.pagerTabIndicatorOffset` which is better to sync the tab indicator with the pager, but I've used this simpler approach to reproduce the issue. My actual issue is that I have a `LaunchedEffect(state.currentPage)` to notify my ViewModel when the active page changes, and this causes some glitches.

chrisbanes:
Reading the selected page is a common use for pagers, especially when used with things like Tabs. This PR clarifies what currentPage and currentPageOffset to mean the 'selected page', and thus only get updated after a scroll has completed.

This allows us to tidy up the Tab offset modifier and make it a provided function in the library.

Internally, we've now separated the 'selected page' from the internal 'layout page' (and offset), which allows us to set the selected page only once a scroll has finished.

Fixes #280

### DIFF
--- a/docs/pager.md
+++ b/docs/pager.md
@@ -131,15 +131,11 @@ HorizontalPager(state = pagerState) { page ->
 
 ## Reacting to page changes
 
-From reading above, you might be thinking that reading [`PagerState.currentPage`][currentpage-api] is a good way to know when the selected page changes. Unfortunately `currentPage` does not tell you the currently selected page, but instead tells you which page is (mostly) displayed on screen.
-
-To know when the selected page changes, you can use the [`pageChanges`](../api/pager/pager/com.google.accompanist.pager/page-changes.html) flow:
+The [`PagerState.currentPage`][currentpage-api] property is updated whenever the selected page changes. You can use the `snapshowFlow` function to observe changes in a flow:
 
 ``` kotlin
-import com.google.accompanist.pager.pageChanges
-
 LaunchedEffect(pagerState) {
-    pagerState.pageChanges.collect { page ->
+    snapshotFlow { pagerState.currentPage }.collect { page ->
         // Selected page has changed...
     }
 }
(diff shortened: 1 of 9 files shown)


nickbutcher:
I'm not sure about this change. Having both a `currentPage` and a `selectedPage` feels like a smell. Should the `currentPage` become a private val used only for layout and only expose the `selectedPage`? Also I'd expect to be able to write to this property to instantly change page (unlike `animateScrollToPage`) so exposing it as a state rather than a `var` feels wrong. WDYT?

PR #301, opened 2021-03-31
[Pager] Use passed in width for active indicator

brandonrisell:
I've been using the paging indicator library and noticed that the width was being used for the page indicator containers, but not the active indicator.

I updated it to use the passed in width for the active indicator, instead of the height.


chrisbanes (approved):
Woops, thanks!

PR #308, opened 2021-04-05
Implement swipe refresh features

ricknout:
This implements most of the main swipe refresh features seen in the View version—by doing a best effort adaption of the _very_ messy code—while adding a few new features made easier by Compose.

Basic behavior:

https://user-images.githubusercontent.com/12491712/113629743-1e5ad800-9667-11eb-9f0e-2197977bebf2.mp4

Dark theme:

https://user-images.githubusercontent.com/12491712/113629769-261a7c80-9667-11eb-8633-b72a88067b9d.mp4

Scale:

https://user-images.githubusercontent.com/12491712/113629902-56621b00-9667-11eb-9feb-0a4f37da92b6.mp4

Default and large sizes:

https://user-images.githubusercontent.com/12491712/113629921-611cb000-9667-11eb-88d0-f8517f2edd92.mp4

Custom colors, shape, and elevation:

https://user-images.githubusercontent.com/12491712/113629955-6da10880-9667-11eb-95ef-67e1074343d0.mp4

Some things to note:
- Some features from the View version are missing (custom offset triggers and color lists mainly)
- It uses a `Surface` instead of the previous custom implementation and gets around alpha issues by animating the elevation based on refresh state
- It updates the custom painter that draws the arc/arrow while dragging
- It crossfades between the painter and a `CircularProgressDrawable` in refreshing state (the drawing logic was getting a bit crazy and it felt more maintainable this way...)
- The code organization and general structure probably needs a second pass to refactor appropriately


swiperefresh/src/main/java/com/google/accompanist/swiperefresh/IndicatorSize.kt
```diff
@@ -0,0 +1,69 @@
...
+/**
+ * A enumerated class to encapsulate details of different indicator sizes.
+ *
+ * @param size The overall size of the indicator.
+ * @param arcRadius The radius of the arc.
+ * @param strokeWidth The width of the arc stroke.
+ * @param arrowWidth The width of the arrow.
+ * @param arrowHeight The height of the arrow.
+ */
+enum class IndicatorSize(
```
chrisbanes, line 31: I think this could be a normal data class instead, with `Default` and `Large` being constants in the companion object. This allows devs to provide their own values if they wish.

swiperefresh/src/main/java/com/google/accompanist/swiperefresh/Slingshot.kt
```diff
@@ -0,0 +1,74 @@
...
+
+/**
+ * A utility function that calculates various aspects of 'slingshot' behavior.
+ * Adapted from SwipeRefreshLayout#moveSpinner method.
+ *
+ * @param offsetY The current y offset.
+ * @param maxOffsetY The max y offset.
+ * @param height The height of the item to slingshot.
+ */
+internal fun calculateSlingshot(
```
chrisbanes, line 32: I think we can probably calculate this using a spring, but looks good for now.

swiperefresh/src/main/java/com/google/accompanist/swiperefresh/SwipeRefreshIndicator.kt
```diff
@@ -14,107 +14,125 @@
...
-                else -> 0f
-            }
-        }
-        .background(color = bgColor, shape = CircleShape)
-        .clip(CircleShape)
-
-    Box(modifier) {
-        val ringPainter = remember {
-            ProgressRingPainter()
+    val animatedElevation by animateDpAsState(
```
chrisbanes, line 59: Why `animateDpAsState()` when this is just using `snap()`? This could just be an `if` (I think)

swiperefresh/src/main/java/com/google/accompanist/swiperefresh/SwipeRefreshIndicator.kt
```diff
@@ -14,107 +14,125 @@
...
+        val slingshot = calculateSlingshot(
+            offsetY = offset,
+            maxOffsetY = maxOffset,
+            height = with(LocalDensity.current) { size.size.roundToPx() }
         )
+        painter.startTrim = slingshot.startTrim
+        painter.endTrim = slingshot.endTrim
+        painter.rotation = slingshot.rotation
+        painter.arrowScale = slingshot.arrowScale
+        Crossfade(
```
chrisbanes, line 99: Add a comment here for what the crossfade is doing

chrisbanes (commented):
Thanks! A few comments. Also needs the api files updating: `./gradlew apiDump`

swiperefresh/src/main/java/com/google/accompanist/swiperefresh/Slingshot.kt
ricknout: Yeah that was my thinking and would be ideal to avoid having to maintain this code. I could give it a go in a follow-up PR

swiperefresh/src/main/java/com/google/accompanist/swiperefresh/SwipeRefreshIndicator.kt
ricknout: Whoops, yeah agree

swiperefresh/src/main/java/com/google/accompanist/swiperefresh/Slingshot.kt
chrisbanes: This isn't critical. Maybe just add a TODO comment so we come back to it

chrisbanes (approved):

PR #312, opened 2021-04-07
Expose animation spec in animateScrollToPage()

ln-12:
This change exposes the animationSpec parameter from `PagerState.animateToPage()` (which is private) to `PagerState.animateScrollToPage()` so that the animation can be changed as needed.


chrisbanes (approved):
Looks good to me, thanks!

PR #316, opened 2021-04-09
More inset aware component

Omico:
Enhancement for https://github.com/google/accompanist/issues/207


Omico:
Crash, because of `measureInsetAwareState` function in InsetAwareScaffold.

Close and wait for #409.

PR #321, opened 2021-04-10
[Image] LoadPainter

chrisbanes:
This refactors the image loading libraries again, this time to a `Painter` instance, retrieved through `rememberLoadPainter()`, This allows us to remove our layout composable, and rely solely on Compose Foundation's `Image()` composable.

## Tasks

- [x] Tests passing
- [x] Samples updated
- [ ] API names (TBD at the moment)
- [x] Update `ReplaceWith`s
- [x] kdoc on new API
- [ ] Docs


coil/src/main/java/com/google/accompanist/coil/Coil.kt
```diff
@@ -60,6 +62,30 @@ object CoilImageStateDefaults {
     }
 }
 
+@Suppress("NOTHING_TO_INLINE")
+@Composable
+inline fun rememberCoilPainter(
```
andkulikov, line 67: it is not a good idea to mark such complex functions as inline. it will add a lot of code into every usage. 
so imagine the usage like this
rememberCoilPainter("url")
in this case all not specified params still need to be resolved so their default values are inlined into every usage. there is also some compose generated code around all the composables used in default values like for
CoilImageStateDefaults.defaultImageLoader()
Compose actually has to generated observation groups and so on.
inlining composable functions sometimes useful in order to save on lambda allocations, but most of the times it is not needed

imageloading-core/src/main/java/com/google/accompanist/imageloading/Image.kt
```diff
@@ -31,162 +31,50 @@ import androidx.compose.runtime.rememberCoroutineScope
...
-fun <R : Any> Image(
+fun <R : Any> rememberLoadPainter(
     state: ImageState<R>,
-    contentDescription: String?,
-    modifier: Modifier = Modifier,
-    alignment: Alignment = Alignment.Center,
-    contentScale: ContentScale = ContentScale.Fit,
-    colorFilter: ColorFilter? = null,
     fadeIn: Boolean = false,
     fadeInDurationMs: Int = DefaultTransitionDuration,
```
andkulikov, line 71: it is usually preferred to not use the internal vals as a default values as users can't really use it. imagine someone wants to write something like this
rememberLoadPainter(fadeInDurationMs = if (haveCustomDuration) customDuration else DefaultTransitionDuration)
and they can't do so. we usually or make a public constant or just inline the value. 
I am also curious is 1000 really a good default? isnt it too long?

imageloading-core/src/main/java/com/google/accompanist/imageloading/ImageState.kt
```diff
@@ -0,0 +1,109 @@
...
+import kotlin.coroutines.cancellation.CancellationException
+
+/**
+ * A state base class that can be hoisted to control image loads for [rememberLoadPainter].
+ *
+ * @param shouldRefetchOnSizeChange Initial value for [shouldRefetchOnSizeChange].
+ */
+@Stable
+abstract class ImageState<R : Any>(
+    shouldRefetchOnSizeChange: (currentState: ImageLoadState, size: IntSize) -> Boolean,
```
andkulikov, line 35: I am curious if we really still need ImageState public api at all? it was reasonable as a state for composable but now when we have a Painter maybe we can merge them, Painter is already a stateful thing. I would try to move shouldRefetchOnSizeChange param to rememberLoadPainter(). 
if we want to keep public val loadState we can make LoadPainter class public, introduce it there and return it from rememberLoadPainter().
wdyt?

andkulikov (commented):
I like the simplicity of the resulting api!

coil/src/main/java/com/google/accompanist/coil/Coil.kt
chrisbanes: Done.

imageloading-core/src/main/java/com/google/accompanist/imageloading/ImageState.kt
chrisbanes: Yeah, PTAL. I'll probably finesse the API names, but the general structure feels good to me.

glide/src/main/java/com/google/accompanist/glide/Glide.kt
```diff
@@ -66,175 +63,117 @@ object GlideImageStateDefaults {
...
+    requestBuilder: (RequestBuilder<Drawable>.(size: IntSize) -> RequestBuilder<Drawable>)? = null,
     fadeIn: Boolean = false,
     fadeInDurationMs: Int = 1000,
     @DrawableRes previewPlaceholder: Int = 0,
-): Painter = rememberLoadPainter(
-    state = rememberGlideImageState(
-        data = data,
-        requestManager = requestManager,
+): LoadPainter<Any> {
+    val updatedRequestBuilder by rememberUpdatedState(requestBuilder)
```
andkulikov, line 76: the issue you are solving with this should probably be solved on a different level. the real issue is that when rememberLoadPainter() is reexecuted with an updated loader lambda it still continues using the old one. 
so I suggest to wrap loader param in rememberLoadPainter inside a rememberUpdatedState instead

chrisbanes: I went another way, and made the loader a `@Stable interface`. Feels closer to what we actually want.

imageloading-core/src/main/java/com/google/accompanist/imageloading/Image.kt
chrisbanes: Extracted this default value to a `Defaults` object

andkulikov (approved):

PR #345, opened 2021-04-16
Try out Gradle 7.0's Version catalogs

chrisbanes:
See here for [more information](https://docs.gradle.org/current/userguide/declaring_dependencies.html).


appcompat-theme/build.gradle
```diff
@@ -45,7 +43,7 @@ android {
     }
 
     composeOptions {
-        kotlinCompilerExtensionVersion Libs.AndroidX.Compose.version
+        kotlinCompilerExtensionVersion libs.versions.compose.get()
     }
```
msfjarvis, line 47: Is the `composeOptions` block still necessary? AFAIK it was deprecated and is no longer needed but I might be missing something

chrisbanes: It's still necessary to set the current `kotlinCompilerExtensionVersion` to match your dependency, otherwise you'll get whatever the default is from the AGP version you're using (i.e. will nearly always be older than what you're using).

You might be thinking about the [`kotlinCompilerVersion`](https://developer.android.com/reference/tools/gradle-api/4.1/com/android/build/api/dsl/ComposeOptions#kotlincompilerversion)? You no longer need to set that, as it now just uses whatever Kotlin version you're using.

manuelvicnt (approved):

docs/updating.md
```diff
@@ -37,8 +37,8 @@ Now edit the project to depend on the new Compose SNAPSHOT version:
 
 Edit [dependencies.kt](https://github.com/google/accompanist/blob/main/buildSrc/src/main/java/com/google/accompanist/buildsrc/dependencies.kt):
```
opatry, line 38: @chrisbanes you are still referring to `dependencies.kt` which doesn't exist anymore (same after).

chrisbanes: Thanks: ae26aebd6f3cca8805b56ad5dd7e1b5d7241f17a

Issue #291, opened 2021-03-26
[SystemUiController] Support full screen, hide status bar, navigation bar separately


RinOrz:
On some special screens, we may need to hide part of the system ui or even the full screen, so this should be included in the SystemUiController?

Currently I use:
```kotlin
ViewCompat.getWindowInsetsController(LocalView.current)!!.hide(WindowInsetsCompat.Type.systemBars())
```
But IMHO, this api is not elegant at all...

chrisbanes:
PRs welcome!

RinOrz:
Hahaha okay

RinOrz:
Probably to add docs?

PR #348, opened 2021-04-16
Add more convenient system ui control functions and sample.

RinOrz:
Fixed: #291 

- `var isStatusBarVisible`
- `var isNavigationBarVisible`
- `var isSystemBarsVisible`
- `fun setStatusBarDarkIcons`
- `fun setNavigationBarDarkIcons`
- `fun isDarkStatusBarIcons`
- `fun isDarkNavigationBarIcons`
- Add system ui controller sample

But currently there are some problems with the return values of `isStatusBarVisible`, `isNavigationBarVisible` and `isSystemBarsVisible`. I try to get the visibility of the system bars through `InsetsType.isVisible`, but they are not accurate. I use some other means to get them (but unfortunately they also don’t seem to be accurate). So I am considering changing **mutable property** to **function** and only provide the visibility settings of the system bars without returning the current visibility.

https://user-images.githubusercontent.com/58068445/115105672-fc305680-9f92-11eb-89f4-a296f735d070.mp4


chrisbanes:
First of all, thanks for the PR, but...

> `isStatusBarVisible`, `isNavigationBarVisible` & `isSystemBarsVisible` 

These won't be allowed in. That's what the `WindowInsets` library is for, and detecting the whether the bars are visible or not is usually an anti-pattern.

> `setStatusBarDarkIcons`, `setNavigationBarDarkIcons`, `isDarkNavigationBarIcons()` & `isDarkStatusBarIcons()`

What's the use case here? I think most things are covered by `setNavigationBarColor()` & `setStatusBarColor()` but these could be useful.

RinOrz:
> > `setStatusBarDarkIcons`, `setNavigationBarDarkIcons`, `isDarkNavigationBarIcons()` & `isDarkStatusBarIcons()`
> 
> What's the use case here? I think most things are covered by `setNavigationBarColor()` & `setStatusBarColor()` but these could be useful.

Sometimes we don’t always need to change status bar color, but only need to darken the icons. For example, the status bar color is always transparent but the background image will change.

RinOrz:
> First of all, thanks for the PR, but...
> 
> > `isStatusBarVisible`, `isNavigationBarVisible` & `isSystemBarsVisible`
> 
> These won't be allowed in. That's what the `WindowInsets` library is for, and detecting the whether the bars are visible or not is usually an anti-pattern.

So should they be changed to `set**BarVisible(isVisible: Boolean = true)`?

chrisbanes:
> > What's the use case here? I think most things are covered by `setNavigationBarColor()` & `setStatusBarColor()` but these could be useful.
> 
> Sometimes we don’t always need to change status bar color, but only need to darken the icons. For example, the status bar color is always transparent but the background image will change.

True, in that case please remove `isStatusBarVisible`, `isNavigationBarVisible` & `isSystemBarsVisible` and I'll take a look through the PR.

> So should they be changed to set**BarVisible(isVisible: Boolean = true)?

Maybe, I'm a bit worried that we're just creating a wrapper around [`WindowInsetsController`](https://developer.android.com/reference/androidx/core/view/WindowInsetsControllerCompat). Apps shouldn't be afraid to use Android APIs from within Compose, and the goal of `SystemUiController` isn't to be a copy of `WindowInsetsController`

RinOrz:
> Maybe, I'm a bit worried that we're just creating a wrapper around [`WindowInsetsController`](https://developer.android.com/reference/androidx/core/view/WindowInsetsControllerCompat). Apps shouldn't be afraid to use Android APIs from within Compose, and the goal of `SystemUiController` isn't to be a copy of `WindowInsetsController`

But in the sense of the name of the `SystemUiController`, I think they have at least. In addition, personally, I hope to have a more convenient APIs in the Compose world.

PR #350, opened 2021-04-20
[SwipeRefresh] Make indicator more flexible

chrisbanes:
There are a few improvements here:

- Added more parameters to `SwipeRefresh()` to tweak the positioning and padding of the indicator. This allows apps which use `contentPadding` on `LazyRow` etc, to get the indicator to match.
- Move all positioning of the indicator to `SwipeRefreshIndicator()` itself. This allows devs to provide any type of indicator they want. The update custom indiciator sample shows a fade-in gradient instead demo.
- Remove the 'refresh trigger' value from `SwipeRefreshState`. It's never really been part of the state, and was only there for convenience.


swiperefresh/src/main/java/com/google/accompanist/swiperefresh/SwipeRefreshIndicator.kt
```diff
@@ -144,61 +102,108 @@ inline fun SwipeRefreshIndicator(
...
     shape: Shape = MaterialTheme.shapes.small.copy(CornerSize(percent = 50)),
+    refreshingOffset: Dp = 16.dp,
     largeIndication: Boolean = false,
     elevation: Dp = 4.dp,
 ) {
-    val adjustedElevation = when (offset) {
-        0f -> 0.dp
-        else -> elevation
+    val adjustedElevation = when {
+        state.isRefreshing -> elevation
```
ricknout, line 118: Are these two conditions mutually exclusive?

swiperefresh/src/main/java/com/google/accompanist/swiperefresh/SwipeRefresh.kt
```diff
@@ -232,22 +227,27 @@ fun SwipeRefresh(
     onRefresh: () -> Unit,
     modifier: Modifier = Modifier,
     swipeEnabled: Boolean = true,
-    indicator: @Composable (SwipeRefreshState) -> Unit = { SwipeRefreshIndicator(it) },
+    refreshTriggerDistance: Dp = 80.dp,
```
ricknout, line 230: `80.dp` should maybe be a constant somewhere?

ricknout (approved):
A couple minor nits, existing demos look good and like the custom demo!

nickbutcher (approved):

swiperefresh/src/main/java/com/google/accompanist/swiperefresh/SwipeRefreshIndicator.kt
chrisbanes: Nope, but they can be exclusive

swiperefresh/src/main/java/com/google/accompanist/swiperefresh/SwipeRefresh.kt
chrisbanes: I'm not sure, I personally don't feel the need to outline constants for default simple parameter values. That's what control click is for. 😄

PR #355, opened 2021-04-21
[SwipeRefresh] Simplify drag handling for now

chrisbanes:
Previously we allowed dragging of the indicator when it was
'refreshing'. This no longer works perfectly when the recent
migration the offset logic to the SwipeRefreshIndicator.

This PR removes this feature for now, until we can make it work
well with the new indicator.


ricknout (approved):

ricknout:
FWIW I don't think the View-based swipe refresh allows dragging while in a refreshing state so perhaps this aligns better with that

chrisbanes:
> FWIW I don't think the View-based swipe refresh allows dragging while in a refreshing state so perhaps this aligns better with that

Yeah, I was trying to make the Compose version more playful. Something we revisit

PR #357, opened 2021-04-21
Update README.md to include swipe-to-refresh link

mr3y-the-programmer:
link  swipe-to-refresh library in README to be consistent with other libraries

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,6 +22,9 @@ A library which provides paging layouts for Jetpack Compose, similar to Android'
 ### 🌊 [Flow layouts](./flowlayout/)
 A library that adds a 'flexbox'-like layout to [Jetpack Compose][compose].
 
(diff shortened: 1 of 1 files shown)


mr3y-the-programmer:
I've already signed it before but my email address is made private since then, so, I think I need to use github provided `no-reply` email address

chrisbanes:
> I've already signed it before but my email address is made private since then, so, I think I need to use github provided `no-reply` email address

Hi, I don't see a CLA for your username in the system. Did you use a different account?

mr3y-the-programmer:
Thanks @chrisbanes for your response!, I've already solved the problem

mr3y-the-programmer:
@googlebot I signed it!

chrisbanes:
Cool, thanks!

PR #380, opened 2021-04-29
[Image loading] Support placeholder drawables

chrisbanes:
`Loader` now returns a `Flow<ImageLoadState>` which allows the loader to control the entire stream, including the
'Loading' state. This allows the loading state to contain any placeholder drawables which may be set on the request.

Example:

``` kotlin
Image(
    painter = rememberCoilPainter(
        request = "some.url",
        requestBuilder = {
            placeholder(R.drawable.placeholder)
        }
    ),
    contentDescription = null,
)
```

### DIFF
--- a/glide/src/main/java/com/google/accompanist/glide/Glide.kt
+++ b/glide/src/main/java/com/google/accompanist/glide/Glide.kt
@@ -142,10 +142,14 @@ internal class GlideLoader(
             if (size.height > 0) size.height else Target.SIZE_ORIGINAL
         ) {
             override fun onLoadStarted(placeholder: Drawable?) {
+                if (isClosedForSend) return
+
                 sendBlocking(ImageLoadState.Loading(placeholder, request))
             }
 
             override fun onLoadFailed(errorDrawable: Drawable?) {
+                if (isClosedForSend) return
+
                 sendBlocking(
                     ImageLoadState.Error(
                         result = errorDrawable,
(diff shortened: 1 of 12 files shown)


glide/src/main/java/com/google/accompanist/glide/Glide.kt
```diff
@@ -198,9 +205,11 @@ internal class GlideLoader(
...
             .into(target)
 
-        // If we're cancelled, clear the request from Glide
-        cont.invokeOnCancellation {
-            requestManager.clear(target)
+        // When we're cancelled/closed, clear the request from Glide
+        awaitClose {
+            // Unfortunately we need to post the clear. Glide will crash if the close happens
+            // in the same call frame as the Target calls above (happens in testing)
+            handler.post { requestManager.clear(target) }
```
manuelvicnt, line 212: How about using coroutines for this? `private val handler = Handler(Looper.getMainLooper())` is a hardcoded dependency in this class, if you want to keep it like that, you could create a scope and use `Dispatchers.Main`

manuelvicnt (approved):
Added a nit but LGTM!

glide/src/main/java/com/google/accompanist/glide/Glide.kt
chrisbanes: Turns out that I don't actually need to post now. Removed. 🎉

PR #393, opened 2021-05-07
[WIP] Convert Flowlayout to be KMP

chrisbanes:
Just exploring, nothing to see here.

PR #402, opened 2021-05-11
[Image][Insets][Pager] Remove most unnecessary Android specific API usages

chrisbanes:
Examples:

- Moved to using [Napier](https://github.com/AAkira/Napier) from @AAkira instead of `Log.d` directly.
- Removed the usage of `Drawable` in `ImageLoadState` results. We now use `Painter`.
- Migrate Flow layout tests to use `createComposeRule()` (this is cherry-picked from #393)


flowlayout/src/androidTest/AndroidManifest.xml
```diff
@@ -19,10 +19,8 @@
 
     <uses-sdk android:targetSdkVersion="30" />
 
-    <application>
-        <activity
-            android:name="com.google.accompanist.flowlayout.TestActivity"
-            android:theme="@android:style/Theme.Material.NoActionBar" />
+    <application android:theme="@android:style/Theme.Material.NoActionBar">
+        <activity android:name="androidx.activity.ComponentActivity" />
```
JoseAlcerreca, line 23: Why is this necessary? If it's for compose-test I'm surprise this works here (normally it has to be added to debug/ with `debugImplementation("androidx.compose.ui:ui-test-manifest:$compose_version")`)

chrisbanes: It doesn't have to be in `debugImplementation`, but yeah we can use the manifest dependency instead.

PR #410, opened 2021-05-13
Fix previewPlaceholder declaration in coil docs

pawegio:


pawegio:
Initially I fixed the declaration in docs causing compiler error, but I think it makes sense to remove `previewPlaceholder` from simplest usage as it has a default value and it's described below in **Previews** section.

chrisbanes (approved):

PR #418, opened 2021-05-17
[Pager] Fix PagerState restoration

### LINKED ISSUE #417: [Pager] java.lang.IllegalArgumentException: currentPageOffset must be >= 0 and <= 1 after restoring the state
## Describe the bug
I am using the `HorizontalPager` component. Sometimes when the state is restored, the exception ` java.lang.IllegalArgumentException: currentPageOffset must be >= 0 and <= 1 ` occurs . With some debugging I found out, that this is caused by the `Saver` object.

When the state of the pager is restored internally at [this line](https://github.com/google/accompanist/blob/1a181943e697d662785c7f2584ac940cd9ccf707/pager/src/main/java/com/google/accompanist/pager/PagerState.kt#L635), the value for `currentPageOffset` sometimes is `-1.0` or `-2.0`. This causes the test in [this line](https://github.com/google/accompanist/blob/1a181943e697d662785c7f2584ac940cd9ccf707/pager/src/main/java/com/google/accompanist/pager/PagerState.kt#L621) to fail and results in the mentioned crash. The values for `pageCount` and `currentPage` are always correct when saving the state.

## To Reproduce
Unfortunately, I can't deterministically reproduce the crash. It occurs from time to time when I scroll forth and back, save the state by going to another screen and returning. If I can provide a 100% crashing example, I add it below.

## Expected behavior
The restored value should meet the required range of `(0.0f ... 1.0f)`.

## Environment:
 - Android OS version: 11
 - Device: Xiaomi Mi Mix 2S
 - Accompanist version: 0.9.1

chrisbanes:
There's no need to store the page offset as it could technically be out of bounds (example: if the state was stored during an animation).

Fixes #417

### DIFF
--- a/pager/src/main/java/com/google/accompanist/pager/PagerState.kt
+++ b/pager/src/main/java/com/google/accompanist/pager/PagerState.kt
@@ -627,12 +627,11 @@ class PagerState(
          * The default [Saver] implementation for [PagerState].
          */
         val Saver: Saver<PagerState, *> = listSaver(
-            save = { listOf<Any>(it.pageCount, it.currentPage, it.currentPageOffset) },
+            save = { listOf(it.pageCount, it.currentPage) },
             restore = {
                 PagerState(
-                    pageCount = it[0] as Int,
-                    currentPage = it[1] as Int,
-                    currentPageOffset = it[2] as Float
+                    pageCount = it[0],
+                    currentPage = it[1],
                 )
             }
         )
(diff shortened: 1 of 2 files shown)


pager/src/main/java/com/google/accompanist/pager/PagerState.kt
```diff
@@ -627,12 +627,11 @@ class PagerState(
          * The default [Saver] implementation for [PagerState].
          */
         val Saver: Saver<PagerState, *> = listSaver(
-            save = { listOf<Any>(it.pageCount, it.currentPage, it.currentPageOffset) },
+            save = { listOf<Any>(it.pageCount, it.currentPage) },
```
nickbutcher, line 630: Is the `Any` type still needed now that these are all `Int`?

chrisbanes: Fixed.

nickbutcher (approved):

PR #452, opened 2021-05-30
Fix wrong Coil API usage in Glide sample

fornewid:
This is a minor fix.
Currently in `GlideLazyColumnSample`, `rememberCoilPainter()` is used.
So I changed `rememberCoilPainter()` to `rememberGlidePainter()`.


chrisbanes (approved):
Woops, thanks!

PR #455, opened 2021-06-01
[Pager] Consume post flings and scrolls from nested scrolling content

### LINKED ISSUE #347: [Pager] HorizontalPager continues the fling started on LazyRow
## Describe the bug
I have a HorizontalPager which also sometimes contains a LazyRow. When the contents of the LazyRow are longer than the width of the page, the LazyRow becomes scrollable. And if I fling on it, when it reaches the end, HorizontalPager continues the movement and jumps a bit to the scrolled side.

## To Reproduce
:
1. Have a HorizontalPager with LazyRow in it
2. Fling LazyRow

## Expected behavior
Like in a traditional viewpager + recyclerview, if LazyRow intercepts the fling, it should consume it entirely, the pager should not jump.

## Environment:
 - Device: Samsung S21
 - Accompanist version: v0.7.1

chrisbanes:
Currently any nested unused scrolls + flings from child content is handled by the Pager. This is different to how `ViewPager[2]` behaves and is usually unwanted behavior.

This PR adds a `NestedScrollConnection` which consumes those events before the Pager gets them. It's not perfect, but it's close enough for now.

Fixes #347

### DIFF
--- a/pager/src/androidTest/java/com/google/accompanist/pager/HorizontalPagerScrollingContentTest.kt
+++ b/pager/src/androidTest/java/com/google/accompanist/pager/HorizontalPagerScrollingContentTest.kt
@@ -1,0 +1,99 @@
+/*
+ * Copyright 2021 The Android Open Source Project
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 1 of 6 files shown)


matvei-g (approved):
Thanks, LGTM.

Consider adding a test for it as well please

chrisbanes:
> Consider adding a test for it as well please

Test added 👍

PR #457, opened 2021-06-01
[Placeholder] Prototype for new library

### LINKED ISSUE #336: [Idea] Content placeholder/skeleton UI
Something along the lines of https://www.smashingmagazine.com/2020/04/skeleton-screens-react/

I imagine that we can add a `Modifier.placeholder(enabled: Boolean)`, which will draw some skeleton color/shape in-front of the layout content.

fornewid:
I created a prototype for content placeholders.
If it's okay to go ahead, I'll add some test code.
Or, if you advise on the direction, I will correct it.

-------

New APIs:
- We can use `Modifier.placeholder(visible: Boolean)` to show a content placeholder.
- And we can show an animated content placeholder
   using `PlaceholderAnimatedBrush` like `fadeBrush()` or `shimmerBrush()`.

Fixes #336 

## Demo

Basic | Fade | Shimmer
-- | -- | --
 |  |


chrisbanes:
Thanks, I'll take a look at this tomorrow once the next release is out.

placeholder/src/main/java/com/google/accompanist/placeholder/Placeholder.kt
```diff
@@ -0,0 +1,150 @@
...
+ *
+ * @param visible defines whether the placeholder should be visible
+ * @param color color to paint placeholder with
+ * @param shape desired shape of the placeholder
+ */
+fun Modifier.placeholder(
+    visible: Boolean,
+    color: Color = PlaceholderDefaults.PlaceholderColor,
+    shape: Shape = RectangleShape
+) = takeIf { visible.not() } ?: composed(
```
chrisbanes, line 49: Please add the explicit return type

placeholder/src/main/java/com/google/accompanist/placeholder/Placeholder.kt
```diff
@@ -0,0 +1,150 @@
...
+ *
+ * @param visible defines whether the placeholder should be visible
+ * @param animatedBrush animated brush to paint placeholder with
+ * @param shape desired shape of the placeholder
+ */
+fun Modifier.placeholder(
+    visible: Boolean,
+    animatedBrush: PlaceholderAnimatedBrush,
+    shape: Shape = RectangleShape
+) = takeIf { visible.not() } ?: composed(
```
chrisbanes, line 78: Ditto, return type

placeholder/src/main/java/com/google/accompanist/placeholder/Placeholder.kt
```diff
@@ -0,0 +1,150 @@
...
+        properties["shape"] = shape
+    }
+) {
+    val infiniteTransition = rememberInfiniteTransition()
+    val progress by infiniteTransition.animateFloat(
+        initialValue = animatedBrush.initialValue(),
+        targetValue = animatedBrush.targetValue(),
+        animationSpec = animatedBrush.animationSpec()
+    )
+    PlaceholderModifier(
```
chrisbanes, line 93: This will create a new modifier on every animation frame. It's better if we make the brush `@Stable` and remember it. We can  then expose a `var progress by mutableStateOf(0f)` in the brush, and update it here using `brush.progress = progress`.

It results in the same thing but we only create one brush which is much better for performance.

placeholder/src/main/java/com/google/accompanist/placeholder/Placeholder.kt
```diff
@@ -0,0 +1,150 @@
...
+    }
+
+    override fun hashCode(): Int {
+        var result = color?.hashCode() ?: 0
+        result = 31 * result + (brush?.hashCode() ?: 0)
+        result = 31 * result + shape.hashCode()
+        return result
+    }
+
+    override fun equals(other: Any?): Boolean {
```
chrisbanes, line 141: No need for equals/hashcode here.

placeholder/src/main/java/com/google/accompanist/placeholder/PlaceholderAnimatedBrush.kt
```diff
@@ -0,0 +1,178 @@
...
+import androidx.compose.animation.core.RepeatMode
+import androidx.compose.animation.core.infiniteRepeatable
+import androidx.compose.animation.core.tween
+import androidx.compose.ui.graphics.Brush
+import androidx.compose.ui.graphics.Color
+import androidx.compose.ui.graphics.SolidColor
+import androidx.compose.ui.graphics.lerp
+
+internal object PlaceholderDefaults {
+    val PlaceholderColor = Color.Gray.copy(alpha = 0.5f)
```
chrisbanes, line 31: Did you use `Gray` here to avoid a dependency on Material? Ideally this would be set to `contentColor()` but I'm not sure that we want to depend on Material

placeholder/src/main/java/com/google/accompanist/placeholder/PlaceholderAnimatedBrush.kt
```diff
@@ -0,0 +1,178 @@
...
+ * @param animationSpec the [AnimationSpec] to configure the animation.
+ */
+fun shimmerBrush(
+    color: Color = PlaceholderDefaults.PlaceholderColor,
+    highlightColor: Color = PlaceholderDefaults.PlaceholderHighlightColor,
+    animationSpec: InfiniteRepeatableSpec<Float> = infiniteRepeatable(
+        animation = tween(
+            delayMillis = 500,
+            durationMillis = 500,
+            easing = LinearEasing
```
chrisbanes, line 79: This should be `FastOutSlowInEasing`

placeholder/src/main/java/com/google/accompanist/placeholder/PlaceholderAnimatedBrush.kt
```diff
@@ -0,0 +1,178 @@
...
+ * @param animationSpec the [AnimationSpec] to configure the animation.
+ */
+fun fadeBrush(
+    initialColor: Color = PlaceholderDefaults.PlaceholderColor,
+    targetColor: Color = PlaceholderDefaults.PlaceholderHighlightColor,
+    animationSpec: InfiniteRepeatableSpec<Float> = infiniteRepeatable(
+        animation = tween(
+            delayMillis = 0,
+            durationMillis = 500,
+            easing = LinearEasing
```
chrisbanes, line 51: `FastOutSlowInEasing`

placeholder/src/main/java/com/google/accompanist/placeholder/PlaceholderAnimatedBrush.kt
```diff
@@ -0,0 +1,178 @@
...
+    ),
+): PlaceholderAnimatedBrush {
+    return Shimmer(
+        color = color,
+        highlightColor = highlightColor,
+        animationSpec = animationSpec
+    )
+}
+
+sealed class PlaceholderAnimatedBrush {
```
chrisbanes, line 91: Needs kdoc, same for functions below

placeholder/src/main/java/com/google/accompanist/placeholder/PlaceholderAnimatedBrush.kt
```diff
@@ -0,0 +1,178 @@
...
+import androidx.compose.animation.core.LinearEasing
+import androidx.compose.animation.core.RepeatMode
+import androidx.compose.animation.core.infiniteRepeatable
+import androidx.compose.animation.core.tween
+import androidx.compose.ui.graphics.Brush
+import androidx.compose.ui.graphics.Color
+import androidx.compose.ui.graphics.SolidColor
+import androidx.compose.ui.graphics.lerp
+
+internal object PlaceholderDefaults {
```
chrisbanes, line 30: This should be public, and so needs kdoc

placeholder/src/main/java/com/google/accompanist/placeholder/PlaceholderAnimatedBrush.kt
```diff
@@ -0,0 +1,178 @@
...
+/**
+ * Creates a [Fade] brush with the given initial and target colors.
+ *
+ * @sample com.google.accompanist.sample.placeholder.PlaceholderFadeSample
+ *
+ * @param initialColor initial color of the fade.
+ * @param targetColor target color of the fade.
+ * @param animationSpec the [AnimationSpec] to configure the animation.
+ */
+fun fadeBrush(
```
chrisbanes, line 44: Could this be an ext-fun on `PlaceholderAnimatedBrush` instead? `PlaceholderAnimatedBrush.fade(...)`

placeholder/src/main/java/com/google/accompanist/placeholder/PlaceholderAnimatedBrush.kt
```diff
@@ -0,0 +1,178 @@
...
+/**
+ * Creates a [Shimmer] brush with a highlight color over the given color.
+ *
+ * @sample com.google.accompanist.sample.placeholder.PlaceholderShimmerSample
+ *
+ * @param color
+ * @param highlightColor
+ * @param animationSpec the [AnimationSpec] to configure the animation.
+ */
+fun shimmerBrush(
```
chrisbanes, line 72: Ditto, ext-fun

chrisbanes (commented):
Great start! 👏 👏 👏 👏 👏 

Left a few comments but its not far off!

placeholder/src/main/java/com/google/accompanist/placeholder/Placeholder.kt
fornewid: I was thinking about this too.
Thanks for the good suggestion!
I'll fix it with your suggestion.

placeholder/src/main/java/com/google/accompanist/placeholder/PlaceholderAnimatedBrush.kt
fornewid: Yes, I used Gray to avoid dependency on Material.
But if it's better to rely on the material, I'll change.

placeholder/src/main/java/com/google/accompanist/placeholder/PlaceholderAnimatedBrush.kt
fornewid: 🙆‍♂️

fornewid: Sorry, I didn't fully understand your suggestion.
Are you talking about moving inside the `companion object`?

chrisbanes: Sorry yes.

fornewid:
I applied all comments. :bow:

chrisbanes (approved):

moffpage:
Omg 😍

PR #472, opened 2021-06-09
[All] Only run tests on affected modules

chrisbanes:
We now use https://github.com/dropbox/AffectedModuleDetector to only run tests from affected modules for PRs and commits (i.e. to main).

We also now have a twice-daily workflow which runs all tests on emulators, for sanity checking purposes.

Confirmed that this works in #473

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -63,7 +63,6 @@ jobs:
       fail-fast: false
       matrix:
         api-level: [22, 26, 28, 29]
-        shard: [0, 1, 2]
 
     env:
       TERM: dumb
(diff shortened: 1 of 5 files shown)


manuelvicnt:
Is there a way to avoid showing the sharded checks in the check list? There are some pending `Expected — Waiting for status to be reported` ones

chrisbanes:
> Is there a way to avoid showing the sharded checks in the check list? There are some pending `Expected — Waiting for status to be reported` ones

Yeah, I’ll update the branch rules once this PR has gone in

manuelvicnt (approved):

PR #479, opened 2021-06-10
[Placeholder] Add guide and more doc samples

chrisbanes:
Also added more doc samples and attached them to the relevant code
kdoc.


docs/placeholder.md
```diff
@@ -0,0 +1,208 @@
...
+
+[![Maven Central](https://img.shields.io/maven-central/v/com.google.accompanist/accompanist-placeholder)](https://search.maven.org/search?q=g:com.google.accompanist)
+
+A library which provides a [modifier][modifier] for display 'placeholder' UI while content is loading.
+
+More information on the UX provided by this library can be found on the Material Theming [Placeholder UI](https://material.io/design/communication/launch-screen.html#placeholder-ui) guidelines.
+
+There are actually two versions of the library available:
+
+* *Placeholder Foundation*: Provides the base functionality and depends on Jetpack Compose Foundation. This version requires the app to provide all of the colors to display.
```
nickbutcher, line 11: What about re-ordering these as I expect most to use the materail version and only rarely drop down to foundation?

docs/placeholder.md
```diff
@@ -0,0 +1,208 @@
...
+* *Placeholder Material*. This uses the founation library above, but also provides sensible default colors from your app's Material color palette.
+
+You only need to use one or the other, and most apps should use *Placeholder Material*. The API is mostly the same between the libraries, with only the imports being different. Where possible we have provided equivalent code samples below.
+
+## Basic usage
+
+At the most basic usage, the modifier will draw a shape above your composable content, filled with the provided color.
```
nickbutcher, line 18:
```suggestion
At the most basic usage, the modifier will draw a shape over your composable content, filled with the provided color.
```

docs/placeholder.md
```diff
@@ -0,0 +1,208 @@
...
+### Fade
+
+This highlight fades a color over the entire placeholder in and out.
+
+<figure>
+    <video width="400" controls loop>
+    <source src="fade.mp4" type="video/mp4">
+        Your browser does not support the video tag.
+    </video>
+    <figcaption>Placeholder Shimmer demo</figcaption>
```
nickbutcher, line 66:
```suggestion
    <figcaption>Placeholder Fade demo</figcaption>
```

nickbutcher (approved):
LGTM with one small suggesting.

docs/placeholder.md
chrisbanes: Hmmm, it's hard to explain the material version before explaining the foundation version. We explicitly call out below that most apps should use the Material. Still think it's useful?

I could make the para below more prominent instead?

chrisbanes: Went with this:

Screen Shot 2021-06-10 at 12 48 42

nickbutcher: Works for me.

PR #481, opened 2021-06-11
[Placeholder] Add cross-fade between content + placeholder

chrisbanes:
The content and placeholder is now cross-faded when the `visible` value changes. The transition can be customized via the new transition spec parameters.

https://user-images.githubusercontent.com/227486/121664312-41f33200-ca9f-11eb-84d2-62b7dde75f53.mp4

### DIFF
--- a/placeholder/src/main/java/com/google/accompanist/placeholder/Placeholder.kt
+++ b/placeholder/src/main/java/com/google/accompanist/placeholder/Placeholder.kt
@@ -16,12 +16,18 @@
 
 package com.google.accompanist.placeholder
 
+import androidx.compose.animation.core.FiniteAnimationSpec
 import androidx.compose.animation.core.InfiniteRepeatableSpec
+import androidx.compose.animation.core.MutableTransitionState
 import androidx.compose.animation.core.RepeatMode
+import androidx.compose.animation.core.Transition
 import androidx.compose.animation.core.animateFloat
 import androidx.compose.animation.core.infiniteRepeatable
 import androidx.compose.animation.core.rememberInfiniteTransition
+import androidx.compose.animation.core.spring
 import androidx.compose.animation.core.tween
+import androidx.compose.animation.core.updateTransition
+import androidx.compose.runtime.Composable
 import androidx.compose.runtime.getValue
 import androidx.compose.runtime.mutableStateOf
 import androidx.compose.runtime.remember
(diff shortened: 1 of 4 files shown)


placeholder/src/main/java/com/google/accompanist/placeholder/Placeholder.kt
```diff
@@ -99,85 +119,117 @@ fun Modifier.placeholder(
...
+                // If the content alpha is between 1% and 99%, draw it in a layer with
+                // the alpha applied
+                paint.alpha = contentAlpha
+                withLayer(paint) {
+                    with(this@drawWithContent) {
+                        drawContent()
+                    }
+                }
+            } else if (contentAlpha >= 0.99f) {
+                // If the content alpha is between > 99%, draw it with no alpha
```
nickbutcher, line 172:
```suggestion
                // If the content alpha is > 99%, draw it with no alpha
```

placeholder/src/main/java/com/google/accompanist/placeholder/Placeholder.kt
```diff
@@ -99,85 +119,117 @@ fun Modifier.placeholder(
...
 
+            if (placeholderAlpha in 0.01f..0.99f) {
+                // If the placeholder alpha is between 1% and 99%, draw it in a layer with
+                // the alpha applied
+                paint.alpha = placeholderAlpha
+                withLayer(paint) {
+                    drawPlaceholder()
+                }
+            } else if (placeholderAlpha >= 0.99f) {
+                // If the placeholder alpha is between > 99%, draw it with no alpha
```
nickbutcher, line 197:
```suggestion
                // If the placeholder alpha is > 99%, draw it with no alpha
```

nickbutcher (approved):

placeholder/src/main/java/com/google/accompanist/placeholder/Placeholder.kt
```diff
@@ -99,85 +119,117 @@ fun Modifier.placeholder(
...
-            if (shape === RectangleShape) {
-                // shortcut to avoid Outline calculation and allocation
-                // Draw the initial background color
-                drawRectPlaceholder(
-                    color = color,
-                    highlight = highlight,
-                    progress = progress
-                )
-            } else {
+            fun drawPlaceholder() {
```
nadewad, line 176: I would consider defining this function outside of the draw block as this will allocate on each frame.

placeholder/src/main/java/com/google/accompanist/placeholder/Placeholder.kt
```diff
@@ -196,3 +248,12 @@ private fun ContentDrawScope.drawOutlinePlaceholder(
     // Return the outline we used
     return outline
 }
+
+private inline fun DrawScope.withLayer(
```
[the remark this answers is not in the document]
chrisbanes: I'm not sure that I can use `graphicsLayer` as I'm using `drawWithContent` to apply an alpha effect on both the content and the placeholder independently for the crossfade.

Is there any way that I can do #2 without `graphicsLayer`?

placeholder/src/main/java/com/google/accompanist/placeholder/Placeholder.kt
chrisbanes: Thanks, fixed in 350f4b99fe9c04a074c9e39e25cf0eb7786ef2d7

placeholder/src/main/java/com/google/accompanist/placeholder/Placeholder.kt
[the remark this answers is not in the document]
nadewad: I think you should be able to have the placeholder modifier chain a graphicsLayer and a drawWithContent modifiers together with the alpha configured only on the graphicsLayer.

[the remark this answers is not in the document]
chrisbanes: Thanks, I'll have a look

PR #482, opened 2021-06-11
Experiment to improve PermissionState

chrisbanes:
I only updated PermissionState.
MultiplePermissionsState could probably be improved too.

### DIFF
--- a/permissions/src/main/java/com/google/accompanist/permissions/MutablePermissionState.kt
+++ b/permissions/src/main/java/com/google/accompanist/permissions/MutablePermissionState.kt
@@ -16,21 +16,18 @@
 
 package com.google.accompanist.permissions
 
-import android.content.pm.PackageManager
+import androidx.activity.result.ActivityResultLauncher
 import androidx.compose.runtime.Composable
 import androidx.compose.runtime.DisposableEffect
-import androidx.compose.runtime.MutableState
+import androidx.compose.runtime.LaunchedEffect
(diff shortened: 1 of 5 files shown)


permissions/src/main/java/com/google/accompanist/permissions/MutablePermissionState.kt
```diff
@@ -47,112 +44,66 @@ import kotlinx.coroutines.channels.Channel
...
+    override val permission: String,
+) : PermissionState {
+    override var hasPermission: Boolean by mutableStateOf(false)
+    override var shouldShowRationale: Boolean by mutableStateOf(false)
+    override var permissionRequested: Boolean by mutableStateOf(false)
 
-    fun shouldShow(): Boolean =
-        ActivityCompat.shouldShowRequestPermissionRationale(currentActivity, permission)
+    // This feels a bit memory leaky
+    internal lateinit var launcher: ActivityResultLauncher<String>
```
manuelvicnt, line 98: It looks indeed a bit memory leaky. I think we can solve this by not having the `rememberMutablePermissionState` and have that logic inside `rememberPermissionState`. I don't know how that'd influence the multiple requests flow. 

I'll experiment with this but looks good so far! Thank you

chrisbanes: Cool, I'll let you merge this into your branch. I'll close this PR.

PR #501, opened 2021-06-16
[Permissions] Add the PermissionRequired composables

manuelvicnt:
Adds the `PermissionRequired` and `PermissionsRequired` composables that exercise the permissions workflow as described in the [documentation](https://developer.android.com/training/permissions/requesting#workflow_for_requesting_permissions).


permissions/src/main/java/com/google/accompanist/permissions/PermissionsRequired.kt
```diff
@@ -0,0 +1,120 @@
...
+ * @param permissionsRationaleContent content to show when the user needs to be presented with
+ * a rationale for the required permissions.
+ * @param permissionsDeniedContent content to show when the user denied the permissions.
+ * @param permissionsRequestedContent content to show while the permissions are being requested.
+ */
+@ExperimentalPermissionsApi
+@Composable
+fun PermissionsRequired(
+    multiplePermissionsState: MultiplePermissionsState,
+    permissionsGrantedContent: @Composable (() -> Unit),
```
marcelpinto, line 89: Do you think it's worth to provide the specific permission that was granted/revoked/require_rational or since the caller should have access to the state can check manually?

marcelpinto (commented):
LGTM. It follows the correct flow and makes the API simpler

chrisbanes (approved):

permissions/src/main/java/com/google/accompanist/permissions/PermissionsRequired.kt
manuelvicnt: Yeah, that's the idea. As the caller must hold the state instance, they can be as granular/generic as they want. The `MultiplePermissionsState` has a list of `revokedPermissions` that can be used for this

permissions/src/main/java/com/google/accompanist/permissions/PermissionsRequired.kt
```diff
@@ -0,0 +1,101 @@
...
+ * again for this permission. Attempting to request the permission to the user in this part of
+ * Composition has no effect.
+ * @param content content to show when the permission is granted.
+ */
+@ExperimentalPermissionsApi
+@Composable
+fun PermissionRequired(
+    permissionState: PermissionState,
+    noPermissionContent: @Composable (() -> Unit),
+    doNotAskAgainPermissionContent: @Composable (() -> Unit),
```
marcelpinto, line 44: What about calling this permissionNotAvailableContent?

Theoretically it would be more correct, since the situations where you end up in this state it might not only be because of "do not ask again". For example, in work profile some policies might block certain permissions. At the end this state means that you cannot request this permission anymore and any request will end up in permission revoked directly.

Proposal:
permissionNotGrantedContent
permissionNotAvailableContent
content

permissions/src/main/java/com/google/accompanist/permissions/PermissionsRequired.kt
```diff
@@ -0,0 +1,101 @@
...
+    content: @Composable (() -> Unit),
+) {
+    when {
+        permissionState.hasPermission -> {
+            content()
+        }
+        permissionState.shouldShowRationale || !permissionState.permissionRequested -> {
```
marcelpinto, line 51: Consumer can check the permissionState but I am wondering if it makes sense to split or pass the boolean to distinguish these two cases.

I don't have a strong opinion tbh.

permissions/src/main/java/com/google/accompanist/permissions/PermissionsRequired.kt
manuelvicnt: Updated! Thanks

permissions/src/main/java/com/google/accompanist/permissions/PermissionsRequired.kt
manuelvicnt: I think that'd complicate the API. As you said, they could check the state inside the lambda, so this'd be somewhat redundant.

marcelpinto (approved):

chrisbanes (approved):

PR #511, opened 2021-06-21
[Pager] Fix floorMod function in PagerState

halilozercan:
New implementation employs the actual floorMod algorithm from JDK 1.8 by using floorDiv. This implementation fixes Integer overflow possibility.


chrisbanes (approved):
Thanks!

PR #522, opened 2021-06-25
Expose the Drawable in DrawablePainter

r0adkll:
I have a use case where I need to both render the drawable loaded by Glide and extract it's bitmap to generate a glowing effect around it. With the change in https://github.com/google/accompanist/pull/402 that replaced the `drawable` result param for a `Painter` broke our implementation and I couldn't find an alternative solution to achieve the same effect with the `Painter` and have thus opened this PR.

Thanks for the consideration!


chrisbanes (commented):
LGTM, but you need to run `./gradlew metalavaGenerateSignature`

chrisbanes (approved):

Issue #516, opened 2021-06-24
[Idea] [SwipeRefresh] Add swipe direction option


ArtyomZhukov:
It would be great to have an option that allows to chose swipe direction (top, bottom, both) on the SwipeRefresh Composable function like it made in this library  - https://github.com/omadahealth/SwipyRefreshLayout

chrisbanes:
PRs welcome!

sitatec:
I'm working on it

PR #541, opened 2021-07-07
[SwipeRefresh] Add swipe up to refresh feature

sitatec:
Make the swipe to refresh bidirectional, as proposed in #516
- Add a bottom refresh indicator that rotates in the counterclockwise sense (as the top indicator rotates in the clockwise sense) when the user swipes up at the bottom of the list.
- Make both the top and bottom indicators optional, so the user will use either the top, the bottom, or both (but he must choose one).
- Add a Sample for the bottom indicator and both top and bottom indicator used at the same time.
- Update the tests
- Update doc


sitatec:
Review needed

docs/swiperefresh.md
```diff
@@ -34,7 +47,34 @@ SwipeRefresh(
...
+
+Alternatively, you can use a bottom refresh indicator or even both a top and bottom indicator (e.g: If you are loading the list items by page, each time the user triggers the bottom refresh indicator, you load the next page. If the user triggers the top refresh indicator, you refresh the list):
+
+``` kotlin
+val viewModel = viewModel<SampleViewModel>()
+val isTopIndicatorRefreshing by viewModel.isRefreshing.collectAsState()
+val isBottomIndicatorRefreshing by viewModel.isLoadingNextPage.collectAsState()
+    
+SwipeRefresh(
+    topRefreshIndicatorState = rememberSwipeRefreshState(isTopIndicatorRefreshing),
```
chrisbanes, line 60: A layout should only have 1 state instance which should cover the entire state of the layout. The API for this should look something like imo:

``` kotlin
enum class Position { Top, Bottom }

SwipeRefresh(
    state: SwipeRefreshState,
    position: Position,
)
```

If someone wants to have both top & bottom refreshes (which isn't a great UX anyway), they can nest two `SwipeRefresh()`s.

chrisbanes (changes requested):

docs/swiperefresh.md
sitatec: No problem, that will also simplify the implementation and remove all null checks for the indicator state.

sitatec: Isn't it redundant to have both `position` and `indicatorAlignment` ? In addition, a user can have something like this:
```kotlin
SwipeRefresh(
    ...
    position = Position.Bottom,
    indicatorAlignment = Alignment.TopCenter, // Which is the default.
)
```
Which will cause an unexpected behavior.
We should remove the`indicatorAlignment` or replace its type by `Alignment.Horizontal`.

chrisbanes: Yeah you're right. We could create a `SwipeRefreshIndicator.Alignment` which == `Alignment` minus any vertical centering. We could then update `indicatorAlignment` to use that alignment.

WDYT?

sitatec: Yeah, it's a good idea. But it will be simpler if we use the built-in [Alignment.Horizontal](https://developer.android.com/reference/kotlin/androidx/compose/ui/Alignment.Horizontal) interface as the `indicatorAlignment`'s type. 
The `Alignment.Horizontal` allow only horizontal alignment, any vertical alignment will cause a compile time error:

Screen Shot 2021-07-13 at 8 37 44 AM

swiperefresh/src/main/java/com/google/accompanist/swiperefresh/SwipeRefresh.kt
```diff
@@ -223,12 +287,21 @@ fun SwipeRefresh(
     state: SwipeRefreshState,
     onRefresh: () -> Unit,
     modifier: Modifier = Modifier,
+    position: Position = Position.TOP,
     swipeEnabled: Boolean = true,
     refreshTriggerDistance: Dp = 80.dp,
-    indicatorAlignment: Alignment = Alignment.TopCenter,
+    indicatorHorizontalAlignment: Alignment.Horizontal = Alignment.CenterHorizontally,
```
chrisbanes, line 293: This is an API change we don't really want to do. Please revert the name change.

Thinking about it, can we just keep the old API signature and remove the new `position`? We would need to `check()` that the value doesn't contain a center vertical value, but that seems like the simplest option?

swiperefresh/src/main/java/com/google/accompanist/swiperefresh/SwipeRefresh.kt
```diff
@@ -114,10 +116,18 @@ class SwipeRefreshState(
...
-    private val onRefresh: () -> Unit,
+/**
+ * The position for which the "swipe to refresh" must be enabled:
+ */
+enum class Position {
+    TOP,
+    BOTTOM,
+}
+
+private open class SwipeRefreshNestedScrollConnection(
```
chrisbanes, line 127: Can we remove the sub-class and just pass in a `scrollFromTop: Boolean = true`? Sub-classes are complicating the code here and there's no real need AFAICT.

swiperefresh/src/main/java/com/google/accompanist/swiperefresh/SwipeRefresh.kt
```diff
@@ -46,6 +46,8 @@ import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.launch
 import kotlin.math.absoluteValue
 
+typealias SwipeRefreshIndicator = @Composable (state: SwipeRefreshState, refreshTrigger: Dp) -> Unit
```
chrisbanes, line 49: Any reason for this? This will become public API which we should minimize. If you feel strongly about having this as public API, it needs kdoc.

swiperefresh/src/main/java/com/google/accompanist/swiperefresh/SwipeRefresh.kt
```diff
@@ -223,12 +287,21 @@ fun SwipeRefresh(
...
     refreshTriggerDistance: Dp = 80.dp,
-    indicatorAlignment: Alignment = Alignment.TopCenter,
+    indicatorHorizontalAlignment: Alignment.Horizontal = Alignment.CenterHorizontally,
     indicatorPadding: PaddingValues = PaddingValues(0.dp),
     indicator: @Composable (state: SwipeRefreshState, refreshTrigger: Dp) -> Unit = { s, trigger ->
-        SwipeRefreshIndicator(s, trigger)
+        SwipeRefreshIndicator(
+            state = s,
+            refreshTriggerDistance = trigger,
+            senseOfRotation = if (position == Position.TOP) {
```
chrisbanes, line 299: I think this could just be `clockwise: Boolean`. The enum isn't adding anything here, but does bloat the API.

swiperefresh/src/main/java/com/google/accompanist/swiperefresh/CircularProgressPainter.kt
```diff
@@ -34,12 +34,17 @@ import androidx.compose.ui.graphics.painter.Painter
...
+    COUNTERCLOCKWISE,
+}
+
 /**
  * A private class to do all the drawing of SwipeRefreshIndicator, which includes progress spinner
  * and the arrow. This class is to separate drawing from animation.
  * Adapted from CircularProgressDrawable.
  */
-internal class CircularProgressPainter : Painter() {
+internal class CircularProgressPainter(private val senseOfRotation: SenseOfRotation = SenseOfRotation.CLOCKWISE) : Painter() {
```
chrisbanes, line 47: Ditto comment below, could be a `Boolean`

chrisbanes (changes requested):
Nearly there! Just some comments about the API.

As this is changing the API, we need to update the API files. Please run: `./gradlew metalavaGenerateSignature`

swiperefresh/src/main/java/com/google/accompanist/swiperefresh/SwipeRefresh.kt
sitatec: If we keep the old API: 
```kotlin
@Composable
fun SwipeRefresh(
    state: SwipeRefreshState,
    onRefresh: () -> Unit,
    swipeEnabled: Boolean = true,
    refreshTriggerDistance: Dp = 80.dp,
    indicatorAlignment: Alignment = Alignment.TopCenter,
    indicatorPadding: PaddingValues = PaddingValues(0.dp),
    indicator: @Composable (state: SwipeRefreshState, refreshTrigger: Dp) -> Unit = { s, trigger ->
        SwipeRefreshIndicator(s, trigger)
    },
    clipIndicatorToPadding: Boolean = true,
    content: @Composable () -> Unit,
)
```
 How could we know if it's a bottom or top swipe refresh?

swiperefresh/src/main/java/com/google/accompanist/swiperefresh/SwipeRefresh.kt
sitatec: No, I just forgot removing it, it was there for the previous implementation where we had 2 Indicator params.

swiperefresh/src/main/java/com/google/accompanist/swiperefresh/SwipeRefresh.kt
chrisbanes: > How could we know if it's a bottom or top swipe refresh?

Do you need to know? There's only one swipe direction for each `SwipeRefresh()` so you automatically know which it is.

sitatec: When you said old API, did you mean the API before I made this PR:
```kotlin
@Composable
fun SwipeRefresh(
    state: SwipeRefreshState,
    onRefresh: () -> Unit,
    swipeEnabled: Boolean = true,
    refreshTriggerDistance: Dp = 80.dp,
    indicatorAlignment: Alignment = Alignment.TopCenter,
    indicatorPadding: PaddingValues = PaddingValues(0.dp),
    indicator: @Composable (state: SwipeRefreshState, refreshTrigger: Dp) -> Unit = { s, trigger ->
        SwipeRefreshIndicator(s, trigger)
    },
    clipIndicatorToPadding: Boolean = true,
    content: @Composable () -> Unit,
)
```
or you just want us to keep the old signature for only the `indicatorAlignment ` param?

Please can you respond with the code of signature you were talking about?

JoseAlcerreca:
Some unresolved comments and conflicts.

sitatec:
> Some unresolved comments and conflicts.

@JoseAlcerreca, I resolved them.

PhilipDukhov:
`SwipeRefresh` [the change in line 301](https://github.com/sitatec/accompanist/blob/d0133e25decd10f7a6dd204a9e9bbd1716fe8dd3/swiperefresh/src/main/java/com/google/accompanist/swiperefresh/SwipeRefresh.kt#L301) looks like an unused expression.

phamsarah:
Hi, is this going to be implemented? I would like to swipe up from the bottom.

sitatec:
@phamsarah lt actually have been implemented since a year now, just not merged yet. If the maintainers will review and merge it, I will fix the conflict.

Issue #545, opened 2021-07-08
[Pager] The initialOffscreenLimit value becomes default if you switch between tabs using the bottom navigation


DoTheMonkeyBusiness:
## Describe the bug
I use a pager with several items visible on the screen at the same time, so the value in the `initialOffscreenLimit` is large (e.g. 7). If I open the screen for the first time everything is OK, All elements are visible, but if I switch between tabs and go back I only see 2 items

## To Reproduce
1. Go to the screen with a pager that has few simultaneously visible items on the screen
2. Switch navigation tab.
3. Go back.
4. See error.

## Expected behavior
After switching back, all expected elements should be visible.

## Screenshots
Example:

https://user-images.githubusercontent.com/36794185/124895173-fe7bdd00-dfe4-11eb-9f0d-736db8bc9fbb.mov

## Environment:
 - Android OS version: Android 10.0
 - Device: Emulator, Google Pixel 4
 - Accompanist version: v0.13.0

wthee:
@chrisbanes Maybe initialOffscreenLimit also needs to be added to   `PagerState.Saver`

PR #547, opened 2021-07-09
[Pager] add offscreenLimit to PagerState Saver

wthee:
It looks like all properties need to be saved.

Fix: #545

### DIFF
--- a/pager/src/main/java/com/google/accompanist/pager/PagerState.kt
+++ b/pager/src/main/java/com/google/accompanist/pager/PagerState.kt
@@ -666,6 +666,8 @@ class PagerState(
                 listOf<Any>(
                     it.pageCount,
                     it.currentPage,
+                    it.currentPageOffset,
+                    it.offscreenLimit,
                     it.infiniteLoop,
                 )
             },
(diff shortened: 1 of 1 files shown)


pager/src/main/java/com/google/accompanist/pager/PagerState.kt
```diff
@@ -666,14 +666,18 @@ class PagerState(
                 listOf<Any>(
                     it.pageCount,
                     it.currentPage,
+                    it.currentPageOffset,
```
chrisbanes, line 669: We explicitly do not want the offset to be saved.

chrisbanes (changes requested):

pager/src/main/java/com/google/accompanist/pager/PagerState.kt
wthee: Does `offscreenLimit `need to be saved?

chrisbanes: I don't think so. The value is required to be set at init, and the value rarely changes.

wthee: When navigating to a new page and then returning, the value of `PagerState.offscreenLimit` will always set to 1.  Just like the video of #545

chrisbanes: Ah ok, let's keep that then!

chrisbanes (approved):

PR #550, opened 2021-07-11
[Coil][Glide] Deprecate the libraries

chrisbanes:
Following on from the release of [`coil-compose`](https://coil-kt.github.io/coil/compose/) in Coil 1.3.0, this PR deprecates Accompanist Coil. 

This left Accompanist Glide is a weird situation, where most of `imageloading-core` was there to only support Glide. Since Coil is better suited for use in Compose I've decided to deprecate everything: `accompanist-coil`, `accompanist-glide` and `accompanist-imageloading-core`. All of the libraries will be removed soon (likely v0.16.0)

I need to migrate `DrawablePainter` over to a new library, but that can wait for now.

### DIFF
--- a/coil/api/current.api
+++ b/coil/api/current.api
@@ -2,13 +2,13 @@
 package com.google.accompanist.coil {
 
   public final class CoilKt {
-    method public static androidx.compose.runtime.ProvidableCompositionLocal<coil.ImageLoader> getLocalImageLoader();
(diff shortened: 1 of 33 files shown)


JoseAlcerreca (approved):

sanjeevirajm:
Check this. https://github.com/bumptech/glide/issues/4459
Glide doesn't support compose. Also, the major contributor of Glide said he is not familiar with Compose or Kotlin.
Someone else need to add compose support for Glide.
Instead of deprecating it, can you make this as a recommended way until someone adds support in Glide out of the box

chrisbanes:
No, we won't be un-deprecating the Glide library. I heavily recommend moving to Coil.

sanjeevirajm:
If someone is looking for compose support for Glide, try this one
https://github.com/sanjeevirajm/GlideCompose/

PR #560, opened 2021-07-14
documentation link correcting to S

surajsahani:
adding  for better documentation


chrisbanes (changes requested):
This doesn't look correct 🤔

surajsahani:
> This doesn't look correct

Okay. I was trying to link up the entire word in the documentation

PR #562, opened 2021-07-14
adding highlighting markdown

surajsahani:


chrisbanes (commented):
The original text is correct, they're talking about the general 'thing' rather than the API

PR #565, opened 2021-07-16
Update to Compose snapshot 7555628.

nickbutcher:


chrisbanes:
Force merging this as I've just pushed a workaround for the emulator issues.

PR #574, opened 2021-07-23
[Navigation] Add navigation module

jbw0033:
New module for allowing custom transition in Navigation Compose.

New API: AnimatedNavHost


mkdocs.yml
```diff
@@ -50,6 +50,9 @@ nav:
   - 'Permissions':
     - 'Guide': permissions.md
     - 'API': api/permissions/
+  - 'Navigation':
```
jossiwolf, line 53: Seeing that we'll have navigation-compose-material as well, how do you feel about changing this module to be navigation-animation so that we can keep the dependencies separated? Or are we good with having it all in one module? :)

navigation/build.gradle
```diff
@@ -0,0 +1,106 @@
...
+        // We run a full lint analysis as build part in CI, so skip vital checks for assemble tasks
+        checkReleaseBuilds false
+    }
+
+    testOptions {
+        unitTests {
+            includeAndroidResources = true
+        }
+        animationsDisabled true
+        execution 'ANDROIDX_TEST_ORCHESTRATOR'
```
chrisbanes, line 69: You probably don't need this line

navigation/build.gradle
```diff
@@ -0,0 +1,106 @@
...
+    api libs.androidx.navigation.compose
+    implementation libs.compose.animation.animation
+    implementation libs.compose.foundation.foundation
+    implementation libs.compose.material.material
+
+    // ======================
+    // Test dependencies
+    // ======================
+
+    androidTestUtil libs.androidx.test.orchestrator
```
chrisbanes, line 89: Ditto, probably don't need this

navigation/build.gradle
```diff
@@ -0,0 +1,106 @@
...
+
+    androidTestImplementation libs.junit
+    androidTestImplementation libs.truth
+
+    androidTestImplementation libs.compose.ui.test.junit4
+    androidTestImplementation libs.compose.ui.test.manifest
+    androidTestImplementation libs.compose.foundation.foundation
+    androidTestImplementation libs.androidx.test.runner
+    androidTestImplementation libs.androidx.test.rules
+    androidTestImplementation libs.androidx.test.uiAutomator
```
chrisbanes, line 103: ...or this

mkdocs.yml
chrisbanes: +1. I think this module should be `navigation-animation`.

> Seeing that we'll have navigation-compose-material as well...

Will we?

chrisbanes (commented):
Looks like you need to run `./gradlew spotlessApply`

navigation/build.gradle
jbw0033: removed

mkdocs.yml
jbw0033: changed to `navigation-animation`

ianhanniballake: The intent is for `compose-material` dependent APIs to be separate from the core `navigation-compose` library (as not every Compose app will be a Compose Material app), so that would indeed imply a future `navigation-compose-material` artifact. We're not there yet though :smile:

gradle/libs.versions.toml
```diff
@@ -1,6 +1,6 @@
 [versions]
-compose = "1.0.1-SNAPSHOT"
-composesnapshot = "7584688" # a single character = no snapshot
+compose = "1.1.0-SNAPSHOT"
```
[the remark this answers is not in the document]
jbw0033: I know it works with `1.0.1`, but not with `1.0.0`. Only used `1.1.0` because we currently need the right nav snapshot and all versions that had what we needed came with `1.1.0-SNAPSHOT`.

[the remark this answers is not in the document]
chrisbanes: Ah, makes sense!

sample/src/main/java/com/google/accompanist/sample/navigation/animation/AnimatedNavHostSample.kt
```diff
@@ -0,0 +1,295 @@
...
+            }
+        }
+    }
+}
+
+@ExperimentalAnimationApi
+@Composable
+fun ExperimentalAnimationNav() {
+    val navController = rememberNavController()
+    navController.navigatorProvider += AnimatedComposeNavigator()
```
chrisbanes, line 73: This is going to happen on every composition. Shouldn't this be in an effect?

``` kotlin
DisposableEffect(navController) {
    navController.navigatorProvider += AnimatedComposeNavigator()
    onDispose { ... }
}
```

navigation-animation/src/main/java/com/google/accompanist/navigation/animation/AnimatedNavHost.kt
```diff
@@ -0,0 +1,270 @@
...
+ */
+@Composable
+@ExperimentalAnimationApi
+public fun AnimatedNavHost(
+    navController: NavHostController,
+    startDestination: String,
+    modifier: Modifier = Modifier,
+    route: String? = null,
+    enterTransition: (initial: NavBackStackEntry, target: NavBackStackEntry) -> EnterTransition =
+        { _, _ -> fadeIn(animationSpec = tween(2000)) },
```
chrisbanes, line 78: 2000 feels like a long default! Shouldn't this be ~700ms according to material?

sample/src/main/java/com/google/accompanist/sample/navigation/animation/AnimatedNavHostSample.kt
chrisbanes: Also, could we make `AnimatedNavHost()` do this automatically? Feels like something which devs shouldn't need to think about. We could then make `AnimatedComposeNavigator` internal?

navigation-animation/build.gradle
```diff
@@ -0,0 +1,105 @@
...
+        minSdkVersion 21
+        // targetSdkVersion has no effect for libraries. This is only used for the test APK
+        targetSdkVersion 30
+
+        testInstrumentationRunner "androidx.test.runner.AndroidJUnitRunner"
+
+        // The following argument makes the Android Test Orchestrator run its
+        // "pm clear" command after each test invocation. This command ensures
+        // that the app's state is completely cleared between tests.
+        testInstrumentationRunnerArguments clearPackageData: 'true'
```
chrisbanes, line 40: Can remove this line + comment

navigation-animation/src/main/java/com/google/accompanist/navigation/animation/AnimatedNavHost.kt
```diff
@@ -0,0 +1,270 @@
...
+
+    // on successful recompose we setup the navController with proper inputs
+    // after the first time, this will only happen again if one of the inputs changes
+    navController.setLifecycleOwner(lifecycleOwner)
+    navController.setViewModelStore(viewModelStoreOwner.viewModelStore)
+    if (onBackPressedDispatcher != null) {
+        navController.setOnBackPressedDispatcher(onBackPressedDispatcher)
+    }
+
+    navController.graph = graph
```
chrisbanes, line 150: This should be done in an effect? It's updating global state in a composition which might not be 'committed'

sample/src/main/java/com/google/accompanist/sample/navigation/animation/AnimatedNavHostSample.kt
ianhanniballake: No, this needs to happen immediately (and is idempotent and safe to do every composition). The request for making an official API for this pattern is being tracked in [b/193555369](https://issuetracker.google.com/193555369).

We could certainly make a `rememberAnimatedNavController()` API to encapsulate this work, but this work does need to happen outside of the NavHost.

navigation-animation/src/main/java/com/google/accompanist/navigation/animation/AnimatedNavHost.kt
ianhanniballake: This is also expected at this point.

navigation-animation/build.gradle
jbw0033: Done.

navigation-animation/src/main/java/com/google/accompanist/navigation/animation/AnimatedNavHost.kt
jbw0033: yeah, I was just using that for debugging animations early on. Changed all of the defaults to be 700ms and also made the samples 700ms as well.

sample/src/main/java/com/google/accompanist/sample/navigation/animation/AnimatedNavHostSample.kt
jbw0033: Made the `rememberAnimatedNavController()` API and used that instead.

docs/navigation-animation.md
```diff
@@ -0,0 +1,121 @@
...
+### `AnimatedNavHost`
+
+The `AnimatedNavHost` composable offers a way to add custom transitions to composables in
+Navigation Compose.
+
+```kotlin
+@Composable
+private fun ExperimentalAnimationNav() {
+    val navController = rememberNavController()
+    navController.navigatorProvider += AnimatedComposeNavigator()
```
chrisbanes, line 22: Need to update this sample?

chrisbanes (approved):
Looks good to me. Just need to update the sample on the guide.

PR #592, opened 2021-08-01
[Navigation] Add Navigation-Material

jossiwolf:
New module that allows registering composable bottom sheet destinations with navigation-compose!

This is imported from https://github.com/androidx/androidx/pull/173.

For easier integration with `compose-material`'s `ModalBottomSheetLayout`, this library provides convenience wrappers and a `BottomSheetNavigator` to be able to declare and navigate to bottom sheet destinations, similar to bottom sheet dialog destinations in vanilla navigation.
`BottomSheetNavigator` drives a `ModalBottomSheetLayout`'s `sheetContent` and state, controlling visibility according to the back stack and composing and disposing destinations that are navigated to or away from. `BottomSheetNavigator` offers the `sheetContent` as a composable that should be set as the `ModalBottomSheetLayout`'s `sheetContent`.

- Composable bottom sheet destinations can be added using `NavGraphBuilder#bottomSheet`
- The `BottomSheetNavigator` needs to be registered with the NavController's navigatorProvider
- The main entry point is the `ModalBottomSheetLayout` composable that accepts a `BottomSheetNavigator`


navigation-material/src/main/java/com/google/accompanist/navigation/material/SheetContentHost.kt
```diff
@@ -0,0 +1,147 @@
...
+        // Whenever the composable associated with the latestEntry enters the composition, we
+        // want to show the sheet, and hide it when this composable leaves the composition
+        DisposableEffect(backStackEntry) {
+            scope.launch {
+                sheetState.internalShow()
+                currentOnSheetShown(backStackEntry)
+            }
+            onDispose {
+                scope.launch {
+                    hideCalled = true
```
jossiwolf, line 100: So... I hate this. This stems from us having two sources of truth: A sheet can be dismissed by the user (that's what we look at in the above snapshotFlow in the LaunchedEffect) and when the associated `backStackEntry` leaves the composition in which case we programmatically hide. When we programmatically hide the sheet, we don't want that visibility flow to emit because it means that we already popped the entry off the back stack. Any ideas would be appreciated, I'm hoping there's a simple-ish solution that I am missing here!

chrisbanes:
Looks like you need to run `./gradlew spotlessApply`

navigation-material/src/main/java/com/google/accompanist/navigation/material/BottomSheet.kt
```diff
@@ -0,0 +1,58 @@
...
+import androidx.compose.ui.graphics.Color
+import androidx.compose.ui.graphics.Shape
+import androidx.compose.ui.unit.Dp
+
+/**
+ * Helper function to create a [ModalBottomSheetLayout] from a [BottomSheetNavigator].
+ *
+ * @see [ModalBottomSheetLayout]
+ */
+@ExperimentalMaterialApi
```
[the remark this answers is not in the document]
jossiwolf: Done in 6a7eff6. Should we do the same for navigation-animation? I think we currently propagate `@ExperimentalAnimationApi` usages there :)

navigation-material/src/main/java/com/google/accompanist/navigation/material/SheetContentHost.kt
jossiwolf: Cool, thanks!

chrisbanes (commented):
LGTM. @ianhanniballake to review.

ianhanniballake:
I've done my thorough review as part of androidx/androidx#173 so LGTM

chrisbanes (approved):

chrisbanes (commented):
Looks like you need to run `./gradlew spotlessApply` again. 

Note: Might be worth running `./gradlew check` before pushing 😄

jossiwolf:
Already done :D Yes I'll set up a commit hook, thanks!

jossiwolf:
Looks like the tests that are failing are failing on `snapshot` too fwiw

PR #661, opened 2021-08-18
Add instructions for migration to nav animation docs

jbw0033:
Add instructions for how users can migrate from the jetpack Navigation
Compose APIs to Navigation Accompanist to make it clear to developers.

Also added a link to the Navigation Accompanist blog post.


chrisbanes:
Merged as part of #664

PR #662, opened 2021-08-18
[Navigation Animation] Add AnimatedContentScope and AnimatedVisiblityScope

### LINKED ISSUE #656: [Idea] Support transitionScope similar to AnimatedVisibility API
**Is your feature request related to a problem? Please describe.**
Trying to use expandIn/shrinkOut to do a scale effect, but they don't quite do the desired effect. With AnimatedVisibility, I was able to do it by using the transitionScope via transition which made it even more powerful to customize the transition, but seems the navigation API works differently.

**Describe the solution you'd like**
A similar entry point to AnimatedVisibility, e.g. transitionScope

**Additional context**
Request from internal conversation

ianhanniballake:
The API for a `composable` destination has two important changes:

**1. Animation lambdas now use an [`AnimatedContentScope`](https://developer.android.com/reference/kotlin/androidx/compose/animation/AnimatedContentScope) receiver scope:**

```
enterTransition: (
    AnimatedContentScope<String>.(initial: NavBackStackEntry, target: NavBackStackEntry) -> EnterTransition?
)? = null,
```

This allows developers to use [`slideIntoContainer()`](https://developer.android.com/reference/kotlin/androidx/compose/animation/AnimatedContentScope#slideIntoContainer(androidx.compose.animation.AnimatedContentScope.SlideDirection,androidx.compose.animation.core.FiniteAnimationSpec,kotlin.Function1)) and [`slideOutOfContainer()`](https://developer.android.com/reference/kotlin/androidx/compose/animation/AnimatedContentScope#slideOutOfContainer(androidx.compose.animation.AnimatedContentScope.SlideDirection,androidx.compose.animation.core.FiniteAnimationSpec,kotlin.Function1)) as their `EnterTransition` and `ExitTransition`, respectively.

Note that because the `key` for `AnimatedContent` is the `id` of the `NavBackStackEntry`, this must be an `AnimatedContentScope<String>` and we still need to pass the `initial` and `target` entries.

**2. Content lambdas now use an [`AnimatedVisibilityScope`](https://developer.android.com/reference/kotlin/androidx/compose/animation/AnimatedVisibilityScope) receiver scope:**

```
content: @Composable AnimatedVisibilityScope.(NavBackStackEntry) -> Unit
```

This allows developers to use [`Modifier.animateEnterExit()`](https://developer.android.com/reference/kotlin/androidx/compose/animation/AnimatedVisibilityScope#(androidx.compose.ui.Modifier).animateEnterExit(androidx.compose.animation.EnterTransition,androidx.compose.animation.ExitTransition)) to specify element by element transitions, separate from the content level transitions.

Fixes #656


chrisbanes:
Unfortunately it looks like this PR doesn't work with the new `testStateSavedByCrossFade` test added by @jbw0033

ianhanniballake:
> Unfortunately it looks like this PR doesn't work with the new `testStateSavedByCrossFade` test added by @jbw0033

The coupling between the `targetState` of the `transition` and the `key` that `AnimatedContent` uses internally means we've needed to switch to a `AnimatedContentScope<String>` (using the `id` of the `NavBackStackEntry`) here. Unfortunately, that means we still need to pass the `initial` and `target` entries to each lambda, so the final API is not quite as intuitive here, but it does mean that saving state works and the test now passes.

jbw0033 (approved):

chrisbanes (approved):

PR #663, opened 2021-08-18
Fix remember saveable

jbw0033:


chrisbanes:
Looks like these changes were also part of #664? (I can't see any difference anyway)

PR #671, opened 2021-08-19
Add links to documentation on modules where it was missing

StylianosGakis:
Some modules were lacking the easy to find link inside their README files to the appropriate documentation that I found very neat in other modules. This commit adds the links in those modules too.


chrisbanes (approved):
Thanks!

PR #672, opened 2021-08-20
Remove reference to ScrollableColumn in doc

DSteve595:
What's `ScrollableColumn`? Never heard of it. There's no such thing as pre-1.0 Compose, and the API was perfect from the very first alpha.


chrisbanes (approved):
Woops!

PR #678, opened 2021-08-24
[Pager] Re-write Pager on top of LazyRow|Column

### LINKED ISSUE #566: [Idea] Pager keys
**Is your feature request related to a problem? Please describe.**
Seeing some issues with VerticalPager when the backing dataset is frequently updated with changes in order. Think of a list of chats, where you always sort them by the `lastMessageDate`.

**Describe the solution you'd like**
Besides some weird glitches when scrolling, which I'm investigating - will try to get back with repro instructions - I think it would be very helpful for us if we could use a `key()` with the chat ID. Accompanist currently uses the page index I think which is causing unnecessary invalidations during such order changes.

From my understanding, a big improvement for this use case would be having an API to override the default `key()` and reuse composables efficiently.

### LINKED ISSUE #667: [Idea] Add contentPadding to HorizontalPager
I'm trying to recreate this carousel

image (4)

Seems easy enough 

```
HorizontalPager(
    state = pagerState,
    horizontalAlignment = Alignment.Start,
    modifier = Modifier.fillMaxWidth(),
    itemSpacing = 16.dp) { i ->
    Box(modifier = Modifier.fillMaxWidth(0.9f)) { content[i]() }
}
```

If I add padding(start = 16.dp) to the HP modifier then I DO get the padding on the left side BUT when scrolling, all of the items get clipped when paging to the next item.

contentPadding was the recommended solution by Chris Banes.

### LINKED ISSUE #679: PagerState.targetPage scrolling bug
## Describe the bug
On scrolling left target page is equal current page

## Screenshots?
now it looks like
image
but on scrolling left: offset < 0f so target page is equal current page.
So need to change to(or absolute value of currentPageOffset<0.001f):
image

## Environment:
 - Accompanist version: [e.g. v0.17.0]

### LINKED ISSUE #706: [Pager] Offset screens are not recomposed when their state change
Hi!
I'm using HorizontalPager to make a DatePicker. When I start Pager layout, I already have a value picked, so I have a color mark for it. What happens next:
## To Reproduce
* scroll forward (initialOffscreenLimit can be any)
* Select another date and have a color mark on it
* Scroll back and still have color mark on initial pick date - **no recomposition happens**. Sure thing I use State for the date picked and change its value
* If then scroll forward out of initialOffscreenLimit and then scroll back to the first screen - recomposition **does** happen and my color mark is removed from the initial date.

So I wonder if there's a built-in way to force recomposition on pre-fetch screens (I assume no), any workaround you can advise, or either please add such a feature 🙏 Thank you

## Environment:
 - Accompanist version: 0.18.0

chrisbanes:
## Added features

The migration has allowed us to add some features.

- Ability to specify a `key` for each page. Fixes #566 
- Ability to set `contentPadding`. Fixes #667
- New `SnappingFlingBehavior` which can be used with any `Lazy*` layout (this is analogous to `SnapHelper` from RecyclerView). _Currently hidden._

## Removed features
Unfortunately, the migration has resulted in some features being removed:

- The `infiniteLooping` parameter has been removed. Instead a sample has been created showing you how to implement this.
- `offscreenLimit`. We no longer have any control of what items are laid out 'off screen'.

## API changes

- The `pageCount` parameter has moved from `PagerState` to `HorizontalPager` and `VerticalPager`
- Had to remove the animation spec parameters on `animateScrollToPage()` 

## Other things

- Likely fixes #706
- Fixes #679 

## Tasks
- [x] Re-enable `dragEnabled`
- [x] Add back some offset parameters
- [x] Fix `animateScrollToPage()`. Currently it only animates in one direction.
- [x] Update docs and guide

# Future tasks
- Move `SnappingFlingBehavior` to new module + docs + tests.
- Add ability to fling more than 1 page.


natario1:
Regarding

> offscreenLimit. We no longer have any control of what items are laid out 'off screen'.

Is it possible to know how Lazy* components currently compute the number of laid out items? `offscreenLimit` is very important when items use limited/expensive resources (say `MediaCodec`). Anything bigger than 1 would have a big impact on performance in our app.

andkulikov:
> Is it possible to know how Lazy* components currently compute the number of laid out items? `offscreenLimit` is very important when items use limited/expensive resources (say `MediaCodec`). Anything bigger than 1 would have a big impact on performance in our app.
LazyColumn/Row are composing the currently visible items + 2 previously visible items could be kept as the optimization to reuse for the next items during scrolling + 1 next item in the direction of scrolling could be pre-composed in advance as one more optimization

ColtonIdle:
@andkulikov so it looks like my example of trying to build an airbnb-like carousel will not work with HorizontalPager anymore. Do you have any ideas on how I can accomplish the snapping behavior with just a normal scrollable Row?

https://github.com/google/accompanist/pull/678#issuecomment-907190976

andkulikov:
> @andkulikov so it looks like my example of trying to build an airbnb-like carousel will not work with HorizontalPager anymore. Do you have any ideas on how I can accomplish the snapping behavior with just a normal scrollable Row?
> 
> [#678 (comment)](https://github.com/google/accompanist/pull/678#issuecomment-907190976)

There is indeed no way to replicate exactly offscreenLimit with LazyRow, however you can try a bit hacky way to expand the viewport size which LazyRow is using for the calculations. This allows to compose not next n items, but the next items needed to fill n of the extra width. If you set some large number it would compose all the items you have. See this [gist](https://gist.github.com/andkulikov/b0eda3e5e2e5d32ce307118dc8765d63)

ColtonIdle:
Interesting! That's definitely a clever way to go about it. I'm totally fine with using it, but I'm curious if you have any other recommendations for how to implement this (maybe this is better to ask on kotlinlang compose)

My only constraints are
1. Will be a small set of data. Never more than 10
2. Should have snapping in place (like a view pager)
3. Should be able to set the height of the carousel to the largest height item

andkulikov:
> Interesting! That's definitely a clever way to go about it. I'm totally fine with using it, but I'm curious if you have any other recommendations for how to implement this (maybe this is better to ask on kotlinlang compose)
> 
> My only constraints are
> 
> 1. Will be a small set of data. Never more than 10
> 2. Should have snapping in place (like a view pager)
> 3. Should be able to set the height of the carousel to the largest height item

Realistically I would recommend to not try to solve this issue that way at all. With options of:
1) Make items of the fixed size. Like force the text to always be of 2 or 3 lines. This can also involve making the width flexible instead, so if your text is not fitting into the 3 lines it takes more width. Or use elipsizing for the text.
2) Allow items to have different height and animate the height changes when it happens. Will Modifier.animateContentSize() work here?
It is usually hard to have the full control of what data you display. What if in the future this set of data will be populated by the backend and they will change the amount of items as they don't know about your limitation. Realistically even with your approach it will look a bit weird you have first items with the one lined text and the latest item which is not visible with like 5 lines of text. For the user it will just look like a big hole after the items. They can never scroll that LazyRow in order to understand why it is needed.

ColtonIdle:
Thanks. I _do_ think there are realistic use cases in which you would want to fit to the largest item, but I understand your thinking. I wanted to make sure I wasn't missing some other obvious solution here that I could do instead. If anything, I suppose I could just copy the code from this NonLazyHorizontalPager and paste it into my app. 😅

pager/src/main/java/com/google/accompanist/pager/PagerState.kt
```diff
@@ -248,403 +167,155 @@ class PagerState(
      *
      * To update the scroll position, use [scrollToPage] or [animateScrollToPage].
      */
-    val currentPageOffset: Float
-        get() = absolutePosition - currentPage
+    val currentPageOffset: Float by derivedStateOf {
+        currentLayoutPageInfo?.let { it.index + currentLayoutPageOffset - _currentPage } ?: 0f
```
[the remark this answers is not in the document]
andkulikov: would be nice to add a comment in the code, but ok if it works :)

pager/src/main/java/com/google/accompanist/pager/SnappingFlingBehavior.kt
```diff
@@ -0,0 +1,253 @@
...
+import androidx.compose.runtime.setValue
+import kotlin.math.abs
+import kotlin.math.absoluteValue
+
+/**
+ * Default values used for [SnappingFlingBehavior] & [rememberSnappingFlingBehavior].
+ */
+object SnappingFlingBehaviorDefaults {
+    /**
+     * TODO
```
andkulikov, line 44: do not forget to write something here :)

pager/src/main/java/com/google/accompanist/pager/SnappingFlingBehavior.kt
```diff
@@ -0,0 +1,253 @@
...
+                index = when {
+                    // If the velocity is greater than 1 page per second (velocity is px/s), spring
+                    // in the relevant direction
+                    initialVelocity > pageSize -> {
+                        (pageInfo.index + 1).coerceAtMost(lazyListState.layoutInfo.totalItemsCount - 1)
+                    }
+                    initialVelocity < -pageSize -> pageInfo.index
+                    // If the offset exceeds the scroll threshold (in either direction), we want to
+                    // move to the next/previous item
+                    pageInfo.offset < -pageSize / 2 -> pageInfo.index + 1
```
andkulikov, line 114: do we want to make sure we are no on the last item already, as you did few lines before

andkulikov (commented):
I think it is almost ready!

pager/src/main/java/com/google/accompanist/pager/SnappingFlingBehavior.kt
chrisbanes: Yep, because the velocity might be 0. 

This last 'fail-safe' check uses the position to determine the snap position.

chrisbanes:
> I think it is almost ready!

Nearly! All of the tests pass locally, but it looks like they're failing on CI. I'll take a look later and see what's up.

Then I'll start the process of splitting the snapping fling behaviour into a new module.

danielesegato:
Regarding
> ## Removed features
> * `infiniteLooping`. `Lazy*` doesn't easily support this.

do you know of any plan to re-introduce this behavior? Maybe adding support for it in the `Lazy*` widgets?

We were using the `infiniteLooping` feature for a carousel.

Thanks

pager/src/main/java/com/google/accompanist/pager/SnappingFlingBehavior.kt
chrisbanes: Added a comment to clarify this.

pager/src/main/java/com/google/accompanist/pager/Pager.kt
```diff
@@ -224,146 +213,178 @@ internal fun Pager(
...
+                        page = page,
+                        itemCount = state.pageCount,
+                        pagerState = state,
+                        isVertical = true,
+                        reverseLayout = reverseLayout,
+                        horizontalAlignment = horizontalAlignment,
+                        verticalAlignment = verticalAlignment,
+                        modifier = Modifier
+                            .nestedScroll(connection = ConsumeFlingNestedScrollConnection)
+                            .sizeIn(maxWidth = maxWidth, maxHeight = maxHeight),
```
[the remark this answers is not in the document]
chrisbanes: Done.

pager/src/main/java/com/google/accompanist/pager/Pager.kt
```diff
@@ -224,146 +213,178 @@ internal fun Pager(
...
-                // We can't rely on placeRelative() since that only uses the LayoutDirection, and
-                // we need to cater for our reverseLayout param too. reverseDirection contains
-                // the resolved direction, so we use that to flip the offset direction...
-                placeable.place(
-                    x = xCenterOffset + if (reverseDirection) -xItemOffset else xItemOffset,
-                    y = yCenterOffset + if (reverseDirection) -yItemOffset else yItemOffset,
-                )
+        // Report the start/end sizes back to the PagerState
+        if (page == 0) {
+            pagerState.layoutStartSpacing = if (isVertical) {
```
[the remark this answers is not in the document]
chrisbanes: Done.

andkulikov:
> Regarding
> 
> > ## Removed features
> > 
> > * `infiniteLooping`. `Lazy*` doesn't easily support this.
> 
> do you know of any plan to re-introduce this behavior? Maybe adding support for it in the `Lazy*` widgets?
> 
> We were using the `infiniteLooping` feature for a carousel.
> 
> Thanks

Theoretically it is possible to have the behaviour similar to infinite looping with Lazy lists. What you can do:

 ```
            val myList = listOf(0, 1, 2)
            val startIndex = Int.MAX_VALUE / 2
            val state = rememberLazyListState(initialFirstVisibleItemIndex = startIndex)
            LazyRow(state = state) {
                items(Int.MAX_VALUE) { rawIndex ->
                    val reminder = (rawIndex - startIndex) % myList.size
                    val index = if (reminder >= 0) reminder else myList.size + reminder
                    val item = myList[index]
                    Text("Item $item", Modifier.fillParentMaxSize())
                }
            }
```
Similar technique would be possible on top of the proposed API of Pagers

File: pager/src/main/java/com/google/accompanist/pager/PagerState.kt (hunk 1 above)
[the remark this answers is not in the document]
Comment (chrisbanes) [reply pager/src/main/java/com/google/accompanist/pager/PagerState.kt]: Comment added.

File: pager/src/main/java/com/google/accompanist/pager/Pager.kt [hunk 6]
```diff
@@ -97,75 +67,61 @@ object PagerDefaults {
      * @param snapAnimationSpec The animation spec to use when snapping.
      */
     @Composable
-    fun rememberPagerFlingConfig(
+    fun rememberPagerFlingBehavior(
```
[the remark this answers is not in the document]
chrisbanes: sgtm, changed.

ColtonIdle:
Losing some of these features really makes me wish that we would just have a lazy and non-lazy option for pagers.

chrisbanes:
> Losing some of these features really makes me wish that we would just have a lazy and non-lazy option for pagers.

We won't be maintaining two versions of this 😄. You're welcome to fork/copy and keep the non-lazy version.

andkulikov:
> > Theoretically it is possible to have the behaviour similar to infinite looping with Lazy lists.
> 
> Would that get messy with `key` though? @andkulikov

it makes it a bit strange, agree. but the hard limitation of keys is not to never have items with the same key, but to never have items with the same key composed together at the same time. which is probably true in the same way for any other possible pager implementation with infinite looping. like it will not work correctly if you only have 1 or 2 items

chrisbanes:
Just added a looping sample

ColtonIdle:
> > Losing some of these features really makes me wish that we would just have a lazy and non-lazy option for pagers.
> 
> We won't be maintaining two versions of this 😄. You're welcome to fork/copy and keep the non-lazy version.

Chris. You're free. You don't have to maintain even one version of this. 😂

pager/src/main/java/com/google/accompanist/pager/SnappingFlingBehavior.kt
chrisbanes: Hid this for now. Will make it public in a later PR.

chrisbanes:
Ready for review!

danielesegato:
> > > Theoretically it is possible to have the behaviour similar to infinite looping with Lazy lists.
> > 
> > 
> > Would that get messy with `key` though? @andkulikov
> 
> it makes it a bit strange, agree. but the hard limitation of keys is not to never have items with the same key, but to never have items with the same key composed together at the same time. which is probably true in the same way for any other possible pager implementation with infinite looping. like it will not work correctly if you only have 1 or 2 items

Thanks for this.
This is exactly what I was thinking of doing.

However, I believe the 3 items minimum is only true if each items takes the whole space, otherwise if you are showing part of the items before / after you are gonna need more items to avoid key collisions.

sample/src/main/java/com/google/accompanist/sample/pager/HorizontalPagerLoopingSample.kt
```diff
@@ -0,0 +1,110 @@
...
+            )
+        },
+        modifier = Modifier.fillMaxSize()
+    ) {
+        Column(Modifier.fillMaxSize()) {
+            // Display 10 items
+            val pageCount = 10
+
+            // We start the pager in the middle of the raw number of pages
+            val startIndex = Int.MAX_VALUE / 2
```
danielesegato, line 74: I think this should be
```kotlin
val startIndex = (Int.MAX_VALUE / 2).let { it - it.floorMod(pageCount) }
```

to start from the first index

danielesegato:
I've grabbed the code for the pager at the current state, it looks like it assume each page will always take all the available space?

I can't have the item before and after peak into vision anymore. Has this feature also been removed?

andkulikov:
> I can't have the item before and after peak into vision anymore. Has this feature also been removed?
I think setting contentPadding is a new way of making items to fill the area smaller then the viewport. This should enable nearest items to be partially visible

andkulikov (approved):

danielesegato:
> > I can't have the item before and after peak into vision anymore. Has this feature also been removed?
> 
> I think setting contentPadding is a new way of making items to fill the area smaller then the viewport. This should enable nearest items to be partially visible

Thanks, yes, I had just figured it out, I somehow found the previous behavior more intuitive but it does what I need. (I had just edited my previous comment)

This also means each page will need to have the same width (in an horizontal pager), am I right?

andkulikov:
> This also means each page will need to have the same width (in an horizontal pager), am I right?

Yes. If you have items of different size, or items which are not really filling the screen you probably need not a Pager, but just LazyRow with snapping behavior. Chris is going to make the snapping FlingBehavior for LazyRow public in one of the further commits

chrisbanes:
> Chris is going to make the snapping FlingBehavior for LazyRow public in one of the further commits

Yep, working on it right now.

ColtonIdle:
@chrisbanes sorry for the discussion on a closed item, but I'm back from vacation and just got to setup a proof of concept of where removing "offscreenLimit" is detrimental. I don't bring it up to argue, but instead to see if you have any idea on how to handle this sort of flow.

Essentially, my designer gave us a tutorial/wizard when first opening up the app. There is text pinned to the top, and image right below it and a button pinned to the bottom.

As you can see, if I have offscreenLimit set to 1 (default) then I get this result which my designer does **not** want.


https://user-images.githubusercontent.com/31751141/133387915-65b850c4-33e5-4292-b8cb-b0e8960f70ff.mp4

but if I take the offscreenLimit and set it to the size of the pager, then everything is laid out, and so the height of the horizontalPager is the height of the tallest item, which is what my designer wants. As you can see here:


https://user-images.githubusercontent.com/31751141/133388081-77c00ebd-dc17-4842-afb8-ba754cda7887.mp4

At this point, do you think the "lazy" horizontalPager that will be coming out in the next release in accompanist will be the wrong tool for the job for me or is there someway that I will be able to replicate this behavior using the lazy horizontalPager?

andkulikov:
> As you can see, if I have offscreenLimit set to 1 (default) then I get this result which my designer does **not** want.

There are so many different ways how to solve that. But in general, the design process is the collaboration between the designer and the engineer. They both can discuss all the trade offs and agree on the solution. As I mentioned earlier measuring all the items only to know the height is not a feasible solution for large item counts.
You can assume that you would always have small item count and use the hack I previously shared where you compose more items then visible.
You can say that we can't assume that there will be a small item count and discuss with the designer what other options you have. Again, I already mentioned that what you described is not always a better UX. What if your user never planned to scroll this pager at all? Then they will be surprised to see a huge gap between the Pager and the next component. It is not self-explanatory that this gap will be needed when you scroll and the large item would became visible. I would say that changing the heigh with the animation is probably more preferable solution in my opinion.
Or you can reorganize the screen so they items are not depending on each other size, like the component bellow the pager would always aligned to the bottom.

Do not be afraid to discuss the trade offs with your designers and propose options. If there is a technical limitation which negatively affects the performance sometimes it is better to tweak the design a bit to not make your app slow.

We could explore adding something like offscreenLimit to LazyRow in the future, but even if we do so I would not want our users to solve the use cases like you mentioned with that mechanism.

Tolriq:
Sorry to also post on a closed issue, but not having offscreenLimit means the non visible pages are not rendered and so things like images are no more preloaded. This cause delays and possible jank during scrolling as the screen is composed during the scrolling.
Is there's some easy workaround to address this ?

andkulikov:
> Sorry to also post on a closed issue, but not having offscreenLimit means the non visible pages are not rendered and so things like images are no more preloaded. This cause delays and possible jank during scrolling as the screen is composed during the scrolling. Is there's some easy workaround to address this ?

I expect that in most cases one nearest item will anyway be composed by the prefetching logic. What value was you passing to offscreenLimit previously?
You can also experiment with the workaround explained [here](https://gist.github.com/andkulikov/b0eda3e5e2e5d32ce307118dc8765d63)

Tolriq:
I was using just 1. And unless I missed a param there does not seems to have any prefetch.

Will try your workaround thanks.

ln-12:
Is there a workaround for the missing `dragEnabled` in the meantime?

andkulikov:
> Is there a workaround for the missing `dragEnabled` in the meantime?

You can use the same technique as used for LazyRow in general. See https://stackoverflow.com/questions/66502096/how-to-disable-and-enable-scrolling-in-lazycolumn-lazyrow-in-jetpack-compose

ln-12:
> You can use the same technique as used for LazyRow in general. See https://stackoverflow.com/questions/66502096/how-to-disable-and-enable-scrolling-in-lazycolumn-lazyrow-in-jetpack-compose

Thanks for the quick reply! I also saw that, but it would require to make [the state](https://github.com/google/accompanist/blob/f7f00520dc72482b6cfd3bc7e9981adf0af5045d/pager/src/main/java/com/google/accompanist/pager/PagerState.kt#L95) public (as mentioned above the line). To make it work, I could duplicate the state and pager classes locally in my project, but I wanted to avoid redundant code.

amenon:
> > > You can use the same technique as used for LazyRow in general. See https://stackoverflow.com/questions/66502096/how-to-disable-and-enable-scrolling-in-lazycolumn-lazyrow-in-jetpack-compose
> > 
> > 
> > Thanks for the quick reply! I also saw that, but it would require to make [the state](https://github.com/google/accompanist/blob/f7f00520dc72482b6cfd3bc7e9981adf0af5045d/pager/src/main/java/com/google/accompanist/pager/PagerState.kt#L95) public (as mentioned above the line). To make it work, I could duplicate the state and pager classes locally in my project, but I wanted to avoid redundant code.
> 
> No, you shouldn't need to do it. There is exactly the same [scroll function](https://github.com/google/accompanist/blob/f7f00520dc72482b6cfd3bc7e9981adf0af5045d/pager/src/main/java/com/google/accompanist/pager/PagerState.kt#L308) exposed right on PagerState.

I am attempting to use this method to disable dragging of the pager. However, it looks like this method blocks all gestures within the content of the pager. I cannot scroll vertically or tap buttons within the content. Here is the extension function I added:

```
/**
 * Use this to disable scrolling on a pager
 * eg:
 *   val scope = rememberCoroutineScope()
 *   val pagerState = rememberPagerState()
 *   pagerState.disableScrolling(scope)
 */
@OptIn(ExperimentalPagerApi::class)
fun PagerState.disableScrolling(scope: CoroutineScope) {
    scope.launch {
        scroll(scrollPriority = MutatePriority.PreventUserInput) {
            // Await indefinitely, blocking scrolls
            awaitCancellation()
        }
    }
}
```

Am I missing something here @andkulikov?

PR #694, opened 2021-09-04
ADD: separate shimmer effect

shenzhen2017:
Skeleton screen effect similar to [facebook open source project](https://github.com/facebook/shimmer-android)
The skeleton screen is usually animated in a line rather than an element,so it is best to separate it from the placeholder

The animation effect is as follows

https://user-images.githubusercontent.com/18047645/132097124-0d4dece6-6c90-4700-8fe0-1dbf4798c958.mp4

https://user-images.githubusercontent.com/18047645/132097129-7520ea64-cbc9-4764-85dd-e52bcc6b9f0d.mp4

### DIFF
--- a/placeholder/build.gradle
+++ b/placeholder/build.gradle
@@ -90,6 +90,7 @@ dependencies {
     implementation libs.compose.foundation.foundation
     implementation libs.compose.ui.util
     implementation libs.napier
+    implementation libs.androidx.core
 
     // ======================
     // Test dependencies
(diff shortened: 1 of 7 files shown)


shenzhen2017:
Can someone help me review the PR?

JoseAlcerreca:
Hi!

This is a pretty impressive feature, many thanks for the PR!

However, we think the level of customization is not needed for most apps, so it might make more sense a a separate artifact in a different repo (like https://github.com/shenzhen2017/ComposeShimmer) 

Thanks again!
Jose

PR #696, opened 2021-09-06
Added cutout padding modifier

CaptnBlubber:
This will create a new easy to use modifier for the newly exposed Cutout Insets from #685.

Additionally I have added the cutouts to the existing test


chrisbanes (approved):
Thanks!

PR #716, opened 2021-09-14
Remove swiperefresh's dependency on ui-tooling

mxalbert1996:
As it is not used and will pull in unnecessary dependencies for consumers.


chrisbanes (approved):
Thanks!

PR #721, opened 2021-09-15
[WIP] Move SnappingFlingBehavior to new module

chrisbanes:
Add new functionality which allows setting:

- The snap points for the items, along with alignment-style defaults.
- Maximum fling distance in pixels.

Still very work-in-progress, and the tests don't currently pass.


lazysnap/build.gradle
```diff
@@ -0,0 +1,113 @@
...
+        }
+        androidTest {
+            java.srcDirs += 'src/sharedTest/kotlin'
+            res.srcDirs += 'src/sharedTest/res'
+        }
+    }
+}
+
+dependencies {
+    implementation libs.compose.foundation.foundation
```
andkulikov, line 85: should be api

lazysnap/src/main/java/com/google/accompanist/lazysnap/SnappingFlingBehavior.kt
```diff
@@ -0,0 +1,460 @@
...
+     * TODO
+     */
+    val Center: LazyListLayoutInfo.(LazyListItemInfo) -> Int = {
+        ((viewportEndOffset + viewportStartOffset) / 2) - (it.size / 2)
+    }
+
+    /**
+     * TODO
+     */
+    val End: LazyListLayoutInfo.(LazyListItemInfo) -> Int = { viewportEndOffset - it.size }
```
andkulikov, line 118: this api is a bit confusing. how it was decided what is the receiver and what is the param? I would prefer both to be a param I think

lazysnap/src/main/java/com/google/accompanist/lazysnap/SnappingFlingBehavior.kt
```diff
@@ -0,0 +1,460 @@
...
+        } else {
+            // Otherwise we look at the current offset, and spring to whichever is closer
+            val snapOffset = snapOffsetForItem(lazyListState.layoutInfo, currentItemInfo)
+            if (currentItemInfo.offset < snapOffset - (currentItemInfo.size / 2)) {
+                currentItemInfo.index + 1
+            } else currentItemInfo.index
+        }.coerceIn(0, lazyListState.layoutInfo.totalItemsCount - 1)
+    }
+
+    private suspend fun ScrollScope.performSpringFling(
```
andkulikov, line 312: this and performDecayFling share quite a lot of code, right? maybe we can try to extract some reusable parts?

chrisbanes: Yep, it took a bit of scope gymnastics but managed to get them sharing most of their impl

lazysnap/src/main/java/com/google/accompanist/lazysnap/SnappingFlingBehavior.kt
chrisbanes: Moved `LazyListLayoutInfo` to be a param. Less confusing? My thought is that most apps will use one of these provided impls, but they can easily do something custom if they wish.

lazysnap/build.gradle
chrisbanes: Done.

lazysnap/src/main/java/com/google/accompanist/lazysnap/SnappingFlingBehavior.kt
andkulikov: yep, looks good

chrisbanes:
This is mostly ready for review! Still got some docs + test tasks to do, but the bulk of the implementation is done.

lazysnap/src/main/java/com/google/accompanist/lazysnap/SnappingFlingBehavior.kt
```diff
@@ -0,0 +1,689 @@
...
+/**
+ * Default values used for [SnappingFlingBehavior] & [rememberSnappingFlingBehavior].
+ */
+@ExperimentalLazySnapApi
+object SnappingFlingBehaviorDefaults {
+    /**
+     * [AnimationSpec] used as the default value for the `snapAnimationSpec` parameter on
+     * [rememberSnappingFlingBehavior] and [SnappingFlingBehavior].
+     */
+    val snapAnimationSpec: AnimationSpec<Float> = spring(stiffness = 400f)
```
andkulikov, line 67: use PascalCase for such constants names

lazysnap/src/main/java/com/google/accompanist/lazysnap/SnappingFlingBehavior.kt
```diff
@@ -0,0 +1,689 @@
...
+ */
+@ExperimentalLazySnapApi
+@Composable
+fun rememberSnappingFlingBehavior(
+    lazyListState: LazyListState,
+    snapOffsetForItem: (layoutInfo: LazyListLayoutInfo, itemInfo: LazyListItemInfo) -> Int = SnapOffsets.Center,
+    maximumFlingDistance: (LazyListLayoutInfo) -> Int = SnappingFlingBehaviorDefaults.maximumFlingDistance,
+    decayAnimationSpec: DecayAnimationSpec<Float> = rememberSplineBasedDecay(),
+    snapAnimationSpec: AnimationSpec<Float> = SnappingFlingBehaviorDefaults.snapAnimationSpec,
+    @Px endContentPadding: Int = 0,
```
andkulikov, line 98: users will probably have it in Dps, do we have a Density anywhere so we can convert it?

lazysnap/src/main/java/com/google/accompanist/lazysnap/SnappingFlingBehavior.kt
```diff
@@ -0,0 +1,689 @@
...
+): SnappingFlingBehavior = remember(
```
andkulikov, line 99: we use `snap` in the module name and in SnapOffsets class, but here we use `snapping`. maybe we can unify that, does it sound bad to just call it SnapFlingBehavior? in RV it was called SnapHelper

lazysnap/src/main/java/com/google/accompanist/lazysnap/SnappingFlingBehavior.kt
```diff
@@ -0,0 +1,689 @@
...
+    )
+}.apply {
+    this.endContentPadding = endContentPadding
+}
+
+/**
+ * Contains a number of values which can be used for the `snapOffsetForItem` parameter on
+ * [rememberSnappingFlingBehavior] and [SnappingFlingBehavior].
+ */
+@Suppress("unused")
```
andkulikov, line 121: why is it needed?

lazysnap/src/main/java/com/google/accompanist/lazysnap/SnappingFlingBehavior.kt
```diff
@@ -0,0 +1,689 @@
...
+            // we fallback to the end offset. That value contains the end content padding though.
+            viewportEndOffset
+        }
+}
+
+/**
+ * A snapping [FlingBehavior] for [LazyListState]. Typically this would be created
+ * via [rememberSnappingFlingBehavior].
+ *
+ * Note: the default parameter value for [decayAnimationSpec] is different to the value used in
```
andkulikov, line 165: I would prefer to not provide a default value instead for this param

lazysnap/src/main/java/com/google/accompanist/lazysnap/SnappingFlingBehavior.kt
```diff
@@ -0,0 +1,689 @@
...
+            }
+        )
+
+        return velocityLeft
+    }
+
+    /**
+     * Returns true if we needed to perform a snap back, and the animation should be cancelled.
+     */
+    private inline fun AnimationScope<Float, AnimationVector1D>.checkSnapBack(
```
andkulikov, line 504: you use this function twice and it is quite large, lets remove the inline modifier

lazysnap/src/main/java/com/google/accompanist/lazysnap/SnappingFlingBehavior.kt
```diff
@@ -0,0 +1,689 @@
...
+        }
+
+        return false
+    }
+
+    private val currentItemInfo: LazyListItemInfo?
+        get() = lazyListState.layoutInfo.let { layoutInfo ->
+            layoutInfo.visibleItemsInfo.asSequence()
+                .filter { it.offset <= snapOffsetForItem(layoutInfo, it) }
+                .lastOrNull()
```
andkulikov, line 546: can you use lastOrNull { predicate }?

lazysnap/src/main/java/com/google/accompanist/lazysnap/SnappingFlingBehavior.kt
andkulikov, line 504: probably rename to something like snapBackIsNeeded()

lazysnap/src/main/java/com/google/accompanist/lazysnap/SnappingFlingBehavior.kt
```diff
@@ -0,0 +1,689 @@
...
+    val End: (LazyListLayoutInfo, LazyListItemInfo) -> Int = { layoutInfo, itemInfo ->
+        layoutInfo.layoutSize - itemInfo.size
+    }
+
+    /**
+     * Ideally this would exist on [LazyListLayoutInfo] but it doesn't right now.
+     * Raised https://issuetracker.google.com/issues/200920410 to track.
+     */
+    private val LazyListLayoutInfo.layoutSize: Int
+        get() = if (this is LazyListLayoutInfoWithContentPadding) {
```
andkulikov, line 150: that is a clever idea, I like it

however I though about it more and I guess we shouldn't do it. the reason is that once we add a new property in LazyListLayoutInfo interface it will be not really a binary compatible change as any external implementations of LazyListLayoutInfo will break - they do not implement a new property. we even internally discussing using sealed interfaces for such use cases. which means we should probably not implement LazyListLayoutInfo. which means that we need to provide all the needed info as a param for the lambda blocks you have.
first approach is to just add endContentPadding as a third param, which makes it less clean, agree
second approach could be to try to abstract it away from Lazy apis and introduce our own class which contains all the needed info for calculating snap offsets. like in your current implementations you only used layout size and item size, so simple class containing those two vals will work. if you think about it in fact FlingBehavior is not a concept of lazy lists. this could be also provided for just Modifier.horizontalScroll(), and theoretically why can't we use SnappingFlingBehavior there? maybe the long term approach would be to not use the info from LazyListState directly, but map it into some new representation which SnappingFlingBehavior uses internally. this will also allow the api to be used with lazy grids, for example, as they would have their own state and layout info classes

lazysnap/src/main/java/com/google/accompanist/lazysnap/SnappingFlingBehavior.kt
chrisbanes: Nope, we don't unfortunately (AFAIK)

lazysnap/src/main/java/com/google/accompanist/lazysnap/SnappingFlingBehavior.kt
chrisbanes: Done

lazysnap/src/main/java/com/google/accompanist/lazysnap/SnappingFlingBehavior.kt
chrisbanes: Because library mode doesn't work in Android projects, so we can't declare something as public API. Some of vals in the class aren't used in the project, so lint complains that we can remove them. Added a comment explaining.

lazysnap/src/main/java/com/google/accompanist/lazysnap/SnappingFlingBehavior.kt
chrisbanes: Done.

lazysnap/src/main/java/com/google/accompanist/lazysnap/SnappingFlingBehavior.kt
chrisbanes: Good catch. Done.

lazysnap/src/main/java/com/google/accompanist/lazysnap/SnappingFlingBehavior.kt
chrisbanes: Done.

lazysnap/src/main/java/com/google/accompanist/lazysnap/SnappingFlingBehavior.kt
chrisbanes: Sounds good. Done.

lazysnap/src/main/java/com/google/accompanist/lazysnap/SnappingFlingBehavior.kt
chrisbanes: Done.

lazysnap/src/main/java/com/google/accompanist/lazysnap/SnappingFlingBehavior.kt
andkulikov: this function is a composable one, so we can at least take it as a Dp here and convert using LocalDensity.current. we can even check for 0 first as an optimization to not read the composition local when not needed

lazysnap/src/main/java/com/google/accompanist/lazysnap/SnappingFlingBehavior.kt
chrisbanes: Yeah I agree on wanting to generalize this, but I don't think we should do it too early. The abstractions we write now may not work for the future grid APIs, etc.

I think we should stick to the `endContentPadding` thing for now (and focus on `LazyListState`), and we can work out how to generalize later. WDYT?

andkulikov: yeah, whatever you think is reasonable which is not requiring implementing the interface is ok to me

Issue #725, opened 2021-09-16
Incorrect lifecycle events happen for BottomSheet


WonderCsabo:
First of all, i am not sure if this is a bug in this library, or navigation(-compose). `ModalBottomSheetLayout` works fine without navigation, though.

## Describe the bug
When the bottom sheet is shown, only `ON_START` lifecycle event is reported inside it. Once it is hidden, `ON_RESUME` arrives. At this point, `ON_RESUME` events are reported to the bottom sheet, but only on start to the main screen.

## To Reproduce
I added to a minimal lifecycle event reporting code to the [official sample](https://google.github.io/accompanist/navigation-material/).

```kotlin
                val navController = rememberNavController()
                val bottomSheetNavigator = rememberBottomSheetNavigator()
                navController.navigatorProvider += bottomSheetNavigator
                ModalBottomSheetLayout(bottomSheetNavigator) {
                    NavHost(navController, "home") {
                        composable(route = "home") {
                            Box(modifier = Modifier.fillMaxSize(), contentAlignment = Alignment.Center) {
                                Text(
                                    text = "this is home",
                                    modifier = Modifier.clickable {
                                        navController.navigate("sheet")
                                    }
                                )
                            }

                            val observer = remember {
                                LifecycleEventObserver { _, event ->
                                    Timber.i("home lifecycle $event")
                                }
                            }
                            val lifecycle = LocalLifecycleOwner.current.lifecycle
                            DisposableEffect(lifecycle, observer) {
                                lifecycle.addObserver(observer)
                                onDispose { lifecycle.removeObserver(observer) }
                            }
                        }
                        bottomSheet(route = "sheet") {
                            Box(
                                modifier = Modifier
                                    .fillMaxWidth()
                                    .height(300.dp)
                            ) {
                                Text("This is a cool bottom sheet!")
                            }

                            val observer = remember {
                                LifecycleEventObserver { _, event ->
                                    Timber.i("bottomsheet lifecycle $event")
                                }
                            }
                            val lifecycle = LocalLifecycleOwner.current.lifecycle
                            DisposableEffect(lifecycle, observer) {
                                lifecycle.addObserver(observer)
                                onDispose { lifecycle.removeObserver(observer) }
                            }
                        }
                    }
                }
 ```
                
When we start the app, these are reported:
```
home lifecycle ON_CREATE
home lifecycle ON_START
home lifecycle ON_RESUME
```

When we show the bottom sheet
```
home lifecycle ON_PAUSE
[...]

Issue #720, opened 2021-09-15
Dismiss bottom sheet navigation by swiping or touch scrim area, not removed from navigation stack


wisnukurniawan:
## Describe the bug
Dismiss bottom sheet navigation by swiping or touch scrim area, not removed from navigation stack. 

## To Reproduce
Let say we have 2 bottom sheet.
1. Show first bottom sheet, dismiss it by swiping
2. Repeat step 1 for second bottom sheet
3. First bottom sheet will pop up

## Expected behavior
First bottom sheet should be not pop up

## Screenshots?
<img src="https://user-images.githubusercontent.com/16729158/133418272-eab07199-5fb3-4c21-88ee-759ee7ae3194.gif" width="180">

## Environment:
 - Accompanist version: v0.18.0

## Additional context
Not happen in accompanist **v0.17.0**.

jossiwolf:
See #700 - fix is in progress. Sorry for the inconvenience!

chrisbanes:
Will be fixed in the next release (probably today)

wisnukurniawan:
yey! thanks a lot @jossiwolf @chrisbanes

Issue #700, opened 2021-09-07
[navigation-material] Sometimes bottom sheet entry not correctly popped when dismissed by user


mxalbert1996:
## Describe the bug

Sometimes when the user dismiss the bottom sheet, it pops up again. 

https://user-images.githubusercontent.com/9391933/132286583-4858b9f9-d859-414f-bc3c-8849d2a71e3d.mp4

## To Reproduce

The [official example](https://github.com/google/accompanist/blob/main/sample/src/main/java/com/google/accompanist/sample/navigation/material/BottomSheetNavSample.kt) is enough to reproduce this.

## Findings

It seems that the first time we navigate to a bottom sheet entry, the [`sheetState.show()`](https://github.com/google/accompanist/blob/dd384f0b419bb5b95b2cda608520d2866bb248c7/navigation-material/src/main/java/com/google/accompanist/navigation/material/SheetContentHost.kt#L96) call is always canceled, but `sheetState.targetValue` is `Hidden`, so [`currentOnSheetShown(backStackEntry)`](https://github.com/google/accompanist/blob/dd384f0b419bb5b95b2cda608520d2866bb248c7/navigation-material/src/main/java/com/google/accompanist/navigation/material/SheetContentHost.kt#L103) which marks the transition completed is not called. However, the bottom sheet does pop up after that, making the internal and external states inconsistent, which eventually results in the behavior described above.

jossiwolf:
Thanks for the report! I believe this issue was introduced/resurfaced with #680. The underlying issue is caused by the Swipeable APIs but I'll try to find a workaround next week.

TadejZupancic:
If you click outside the bottom sheet to close it, it closes it, but it does not change the destination. The previous destination that it should pop to can still be used, so it looks like it is working fine. I've noticed it because in my app I have a toolbar button that shows only on that destination and also the bottom navigation does not show any item selected after the bottom sheet is closed.

If you close the bottom sheet with a back press, then it works fine.

I've tried it on the sample app and it also stays on the SHEET destination if you click outside the bottom sheet.

mxalbert1996:
What's the point of closing the the bottom sheet without changing the destination?
Besides, this is obviously neither the intended behavior of the source code nor the behavior before 0.18.0.

TadejZupancic:
That's obviously a bug, I've updated my post as I see that it could be understood differently as I've intended.

rubenquadros12:
[...]

PR #741, opened 2021-09-28
[Navigation Material] Fix Bottom Sheet Re-Appearing after Dismissal

jossiwolf:
Our `show` calls are getting cancelled due to a race condition (https://issuetracker.google.com/issues/200980998) so we can't rely on their completion as a signal that the sheet is shown and we can mark the sheet's back stack entry's transition as complete. I talked to Matvei and he suggested using `withFrameNanos` which works. It's more of a workaround/hack and I'd like to put a better solution in place soon, but this works for the time being.

Test: Existing tests still pass (as they did despite the race condition...). Will try to create tests that fail before the fix. Manual testing apart from that.
Fixes: #725, #720, #700

### DIFF
--- a/navigation-material/src/androidTest/java/com/google/accompanist/navigation.material/SheetContentHostTest.kt
+++ b/navigation-material/src/androidTest/java/com/google/accompanist/navigation.material/SheetContentHostTest.kt
@@ -137,16 +137,19 @@ internal class SheetContentHostTest {
     @Test
     fun testOnSheetShownCalled_onBackStackEntryEnter() = runBlockingTest(testClock) {
         val sheetState = ModalBottomSheetState(ModalBottomSheetValue.Hidden)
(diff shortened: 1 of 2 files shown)


ianhanniballake (approved):
What a hack. If it works, it works though.

If you can, I'd strongly suggest working on upstreaming a fix for this cancellation issue into Compose 1.1 (which we might be able to cherry pick to a Compose 1.0.X bug fix release) as I'd really prefer to remove these hacks and the need for `ExperimentalCoroutinesApi` here.

jossiwolf:
There's nothing I'd love more but I don't see where we could apply this to bottom sheet to improve :/ Trying to contribute another though and I'm hoping that that'll improve things!

chrisbanes (approved):

chrisbanes:
Closing as this was merged as part of #742

PR #742, opened 2021-09-28
[Navigation Material] Fix Bottom Sheet Fully Expanding

jossiwolf:
Kinda similar to #741 - we were calling `show` too early. That meant that the `SwipeableState`/`ModalBottomSheetState`'s `anchors` wouldn't have been calculated properly yet. By default, a sheet will have two anchors: `Hidden` and `Expanded`. If `sheetHeight > containerHeight / 2`, the sheet should have a `HalfExpanded` state. For that, `ModalBottomSheetLayout` needs to know the sheet's height, making it two-pass. The effect we use to `show` the sheet got called on the first pass though, when the sheet content is composed to measure it, but the anchors didn't have the correct sheet height yet, causing the sheet to fully expand despite it being tall enough to have a `HalfExpanded` state.

This isn't a nice solution but in agreement with Matvei, it seems to be the most reasonable fix for the moment.

### DIFF
--- a/navigation-material/src/main/java/com/google/accompanist/navigation/material/SheetContentHost.kt
+++ b/navigation-material/src/main/java/com/google/accompanist/navigation/material/SheetContentHost.kt
@@ -33,15 +33,19 @@ import androidx.compose.runtime.rememberUpdatedState
 import androidx.compose.runtime.saveable.SaveableStateHolder
 import androidx.compose.runtime.setValue
 import androidx.compose.runtime.snapshotFlow
+import androidx.compose.runtime.withFrameNanos
 import androidx.compose.ui.Modifier
 import androidx.compose.ui.unit.dp
 import androidx.navigation.NavBackStackEntry
 import androidx.navigation.compose.LocalOwnersProvider
+import kotlinx.coroutines.CancellationException
+import kotlinx.coroutines.ExperimentalCoroutinesApi
 import kotlinx.coroutines.flow.collect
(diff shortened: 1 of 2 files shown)


navigation-material/src/main/java/com/google/accompanist/navigation/material/SheetContentHost.kt
```diff
@@ -86,20 +92,48 @@ internal fun SheetContentHost(
...
+                    repeat(AWAIT_FRAMES_BEFORE_SHOW) { awaitFrame() }
                     sheetState.show()
+                } catch (sheetShowCancelled: CancellationException) {
+                    // There is a race condition in ModalBottomSheetLayout that happens when the
+                    // sheet content changes due to the anchors being re-calculated. This causes an
+                    // animation to run for a short time, cancelling any currently running animation
+                    // such as the one triggered by our `show` call.
+                    // The sheet will still snap to the EXPANDED or HALF_EXPANDED state.
+                    // In that case we want to wait until the sheet is visible. For safety, we only
+                    // wait for 800 seconds - if the sheet is not visible until then, something has
```
hrach, line 123: milliseconds

jossiwolf: That is... a good point😂 thanks!

ianhanniballake (approved):
Gross.

chrisbanes (approved):

Issue #724, opened 2021-09-16
[Idea] Navigating back from a bottom sheet should animate


WonderCsabo:
I am not sure if this is a bug or not, i open it as a feature request just to be sure. :)

**Current behaviour:**

Currently when you close a bottom sheet using `navController.popBackStack()`, the bottom sheet immediately disappears. In the code, i see that `snapTo()` is used, and also the sheet is replaced with an empty view.

**Desired behaviour:**

When `navController.popBackStack()` is called, the sheet should disappear with an animation, just like when we call `animateTo()`.

I understand that we could just call `animateTo()` instead of navigation, however that state is not generally available, and should not be used in VM logic anyway.

jossiwolf:
`snapTo` is used because of a bug in `ModalBottomSheetLayout` that hadn't been fixed last time I checked - I'll recheck if we can use `animateTo` now :)

WonderCsabo:
Thank you! And also thanks for the great library by the way. 🙏

WonderCsabo:
@jossiwolf what should we do with this issue? The Github bot is always complaining. :)

jossiwolf:
@WonderCsabo Sorry, we reconfigured the bot now :) The fix for this depends on a larger refactoring that I'm still working on, so there is no good quick fix in the meantime. Sorry about that!

WonderCsabo:
No worries! Thanks for the update!

jossiwolf:
Obviously, the exemption rule doesn't seem to work🙈 This is still dependent on a larger in-progress refactoring and will take a bit to resolve. Just for further explanation: when navigating, the back stack is updated. We listen for back stack changes and update the sheet's content based on that, which causes the anchors to be recalculated. This means that a longer-running animation (triggered by `animateTo`) will be cancelled, aka the sheet will not be hidden, while a quick animation (`snapTo`) will finish. Unfortunately, this is a known issue upstream that is not trivial to fix, so until we get our own implementation in, there is no viable way to fix this issue.

leinardi:
Is there a link to the upstream bug that we can subscribe to?

alfroid2k:
Are there any updates on this issue? any workaround?

jossiwolf:
No workaround for now, unfortunately. The overall upstream issue can be tracked here: https://issuetracker.google.com/issues/167966118

Luis-Gangtime:
[...]

PR #743, opened 2021-09-28
[Navigation Material] Animate Bottom Sheet Dismissal

jossiwolf:
We're now using ModalBottomSheetState#hide instead of snapTo which will give us an animation when popping the back stack/dismissing the sheet!

Fixes: #724


jossiwolf:
Looks like the synchronization between nulling out the sheet content and hiding the sheet isn't working out here. We'll keep using `snapTo` for now - I have a way to refactor our code a bit in mind that I'll give a shot soon.

PR #750, opened 2021-09-30
[Pager] Stop consuming nested flings in cross-axis

### LINKED ISSUE #740: Make Pager only consume fling on the corresponding axis
Currently, when using either `HorizontalPager` or `VerticalPager`, the pager itself consumes all fling and scroll in *both* axis, which breaks nested scrolling hierarchy, for example, when using a `LazyColumn` inside a `HorizontalPager` within a vertically scrollable `Column`, like:
```kotlin
Column(Modifier.verticalScroll(scrollState)){
    HorizontalPager(pagerState) {
        LazyColumn {
            /*List content*/
        }
    }
}
```

**Proposal**
Transform `ConsumeFlingNestedScrollConnection` object into a class, or otherwise specify the axis in which the fling and scroll are to be consumed.

chrisbanes:
Currently Pager consumes nested scrolls and flings in
all directions. This PR changes that so that it only
consumes nested scrolls in the main scrolling axis.

Fixes #740

### DIFF
--- a/pager/src/main/java/com/google/accompanist/pager/Pager.kt
+++ b/pager/src/main/java/com/google/accompanist/pager/Pager.kt
@@ -219,6 +219,13 @@ internal fun Pager(
 
     val pagerScope = remember(state) { PagerScopeImpl(state) }
 
+    // We only consume nested flings in the main-axis, allowing cross-axis flings to propagate
+    // as normal
+    val consumeFlingNestedScrollConnection = ConsumeFlingNestedScrollConnection(
+        consumeHorizontal = !isVertical,
+        consumeVertical = isVertical,
+    )
+
     if (isVertical) {
         LazyColumn(
             state = state.lazyListState,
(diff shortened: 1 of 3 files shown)


chrisbanes:
@andkulikov @matvei-g

pager/src/main/java/com/google/accompanist/pager/Pager.kt
```diff
@@ -278,24 +285,44 @@ internal fun Pager(
     }
 }
 
-private object ConsumeFlingNestedScrollConnection : NestedScrollConnection {
+private class ConsumeFlingNestedScrollConnection(
+    private val consumeHorizontal: Boolean,
+    private val consumeVertical: Boolean,
```
matvei-g, line 290: just curious: why is `isVertical: Boolean` not enough? :)

pager/src/main/java/com/google/accompanist/pager/Pager.kt
```diff
@@ -219,6 +219,13 @@ internal fun Pager(
 
     val pagerScope = remember(state) { PagerScopeImpl(state) }
 
+    // We only consume nested flings in the main-axis, allowing cross-axis flings to propagate
+    // as normal
+    val consumeFlingNestedScrollConnection = ConsumeFlingNestedScrollConnection(
```
matvei-g, line 224: It's a class now, let's remember it as it is very important to remember the connection so we do less work reattaching it.

matvei-g (approved):
Thanks Chris, look good to me. One nitpick and one comment about connection.

LGTM given that it's fixed.

pager/src/main/java/com/google/accompanist/pager/Pager.kt
chrisbanes: It would have been. I just wanted to keep it more generalised, so people can copy-pasta it if they need, etc.

PR #770, opened 2021-10-04
Update to Compose SNAPSHOT 7790263

nickbutcher:


manuelvicnt (approved):
LGTM but we need to discuss how we're going to release 1.1 alpha versions

PR #816, opened 2021-10-24
[WIP] Add Google Maps wrapper

bentrengrove:
Adds a basic Google Maps wrapper Composable. 
Supports Compose state of latitude, longitude and zoom.

Putting this up for feedback on the approach I am taking. I have used the `SupportMapFragment` instead of just the plain `MapView` so that the maps library handles saving its own state. I don't think there is any problem in using `FragmentContainerView` inside an `AndroidView` composable that I know of. There is a lot of state not currently being managed by Compose that people might run into, things such as adding Pins or Shapes to the map. By using a fragment these should at least survive rotation.

I think the dream Compose API for Maps would be something along the lines of
```
Map(lat, lng) {
   Pin(pinLat, pinLng)
}
```

With a `MapScope.()->Unit content` parameter that made all the features of Google Maps composable and declarative. This was too much effort for this first PR so right now the Composable has some callbacks that allow you to either add things to the map on first load, or on each update.

TODO:
- Tests
- Docs


maps/src/main/java/com/google/accompanist/maps/GoogleMaps.kt
```diff
@@ -0,0 +1,109 @@
...
+    modifier: Modifier = Modifier,
+    onMapReady: (GoogleMap) -> Unit = {},
+    onMapUpdated: (GoogleMap) -> Unit = {},
+) {
+    val context = LocalContext.current
+    val scope = rememberCoroutineScope()
+
+    val cameraPosition = CameraPosition.fromLatLngZoom(LatLng(mapState.latitude, mapState.longitude), mapState.zoom)
+
+    LaunchedEffect(key1 = context, block = {
```
curioustechizen, line 83: I wonder whether it would be possible to achieve the same result by moving some of this into the `factory` lambda of `AndroidViewBinding`? For example that would ensure `onMapReady` is called exactly once in map inflation.

bentrengrove: Yeah I agree. Definitely doesn't feel right at the moment now there is nothing obvious to use as a key like there was before when I was just using `AndroidView`. Will try move things in to the factory

bentrengrove:
Closing this one for now while I work out these issues

PR #821, opened 2021-10-26
Accept a lamda to run after requesting permissions

jdkoren:
Callers of rememberPermissionState can provide a lamda to run with the
result after using launchPermissionRequest. Similar addition for
rememberMultiplePermissionsState and launchMultiplePermissionsRequest.


bentrengrove:
Superseded by #990

PR #822, opened 2021-10-27
[Web] Add WebView wrapper

bentrengrove:
Adds a basic WebView wrapper and sample.

The WebView supports loading Urls and capturing back presses automatically to navigate back automatically. It is only meant as a basic version of WebView and not a full replacement. For most cases people though I believe this will be enough and valuable enough to add to Accompanist.

The sample provided is a super basic web browser.


bentrengrove:
Looks like there is no WebView on the test AVD

web/src/androidTest/kotlin/com/google/accompanist/web/DummyTests.kt
```diff
@@ -0,0 +1,57 @@
...
+
+import org.junit.Test
+import org.junit.runner.RunWith
+import org.junit.runners.JUnit4
+
+/**
+ * Dummy tests to help with sharding: https://github.com/android/android-test/issues/973
+ */
+@RunWith(JUnit4::class)
+class DummyTests {
```
nickbutcher, line 27: Rename to be more inclusive https://developers.google.com/style/word-list?hl=en#dummy-variable

web/src/androidTest/kotlin/com/google/accompanist/web/WebTest.kt
```diff
@@ -0,0 +1,152 @@
...
+        lateinit var state: WebViewState
+
+        rule.setContent {
+            state = rememberWebViewState(uri = Uri.parse(LINK_URL))
+            WebTestContent(
+                state
+            ) { state.content = it }
+        }
+
+        rule.waitForIdle()
```
JoseAlcerreca, line 75: Add a comment explaining why you need this. I guess it's because you're using a Espresso finder, which doesn't wait for composition idleness

web/src/main/java/com/google/accompanist/web/WebView.kt
```diff
@@ -0,0 +1,164 @@
...
+    var view by remember { mutableStateOf<WebView?>(null) }
+    var webViewLoaded by remember(view) { mutableStateOf(false) }
+
+    BackHandler(captureBackPresses && state.canGoBack) {
+        view?.goBack()
+    }
+
+    AndroidView(
+        factory = {
+            WebView(it).apply {
```
nickbutcher, line 64: Name `context` param for clarity?

web/src/main/java/com/google/accompanist/web/WebView.kt
```diff
@@ -0,0 +1,164 @@
...
+ */
+@Composable
+fun WebView(
+    state: WebViewState,
+    onContentChanged: (WebContent) -> Unit,
+    modifier: Modifier = Modifier,
+    captureBackPresses: Boolean = true,
+    onCreated: (WebView) -> Unit = {}
+) {
+    var view by remember { mutableStateOf<WebView?>(null) }
```
nickbutcher, line 55: WDYT about renaming `webView`?

web/src/main/java/com/google/accompanist/web/WebView.kt
```diff
@@ -0,0 +1,164 @@
...
+        internal set
+}
+
+/**
+ * Creates a WebView state that is remembered across Compositions.
+ *
+ * @param uri The uri to load in the WebView
+ */
+@Composable
+fun rememberWebViewState(uri: Uri) =
```
nickbutcher, line 154: Wondering about providing a function which accepts a Uri as a `String`? WDYT?

web/src/main/java/com/google/accompanist/web/WebView.kt
```diff
@@ -0,0 +1,164 @@
...
+    remember { WebViewState(WebContent.Url(uri)) }
```
nickbutcher, line 155:
```suggestion
    remember(uri) { WebViewState(WebContent.Url(uri)) }
```

web/src/main/java/com/google/accompanist/web/WebView.kt
```diff
@@ -0,0 +1,164 @@
...
+
+/**
+ * Creates a WebView state that is remembered across Compositions.
+ *
+ * @param data The uri to load in the WebView
+ */
+@Composable
+fun rememberWebViewState(data: String) =
+    remember { WebViewState(WebContent.Data(data)) }
```
nickbutcher, line 164:
```suggestion
    remember(data) { WebViewState(WebContent.Data(data)) }
```

web/src/main/java/com/google/accompanist/web/WebView.kt
```diff
@@ -0,0 +1,164 @@
...
+ * the WebView back.
+ * @param onCreated Called when the WebView is first created, this can be used to set additional
+ * settings on the WebView.
+ *
+ * @sample com.google.accompanist.sample.webview.BasicWebViewSample
+ */
+@Composable
+fun WebView(
+    state: WebViewState,
+    onContentChanged: (WebContent) -> Unit,
```
nickbutcher, line 50: Do we need this callback? Couldn't we set `state.content` directly and consumers can watch this if they need to react? This would also alleviate callers from havig to do `{ state.content = it }`

sample/src/main/java/com/google/accompanist/sample/webview/BasicWebViewSample.kt
```diff
@@ -0,0 +1,85 @@
...
+import com.google.accompanist.web.WebView
+import com.google.accompanist.web.rememberWebViewState
+
+class BasicWebViewSample : ComponentActivity() {
+    @SuppressLint("SetJavaScriptEnabled")
+    override fun onCreate(savedInstanceState: Bundle?) {
+        super.onCreate(savedInstanceState)
+        setContent {
+            AccompanistSampleTheme {
+                var textFieldValue by remember { mutableStateOf("") }
```
nickbutcher, line 48: Destructure this for easier passing to the `OutlinedTextField`?
```suggestion
                val (textFieldValue, setTextFieldValue) = remember { mutableStateOf("") }
```

web/src/main/java/com/google/accompanist/web/WebView.kt
```diff
@@ -0,0 +1,164 @@
...
+fun rememberWebViewState(uri: Uri) =
+    remember { WebViewState(WebContent.Url(uri)) }
+
+/**
+ * Creates a WebView state that is remembered across Compositions.
+ *
+ * @param data The uri to load in the WebView
+ */
+@Composable
+fun rememberWebViewState(data: String) =
```
manuelvicnt, line 163: What about the optional `baseUrl`? Should that be passed in here as well?

web/src/main/java/com/google/accompanist/web/WebView.kt
manuelvicnt: +1

web/src/main/java/com/google/accompanist/web/WebView.kt
manuelvicnt, line 64: Does the WebView need to follow any specific lifecycle? or with being `CREATED` and `DESTROYED` is enough? Are there any implications if the lifecycle is `STOPPED`, for example, but the WebView isn't?

web/src/main/java/com/google/accompanist/web/WebView.kt
```diff
@@ -0,0 +1,164 @@
...
+
+                webViewClient = object : WebViewClient() {
+                    override fun onPageStarted(view: WebView?, url: String?, favicon: Bitmap?) {
+                        super.onPageStarted(view, url, favicon)
+                        state.isLoading = true
+
+                        // When a WebView is first created it sends a page started event
+                        // with about:blank. We don't want to forward this on.
+                        // It's important we still support a link to about:blank though.
+                        if (webViewLoaded || url != "about:blank") {
```
manuelvicnt, line 75: I'm not familiar with the WebViewClient callback but.. can `onPageStarted` be called when `webViewLoaded == true`? In that case, why does it notified that the content changed, but you don't change the content?

web/src/main/java/com/google/accompanist/web/WebView.kt
```diff
@@ -0,0 +1,164 @@
...
+
+                    override fun shouldOverrideUrlLoading(
+                        view: WebView?,
+                        request: WebResourceRequest?
+                    ): Boolean {
+                        // Override all url loads to make the single source of truth
+                        // of the URL the state holder Url
+                        request?.let {
+                            val content = WebContent.Url(it.url)
+                            onContentChanged(content)
```
manuelvicnt, line 94: Ditto. Why calling `onContentChanged` instead of just do `state.content = ...`?

web/src/main/java/com/google/accompanist/web/WebView.kt
```diff
@@ -0,0 +1,164 @@
...
+            is WebContent.Data -> {
+                webView.loadDataWithBaseURL(l.baseUrl, l.data, null, null, null)
+            }
+        }
+
+        state.canGoBack = webView.canGoBack()
+    }
+}
+
+sealed class WebContent {
```
manuelvicnt, line 120: Can this be expanded in the future? If so, changes would provoke breaking changes

web/src/main/java/com/google/accompanist/web/WebView.kt
```diff
@@ -0,0 +1,164 @@
...
+    /**
+     * Whether the WebView is currently loading data in it's main frame
+     */
+    var isLoading: Boolean by mutableStateOf(false)
+        internal set
+
+    /**
+     * Whether the WebView can navigate back
+     */
+    var canGoBack: Boolean by mutableStateOf(false)
```
manuelvicnt, line 144: Why is this useful as a public API? What could developers do with it?

web/src/main/java/com/google/accompanist/web/WebView.kt
bentrengrove: Because I also wanted the url to be able to be changed from compose state outside the webview and it just felt like having the callback is how other APIs handle this

web/src/main/java/com/google/accompanist/web/WebView.kt
bentrengrove: Good spotting, yes. Will fix

web/src/main/java/com/google/accompanist/web/WebView.kt
bentrengrove: I was picturing a back button that can enable or disable but without a way to go forward it probably isn't useful. I will move it to private

web/src/main/java/com/google/accompanist/web/WebView.kt
bentrengrove: Ignore that, I have updated it and it is much better. Will remove onContentChanged

web/src/main/java/com/google/accompanist/web/WebView.kt
bentrengrove: Yeah the constant casting as WebView uses String itself it annoying. I will remove Uri and just use String I think

web/src/main/java/com/google/accompanist/web/WebView.kt
bentrengrove: Shadows the parameter name in the update lambda of AndroidView then, unless you can think of a better name for that parameter?

bentrengrove: Actually, they are better swapped so will do that

web/src/main/java/com/google/accompanist/web/WebView.kt
bentrengrove: This was to get around webview updating the url itself sometimes like in the case of redirects and navigating back. I have found a better way to do it though that removes the need for the state update here.

web/src/androidTest/kotlin/com/google/accompanist/web/DummyTests.kt
bentrengrove: Used throughout the project so will need updating elsewhere but will update this one for now

web/src/main/java/com/google/accompanist/web/WebView.kt
bentrengrove: The only thing I am aware of that isn't currently handled would be the render process death (https://developer.android.com/guide/webapps/managing-webview#termination-handle).

Is there a way to test out if there are any other concerns? I guess we could check for leaks?

web/src/main/java/com/google/accompanist/web/WebView.kt
bentrengrove: I can't think of any other cases

web/src/androidTest/kotlin/com/google/accompanist/web/WebTest.kt
bentrengrove: Removed most of them, replaced with an idling resource to wait for the webview to finish loading after Compose changes it's state

web/src/main/java/com/google/accompanist/web/WebView.kt
```diff
@@ -0,0 +1,195 @@
...
+ * using the rememberWebViewState(uri) function.
+ */
+class WebViewState(webContent: WebContent) {
+    /**
+     *  The content being loaded by the WebView
+     */
+    var content by mutableStateOf<WebContent>(webContent)
+
+    /**
+     * Whether the WebView is currently loading data in it's main frame
```
JoseAlcerreca, line 173: its

web/src/main/java/com/google/accompanist/web/WebView.kt
```diff
@@ -0,0 +1,189 @@
...
+    remember(url) { WebViewState(WebContent.Url(url)) }
+
+/**
+ * Creates a WebView state that is remembered across Compositions.
+ *
+ * @param data The uri to load in the WebView
+ */
+@Composable
+fun rememberWebViewStateWithHTMLData(data: String, baseUrl: String? = null) =
+    remember(data) { WebViewState(WebContent.Data(data, baseUrl)) }
```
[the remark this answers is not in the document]
bentrengrove: Thanks Alex, good spotting!

bentrengrove:
@manuelvicnt @nickbutcher Are there any further comments on this? I believe I have fixed up all of the above.

sample/src/main/AndroidManifest.xml
```diff
@@ -349,6 +349,16 @@
             </intent-filter>
         </activity>
 
+        <activity
+            android:name=".webview.BasicWebViewSample"
+            android:label="@string/webview_title_basic"
+            android:exported="false">
```
manuelvicnt, line 355: Nit: In a recent PR, we had to update all of these to be `exported="true"`

web/src/main/java/com/google/accompanist/web/WebView.kt
```diff
@@ -0,0 +1,189 @@
...
+                        isReload: Boolean
+                    ) {
+                        super.doUpdateVisitedHistory(view, url, isReload)
+                        // WebView will often update the current url itself.
+                        // This happens in situations like redirects and navigating through
+                        // history. We capture this change and update our state holder url.
+                        // On older APIs (28 and lower), this method is called when loading
+                        // html data. We don't want to update the state in this case as that will
+                        // overwrite the html being loaded.
+                        if (url != null &&
```
manuelvicnt, line 96: Nit: formatting. `url != null &&` should be in a new line, right?

web/src/main/java/com/google/accompanist/web/WebView.kt
```diff
@@ -0,0 +1,189 @@
...
+                            state.content = content
+                        }
+                        return true
+                    }
+                }
+            }.also { webView = it }
+        },
+        modifier = modifier
+    ) { view ->
+        when (val l = state.content) {
```
manuelvicnt, line 130: Nit: what's `val l` in this context? Can we be a bit more explicit with the val name?

web/src/main/java/com/google/accompanist/web/WebView.kt
```diff
@@ -0,0 +1,189 @@
...
+    fun getCurrentUrl(): String? {
+        return (this as? Url)?.url
+    }
+}
+
+/**
+ * A state holder to hold the state for the WebView. In most cases this will be remembered
+ * using the rememberWebViewState(uri) function.
+ */
+class WebViewState(webContent: WebContent) {
```
manuelvicnt, line 160: To check with eng, but I think this class should be annotated with `@Stable`. I'm not sure if the compiler is able to infer that `content` and `isLoading` are fields using `mutableStateOf` APIs.

sample/src/main/java/com/google/accompanist/sample/webview/BasicWebViewSample.kt
```diff
@@ -0,0 +1,80 @@
...
+                        ) {
+                            Text("Go")
+                        }
+                    }
+
+                    if (state.isLoading) {
+                        LinearProgressIndicator(Modifier.fillMaxWidth())
+                    }
+
+                    WebView(
```
manuelvicnt, line 69: To make the sample more complete, would it make sense to show how to handle errors using the `onError` param? People might wonder what to do with it :)

web/src/androidTest/kotlin/com/google/accompanist/web/WebTest.kt
```diff
@@ -0,0 +1,205 @@
...
+import org.junit.After
+import org.junit.Before
+import org.junit.Rule
+import org.junit.Test
+import org.junit.runner.RunWith
+
+@RunWith(AndroidJUnit4::class)
+// Emulator image doesn't have a WebView until API 26
+@SdkSuppress(minSdkVersion = 26)
+class WebTest {
```
manuelvicnt, line 47: Is there a reliable way to include tests for checking errors that the WebView can return?

manuelvicnt (commented):
Overall, LGTM! Great job, added some nits and potential improvements :) thank you!

web/src/androidTest/kotlin/com/google/accompanist/web/WebTest.kt
bentrengrove: I actually don't know any way to properly test errors in a webview. I did include one for a failing image load to test the new error state list.

sample/src/main/java/com/google/accompanist/sample/webview/BasicWebViewSample.kt
bentrengrove: Extended the sample just to show an error icon. Any thing else you would like to see?

web/src/main/java/com/google/accompanist/web/WebView.kt
bentrengrove: Pretty sure you are right so I have added it.

bentrengrove:
*sigh* now the API 26 level emulator is complaining it doesn't have a webview. I wonder if the image changed?

web/src/androidTest/kotlin/com/google/accompanist/web/WebTest.kt
manuelvicnt: No worries then, thank you :)

sample/src/main/java/com/google/accompanist/sample/webview/BasicWebViewSample.kt
manuelvicnt: Thank you, although I thought you'd do something similar by doing something with the `onError` lambda that the `WebView` API provides?

manuelvicnt: Ah cool! Just saw that you explain it in the WebViewState API KDocs. SGTM, thanks!

manuelvicnt (approved):

PR #832, opened 2021-10-27
fixed-on tapping tab nothing is happening

deepajais95:
issue: while tapping on a particular tab nothing is happening
i have moved TabRow code from Scaffold content to Scaffold topBar now its working


bentrengrove:
Sorry for the slow response on this. I wasn't able to reproduce this issue so I will close this as it seems to have been fixed in the mean time.

PR #835, opened 2021-10-28
Fix the edge of Ripple in BottomNavigation

Omico:
Let Ripple pass through the navigation bar in `Navigation Gestures (left and right)` mode.
This will be consistent with the effect of the status bar.

### DIFF
--- a/insets-ui/src/main/java/com/google/accompanist/insets/ui/BottomNavigation.kt
+++ b/insets-ui/src/main/java/com/google/accompanist/insets/ui/BottomNavigation.kt
@@ -63,8 +63,8 @@ fun BottomNavigation(
         Row(
             modifier = Modifier
                 .fillMaxWidth()
-                .height(BottomNavigationHeight + contentPadding.calculateBottomPadding())
                 .padding(contentPadding)
+                .height(BottomNavigationHeight)
                 .selectableGroup(),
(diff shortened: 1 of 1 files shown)


chrisbanes (approved):

insets-ui/src/main/java/com/google/accompanist/insets/ui/BottomNavigation.kt
```diff
@@ -47,19 +52,28 @@ fun BottomNavigation(
...
-            backgroundColor = Color.Transparent,
-            contentColor = contentColor,
-            elevation = 0.dp,
-            modifier = Modifier.padding(contentPadding),
-            content = content
+        Row(
+            modifier = Modifier
+                .fillMaxWidth()
+                .height(BottomNavigationHeight + contentPadding.calculateBottomPadding())
+                .padding(contentPadding)
```
alexvanyo, line 67: Tiny nit: If the `.padding` is applied before the `.height`, then you don't have to separately calculate the padding content:
```suggestion
                .padding(contentPadding)
                .height(BottomNavigationHeight)
```

alexvanyo (approved):
Thank you! Looks good to me with just an optional nitpick.

bentrengrove (approved):

PR #848, opened 2021-11-02
Wrap the contents of the BottomSheet composable into a Column composable

nilTheDev:
Wrapped the contents of the `BottomSheet` composable into a `Column` composable. The lack of a parent composable like `Row` or `Column` causes the contents in the `BottomSheet` function to be stacked up on top of each other.

### DIFF
--- a/sample/src/main/java/com/google/accompanist/sample/navigation/material/BottomSheetNavSample.kt
+++ b/sample/src/main/java/com/google/accompanist/sample/navigation/material/BottomSheetNavSample.kt
@@ -101,11 +101,15 @@ private fun HomeScreen(showSheet: () -> Unit, showFeed: () -> Unit) {
 
 @Composable
 private fun BottomSheet(showFeed: () -> Unit, showAnotherSheet: () -> Unit, arg: String) {
-    Text("Sheet with arg: $arg")
-    Button(onClick = showFeed) {
-        Text("Click me to navigate!")
-    }
-    Button(onClick = showAnotherSheet) {
-        Text("Click me to show another sheet!")
+    Column(
+        horizontalAlignment = Alignment.CenterHorizontally
+    ) {
+        Text("Sheet with arg: $arg")
+        Button(onClick = showFeed) {
+            Text("Click me to navigate!")
+        }
+        Button(onClick = showAnotherSheet) {
+            Text("Click me to show another sheet!")
+        }
     }
 }


nilTheDev:
@googlebot I signed it!

jossiwolf:
This should already be working by default - I'll check what's going on and get back to you :)

sample/src/main/java/com/google/accompanist/sample/navigation/material/BottomSheetNavSample.kt
```diff
@@ -101,11 +101,15 @@ private fun HomeScreen(showSheet: () -> Unit, showFeed: () -> Unit) {
 
 @Composable
 private fun BottomSheet(showFeed: () -> Unit, showAnotherSheet: () -> Unit, arg: String) {
-    Text("Sheet with arg: $arg")
-    Button(onClick = showFeed) {
-        Text("Click me to navigate!")
-    }
-    Button(onClick = showAnotherSheet) {
-        Text("Click me to show another sheet!")
```
nilTheDev, line 109: @jossiwolf These UI components aren't wrapped in a wrapper. That's why they are getting stacked on top of each other.

sample/src/main/java/com/google/accompanist/sample/navigation/material/BottomSheetNavSample.kt
```diff
@@ -101,11 +101,15 @@ private fun HomeScreen(showSheet: () -> Unit, showFeed: () -> Unit) {
...
+    Column(
+        horizontalAlignment = Alignment.CenterHorizontally
+    ) {
+        Text("Sheet with arg: $arg")
+        Button(onClick = showFeed) {
+            Text("Click me to navigate!")
+        }
+        Button(onClick = showAnotherSheet) {
+            Text("Click me to show another sheet!")
+        }
```
nilTheDev, line 113: @jossiwolf Using a `column` composable wraps the elements and lays them out in a column. Thus avoiding the issue.

image

nilTheDev:
I have added some more context above. Hopefully, it helps.

pavan-byjus (approved):

jossiwolf:
@nilTheDev Thank you for the pull request. This stems from us [wrapping the actual sheet content](https://github.com/google/accompanist/blob/main/navigation-material/src/main/java/com/google/accompanist/navigation/material/SheetContentHost.kt#L156) in a `Box` but then using the `ColumnScope` that we pass in - ideally we wouldn't need the box anyway. The example should stay the way it is, I will fix the underlying issue in another PR.

PR #873, opened 2021-11-12
Disallow null animations on AnimatedNavHost

### LINKED ISSUE #856: [navigator-animation] NullPointerException in AnimatedNavHost
In `AnimatedNavHost` there is allowed to set `enterTransition` and `exitTransition` and others as null. But then, when `finalEnter` and `finalExit` are defining, there is a strange non-null cast. But it could be null. Example: 
```
initialDestination.popExitTransition?.invoke(this, initialEntry, targetEntry) ?: popExitTransitions[(initialDestination.hierarchy.first {popExitTransitions.containsKey(it.route)}).route]?.invoke(this, initialEntry, targetEntry) as ExitTransition
```
It actually happens 4 times per each combination enter, exit, popEnter and popExit. Either there should not be possibility to define animations as null or need to do something with this cast. Personally I want to see possibility to disable screen animations when I don't want to use it. Now, it always take default animation from AnimatedNavHost even if I set null in `compose` method.

jbw0033:
The AnimatedNavHost should always provide some value for animations as
it is the top of the hierarchy for all destinations. If no animations
are desired, then EnterTransition.None or ExitTransition.None should be
passed in, not null.

Fixes #856


ianhanniballake (approved):

navigation-animation/src/main/java/com/google/accompanist/navigation/animation/AnimatedNavHost.kt
```diff
@@ -174,43 +169,27 @@ public fun AnimatedNavHost(
...
-                    ?: popEnterTransitions[
-                        (
-                            targetDestination.hierarchy.first {
-                                popEnterTransitions.containsKey(it.route)
-                            }
-                            ).route
-                    ]?.invoke(this) as EnterTransition
+                targetDestination.hierarchy.mapNotNull { destination ->
+                    popEnterTransitions[destination.route]?.invoke(this)
+                }.firstOrNull() ?: popEnterTransition.invoke(this)
```
hrach, line 174: Isn't this a 
```kotlin
 targetDestination.hierarchy.firstNotNullOfOrNull { destination ->
     popEnterTransitions[destination.route]?.invoke(this)
 } ?: popEnterTransition.invoke(this)
```
https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.collections/first-not-null-of-or-null.html

jbw0033: replaced it. Thanks for the suggestion!

PR #886, opened 2021-11-22
Fix set status bar icons on API 30

### LINKED ISSUE #881: SystemUiController.setStatusbarColor not changed icons color in status bar
## Describe the bug
SystemUiController.setStatusbarColor not changed icons color in status bar if android:windowLightStatusBar is "true" in themes.xml. If  android:windowLightStatusBar is "false" SystemUiController.setStatusbarColor works correct. Is it correct behavior?
## To Reproduce
1. Go to themes.xml and add `<item name="android:windowLightStatusBar">true</item>` to current theme which selected in AndroidManifiest.xml
2. Try to change icons color in status bar to light:
```
SideEffect {
        systemUiController.setStatusBarColor(
            color = statusBarColor,
            darkIcons = false
        )
    }
```
3. Icons in status bar still dark.
## Expected behavior
Icons color should be light

## Screenshots?
examplecolor

## Environment:
 - Device: OnePlus 9R, OnePlus 3
 - Accompanist version: v0.21.2-beta

alexvanyo:
This fixes #881 

The issue occurs with the `Activity`'s theme contains `android:windowLightStatusBars="true"`. In other words, the status bars icons are dark by default.

There was a bug in `androidx.core` in this case that prevented showing the light icons on API 30: https://issuetracker.google.com/issues/180881870.

This was marked as fixed, and released with `1.7.0` (which `accompanist` is using), but there is a follow-up issue: Depending on how the `WindowInsetsControllerCompat` is acquired, the workaround fixing the issue on API 30 may or may not be used: https://issuetracker.google.com/issues/207401542

In our case, we were using a codepath that doesn't have the workaround, meaning that we are still encountering the "fixed" issue (#881). By switching from `ViewCompat. getWindowInsetsController` to the direct constructor of `WindowInsetsControllerCompat`, we will ensure we still get the workaround in the meantime while the upstream issue is fixed.

### DIFF
--- a/systemuicontroller/src/main/java/com/google/accompanist/systemuicontroller/SystemUiController.kt
+++ b/systemuicontroller/src/main/java/com/google/accompanist/systemuicontroller/SystemUiController.kt
@@ -32,6 +32,7 @@ import androidx.compose.ui.graphics.toArgb
 import androidx.compose.ui.platform.LocalView
 import androidx.core.view.ViewCompat
 import androidx.core.view.WindowInsetsCompat
+import androidx.core.view.WindowInsetsControllerCompat
 
 /**
  * A class which provides easy-to-use utilities for updating the System UI bar
@@ -173,8 +174,10 @@ fun rememberSystemUiController(): SystemUiController {
 internal class AndroidSystemUiController(
     private val view: View
 ) : SystemUiController {
-    private val window = view.context.findWindow()
-    private val windowInsetsController = ViewCompat.getWindowInsetsController(view)!!
+    private val window = requireNotNull(view.context.findWindow()) {
+        "The Compose View must be hosted in an Activity with a Window!"
+    }
+    private val windowInsetsController = WindowInsetsControllerCompat(window, view)
 
     override fun setStatusBarColor(
         color: Color,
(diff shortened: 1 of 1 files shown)


systemuicontroller/src/main/java/com/google/accompanist/systemuicontroller/SystemUiController.kt
```diff
@@ -173,8 +174,10 @@ fun rememberSystemUiController(): SystemUiController {
 internal class AndroidSystemUiController(
     private val view: View
 ) : SystemUiController {
-    private val window = view.context.findWindow()
-    private val windowInsetsController = ViewCompat.getWindowInsetsController(view)!!
+    private val window = requireNotNull(view.context.findWindow()) {
+        "The Compose View must be hosted in an Activity with a Window!"
+    }
```
alexvanyo, line 179: I made the `requireNotNull` explicit here on the window, since we were effectively requiring that anyway with the old code (at least on API 29 and below). This simplified a bit of the logic here as well.

chrisbanes (approved):
:shipit:

manuelvicnt (approved):

PR #900, opened 2021-12-01
Fix destination animations being ignored

### LINKED ISSUE #882: Regression: NavHost ignores the user specified transition animation.
## Describe the bug
The transitions specified in the `build` block of `AnimatedNavHost` function are completely ignored.

As the document says, we could pass transition blocks like `exitTransition` or `popEnterTransition` to `AnimatedNavHost` composable function to specify the animation transitions when navigate to a target or navigate back.

However as the change of #873 , the user specified transitions are totally ignored.
Take enter transition for example, when `AnimatedNavHost` is actually built, the `finalEnter` transition is no longer looking for `targetDestination.enterTransition` value, but determined by the global `enterTransitions` map, which is empty by default and then fallback to the `enterTransition` parameter which has default value of `fadeIn` animation.

The `finalExit` transition has the same issue.

## To Reproduce
Creating a simple test app using `accompanist-navigation-animation:0.21.3-beta` would produce.
I've attached a sample app.

## Expected behavior
User specified transitions should not be ignored.

## Environment:
 - Accompanist version: accompanist-navigation-animation:0.21.3-beta

## Additional context
0.21.2-beta does not have this issue.

## Sample app
[MyApplication.zip](https://github.com/google/accompanist/files/7577994/MyApplication.zip)

jbw0033:
Adding destination custom animations to the animation maps to ensure
they are considered when selecting the final animation.

Fixes #882

### DIFF
--- a/navigation-animation/api/current.api
+++ b/navigation-animation/api/current.api
@@ -7,7 +7,7 @@ package com.google.accompanist.navigation.animation {
   }
 
   @androidx.compose.animation.ExperimentalAnimationApi @androidx.navigation.NavDestination.ClassType(Composable::class) public static final class AnimatedComposeNavigator.Destination extends androidx.navigation.NavDestination {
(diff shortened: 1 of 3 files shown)


ianhanniballake (changes requested):
With this change, we no longer need to store the transitions on the `AnimatedComposeNavigator.Destination` class, right? Please remove those fields if we no longer need them.

jbw0033:
Removed them.

ianhanniballake (approved):

PR #909, opened 2021-12-05
[All] Renaming sharding tests to more inclusive language

bentrengrove:
As per comment on #822, this PR expands that fix to the whole project

https://developers.google.com/style/word-list?hl=en#dummy-variable


alexvanyo (approved):
LGTM, verified that no uses are left anywhere in the project now.

PR #922, opened 2021-12-11
[Navigation] fix hard-coded  values on navigation example by using Enum class

kareemradwan:
[Navigation] fix hard-coded  values on navigation example by using Enum class fix #922


kareemradwan:
@JoseAlcerreca  can you please review this PR. 
thanks :)

ianhanniballake:
As per the comment on the issue, we don't want to do this.

PR #929, opened 2021-12-17
[SwipeRefresh] Clarify that SwipeRefresh child needs to be vertically scrollable

### LINKED ISSUE #823: SwipeRefresh does not work when child scrolls horizontally
## Describe the bug

When Swipe to refresh has a horizontally scrollable child (such as Row with `horizontalScroll` modifier or LazyRow), it stops working (swiping does nothing).

## To Reproduce

1. Display following Composable on the screen:

```kotlin
SwipeRefresh(state = rememberSwipeRefreshState(isRefreshing = false), onRefresh = { }) {
    Row(Modifier.horizontalScroll(rememberScrollState())) {
            Box(Modifier.height(500.dp).width(500.dp).background(Color.Red)) {}
            Box(Modifier.height(500.dp).width(500.dp).background(Color.Green)) {}
            Box(Modifier.height(500.dp).width(500.dp).background(Color.Blue)) {}
    }
}
```

2. Attempt to swipe to refresh

## Environment:
 - Device: OnePlus 7 Pro
 - Accompanist version: 0.20.0

matejdro:
Currently documentation only states that container needs to be scrollable. However, it needs to be **vertically** scrollable, which means that LazyRow does not count.

This update hopes to clarify that.

This fixes #823

### DIFF
--- a/docs/swiperefresh.md
+++ b/docs/swiperefresh.md
@@ -36,7 +36,7 @@ SwipeRefresh(
 
 The full example, including the view model implementation can be found [here](https://github.com/google/accompanist/blob/main/sample/src/main/java/com/google/accompanist/sample/swiperefresh/DocsSamples.kt).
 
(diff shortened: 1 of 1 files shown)


bentrengrove (approved):
LGTM! Thanks for the addition.

bentrengrove:
Can you please try to rebase this on to the latest? I think that might kick off the CI

PR #936, opened 2021-12-21
[type-safe-args] Type safe method generation for navigation in compose

dilrajsingh1997:
Provide a way to generate type-safe methods for navigation in jetpack compose. This PR contains an annotation processor, which will generate type-safe methods at compilation, making navigation easier. I am using [KSP](https://github.com/google/ksp) for annotation processing. This PR is in continuation of [#188693139](https://issuetracker.google.com/issues/188693139).
```
@ComposeDestination
interface TncPage {
    val tncUrl: String

    companion object
}
```
This is an example contract for a compose destination, The annotation processor will generate the following methods.
```
data class TncPageArgs (
    val tncUrl: kotlin.String, 
)
fun TncPage.Companion.parseArguments(backStackEntry: NavBackStackEntry): TncPageArgs {
    return TncPageArgs(
        tncUrl = backStackEntry.arguments?.getString("tncUrl") ?: "", 
    )
}
val TncPage.Companion.argumentList: MutableList<NamedNavArgument> 
    get() = mutableListOf(
        navArgument("tncUrl") {
            type = NavType.StringType
        },
    )
fun TncPage.Companion.getDestination(tncUrl: kotlin.String, ): String {
    return "TncPage?" + 
            "tncUrl=$tncUrl" + 
            ""
}
val TncPage.Companion.route
    get() = "TncPage?tncUrl={tncUrl}"
```

If we do not specify a companion object, then the annotation processor will generate a new class of the name `${destination-contract-class-name}Destination`. The following is an example of `TncPage` generated methods if we do not supply a companion object:
```
class TncPageDestination {
    data class TncPageArgs (
        val tncUrl: kotlin.String, 
    )
    companion object {
        fun parseArguments(backStackEntry: NavBackStackEntry): TncPageArgs {
            return TncPageArgs(
                tncUrl = backStackEntry.arguments?.getString("tncUrl") ?: "", 
            )
        }
        val argumentList: MutableList<NamedNavArgument> 
            get() = mutableListOf(
                navArgument("tncUrl") {
                    type = NavType.StringType
                },
            )
        fun getDestination(tncUrl: kotlin.String, ): String {
            return "TncPage?" + 
                    "tncUrl=$tncUrl" + 
                    ""
        }
        val route
            get() = "TncPage?tncUrl={tncUrl}"
    }
}
```
This library has support for data types that can be packed into a bundle (including ArrayList, parcelable and serializable), as well as for default values of arguments. [This](https://github.com/dilrajsingh1997/safe-compose-args/blob/main/AnnotationProcessorReleaseNotes.md) contains an example of how default arguments can be provided.


bentrengrove:
Thank you for submitting this. Unfortunately this is not something we can support at this time. I encourage you to open source this in your own repository for developers to use if they choose.

PR #815, opened 2021-10-23
PageState.targetPage didn't perform as we expected  when page offset between -0.5f and 0f.

MuffinMyHeart:
Fix the PagerState.targetPage when the currentPageOffset between -0.5f and 0f, the targetPage is currentPage + 1 which is wrong, it should be currentPage - 1.

PR #950, opened 2022-01-08
Pager Indicator Bug

guidoperre:
### Description

The pending target page for any on-going animations or scrolls by the user.
The difference between [pendingPage] and [targetPage] is that [targetPage] detect the
page change when the offset is more than 50%, on the other hand [pendingPage] detect the
page change when the offset is different than zero.
Returns the current page if a scroll or animation is not currently in progress.
    
### Description

In somes configurations, targetPage dont has the expected value. This cause some bugs in tab indicator.

| BUG | FIX |
| :---: | :---: |
bug | fix |


andkulikov:
Are you sure you tested on the latest version? I think this issue was already fixed by https://github.com/google/accompanist/pull/815

PR #951, opened 2022-01-10
[material-navigation] Backdrop Navigation

TheKingOfAtlantis:
# Description
Modified the existing implementation for ModelBottomSheetLayout to implement the same functionality for BackdropScaffold. Allowing the user to add destinations for the back layer in the nav host.
To allow someone to do the following:

``` kotlin

@Composable fun Example() {
    val backdropNavigator = rememberBackdropNavigator()
    val navController = rememberNavHostController(backdropNavigator)

    BackdropScaffold(backdropNavigator) {
        NavHost(navController) {
            composable("main") {  Text("Front Layer Content") }
            backdrop("backlayer")  {  Text("Back Layer Content") }
        }
    }
}

```

### DIFF
--- a/docs/navigation-material.md
+++ b/docs/navigation-material.md
@@ -58,6 +58,53 @@ This features composable bottom sheet destinations.
     }
     ```
 
+### Backdrop Destinations
+
+1. Create a `BackdropNavigator` and add it to the `NavController`:
+
+    ```kotlin
+    @Composable
+    fun MyApp() {
+        val backdropNavigator= rememberBackdropNavigator()
+        val navController = rememberNavController(backdropNavigator)
+    }
+    ```
+
+2. Wrap your `NavHost` in the `BackdropScaffold` composable that accepts a `BackdropScaffoldNavigator`.
+
+    ```kotlin
+    @Composable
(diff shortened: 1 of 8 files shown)


ianhanniballake:
From the [Backdrop material guidance](https://material.io/components/backdrop):

> A backdrop is composed of two surfaces: a back layer and a front layer. The back layer displays actions and context, and these control and inform the front layer's content.

This means that a backdrop is exactly the **wrong** use case for an independent destination in your navigation graph: the backdrop content is intrinsically linked to the foreground content; it isn't something you `navigate` to and there's no separate back stack of backdrop destinations that are `FloatingWindow` destinations above the front layer.

If anything I'd expect any kind of backdrop type of integration to instead provide a `backdrop` destination that has two content lambdas - one for the background layer (maybe supporting `null` to indicate that there shouldn't be any backdrop when you are on that destination) and for the foreground layer. But this doesn't seem to be the case?

jossiwolf:
@TheKingOfAtlantis Since there hasn't been any activity in a bit, I'm going to close this PR. Happy to reopen once we've decided on how to move forward.

Issue #657, opened 2021-08-17
[Idea] Add ability to fully expanded BottomSheetNavigator by default


jgavazzisp:
**Is your feature request related to a problem? Please describe.**
I'm trying to migrate some bottom sheets from "ScafoldBottomSheet" to the new Navigation Compose Material, however, I cannot find a way to start the sheet as fully expand — similar to the behaviour of ScafoldBottomSheet —, so it always starts as HalfExpanded which is not ideal for some use cases.

**Describe the solution you'd like**
The solution is to lock the expansion to Expanded and skip HalfExpanded when the size of the sheet content is over half of the screen height.

**Describe alternatives you've considered**
The class "SheetContentHost" contains a method called "ModalBottomSheetState.internalShow()" which could control if the initial expansion is the default behaviour (half expanded if the sheet is over half of the screen height) or it should lock to the full height of the content.

The "rememberBottomSheetNavigator" could receive the extra parameter controlling the expansion and pass it down to the SheetContentHost when appropriate.

Thanks

chrisbanes:
@jossiwolf

jossiwolf:
@jgavazzisp Thanks for the request! This is currently not supported by Compose Material's `ModalBottomSheetLayout` either and I would not like Navigation Material's `ModalBottomSheetLayout` to diverge from the actual implementation.

See https://issuetracker.google.com/issues/186669820 for a feature request for Compose Material's implementation; as soon as it's available there we will add it here as well.

jgavazzisp:
hi @jossiwolf thanks for your reply. I will my eyes on the 186669820 in the meantime.

miszmaniac:
Looking at material docs It looks like this should be supported. 
https://material.io/components/sheets-bottom#expanding-bottom-sheet

There are times where it's applicable to animate BottomSheet to full screen and let user swipe it down.

I thought I was doing something wrong, when I created view with `Modifier.fillMaxHeight(0.4f)` and everything was ok, but after adding 0.9f fraction only half was visible:)

tordjon:
Remove stale.

jossiwolf:
Reopening this for folks to be able to track progress - this still depends on the upstream issue (https://issuetracker.google.com/issues/186669820) being fixed.

jossiwolf:
[...]

PR #957, opened 2022-01-15
[Material-Navigation] Add skip half expanded navigator-level API

colintheshots:
This adds a skip half expanded flag to the `rememberBottomSheetNavigator()` function thereby fixing #657 for the navigator only.

### DIFF
--- a/docs/navigation-material.md
+++ b/docs/navigation-material.md
@@ -18,7 +18,7 @@ This features composable bottom sheet destinations.
     ```kotlin
     @Composable
     fun MyApp() {
-        val bottomSheetNavigator = rememberBottomSheetNaviginator()
+        val bottomSheetNavigator = rememberBottomSheetNavigator()
         val navController = rememberNavController(bottomSheetNavigator)
     }
     ```
--- a/docs/navigation-material.md
+++ b/docs/navigation-material.md
@@ -18,10 +18,13 @@ This features composable bottom sheet destinations.
     ```kotlin
     @Composable
     fun MyApp() {
-        val bottomSheetNavigator = rememberBottomSheetNavigator()
+        val bottomSheetNavigator = rememberBottomSheetNaviginator()
         val navController = rememberNavController(bottomSheetNavigator)
     }
     ```
+   
+Passing the `skipHalfExpanded=true` flag to `rememberBottomSheetNavigator()` will fully expand bottom 
+sheets by default instead of limiting them to half of screen height.
 
 2. Wrap your `NavHost` in the `ModalBottomSheetLayout` composable that accepts a `BottomSheetNavigator`.
 
--- a/navigation-material/src/androidTest/java/com/google/accompanist/navigation.material/BottomSheetNavigatorTest.kt
+++ b/navigation-material/src/androidTest/java/com/google/accompanist/navigation.material/BottomSheetNavigatorTest.kt
@@ -39,6 +39,7 @@ import androidx.compose.ui.unit.dp
 import androidx.navigation.NavHostController
 import androidx.navigation.compose.NavHost
 import androidx.navigation.compose.rememberNavController
+import androidx.navigation.plusAssign
 import androidx.navigation.testing.TestNavigatorState
 import androidx.test.ext.junit.runners.AndroidJUnit4
 import androidx.test.filters.LargeTest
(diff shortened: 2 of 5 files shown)


colintheshots:
I signed a CLA. Please rescan when you can. Thank you!

docs/navigation-material.md
```diff
@@ -18,10 +18,13 @@ This features composable bottom sheet destinations.
     ```kotlin
     @Composable
     fun MyApp() {
-        val bottomSheetNavigator = rememberBottomSheetNavigator()
+        val bottomSheetNavigator = rememberBottomSheetNaviginator()
```
jossiwolf, line 21:
```suggestion
        val bottomSheetNavigator = rememberBottomSheetNavigator()
```

navigation-material/src/androidTest/java/com/google/accompanist/navigation.material/BottomSheetNavigatorTest.kt
```diff
@@ -127,7 +126,10 @@ internal class BottomSheetNavigatorTest {
                 secondDestinationCompositions++
                 onDispose { secondDestinationCompositions = 0 }
             }
-            Box(Modifier.size(64.dp).testTag(secondDestinationContentTag))
+            Box(
```
jossiwolf, line 129: We can probably don't need to change the formatting here :)

navigation-material/src/androidTest/java/com/google/accompanist/navigation.material/HalfExpandedTests.kt
```diff
@@ -0,0 +1,102 @@
...
+                bottomSheetNavigator = rememberBottomSheetNavigator(skipHalfExpanded = skipHalfExpanded)
+
+                navController.navigatorProvider += bottomSheetNavigator
+
+                NavHost(
+                    navController = navController,
+                    startDestination = "start"
+                ) {
+                    composable("start") {
+                        ModalBottomSheetLayout(
```
jossiwolf, line 84: I hadn't considered this usage of the API before! While it works for this test because the `bottomSheet` destination is a `FloatingWindow` that is displayed above the `start` destination, this would break if we navigated to another non-floating destination because the `ModalBottomSheetLayout` would leave the composition.

Is there a reason we don't wrap the `NavHost` with the `ModalBottomSheetLayout` (see the [docs](https://google.github.io/accompanist/navigation-material/#bottom-sheet-destinations))?

navigation-material/src/androidTest/java/com/google/accompanist/navigation.material/HalfExpandedTests.kt
```diff
@@ -0,0 +1,102 @@
...
+
+    private suspend fun trySkipHalfExpanded(skipHalfExpanded: Boolean, screenPercent: Float, expectedSheetState: ModalBottomSheetValue) {
+        coroutineScope {
+            lateinit var navController: NavHostController
+            lateinit var bottomSheetNavigator: BottomSheetNavigator
+            composeTestRule.setContent {
+                navController = rememberNavController()
+                bottomSheetNavigator = rememberBottomSheetNavigator(skipHalfExpanded = skipHalfExpanded)
+
+                navController.navigatorProvider += bottomSheetNavigator
```
jossiwolf, line 77: Nit: Construct the nav controller after the navigator and pass it as a parameter to `rememberNavController` instead for consistency

jossiwolf (changes requested):
Thank you for your contribution! Looking good, just a few small comments :)

ianhanniballake:
I think it makes a lot more sense if the `skipHalfExpanded` is a property at the destination level and *not* as a global state on the Navigator - there's most certainly a case where one destination should skip the half expanded state while another should have a half expanded state.

jossiwolf:
Yep, I agree. Unfortunately, the only way to add a `skipHalfExpanded`flag to Material's current bottom sheet API was making it a constructor param of the state as it is needed when calculating which state to expand to.
Mutating the flag means that the anchors need to be re-calculated. Updating anchors already causes us a lot of pain so I don't want to risk more issues with it 😄

I'm hoping CaaV can help us, and I am also working on an Accompanist-owned bottom sheet implementation but that's all a bit out. Until then, I think it's better to serve the users who never mutate the `skipHalfExpanded` flag. What do you think?

It'd probably be a good idea to add a disclaimer about this though!

navigation-material/src/androidTest/java/com/google/accompanist/navigation.material/BottomSheetNavigatorTest.kt
colintheshots-meetup: This happened automatically from Command-option-L to clean up the imports and code style. I can change it back.

colintheshots:
I accidentally used my other Github account and signed the CLA with that one as well. Please rescan.

PR #970, opened 2022-01-25
Revert "Add navigator-level skip half expanded flag"

jossiwolf:
Reverts this change until we are able to provide a destination-level API.


ianhanniballake (approved):
I'm open to revisiting this in addition to the destination level flag, but I feel like we should add both at the same time (and hence, should revert this until such time that we can do both at the same time).

Issue #876, opened 2021-11-15
[Pager] animateScrollToPage tries to render all pages after changing to LazyRow


micHar:
## Describe the bug
This is a regression of https://github.com/google/accompanist/issues/368 after Pager has been reimplemented with LazyRow / LazyColumn.
> Pager seems to really try and scroll through all the pages and render them while doing so. In effect, when I animateScrollToPage from, say, page 1 to page 80, it looks very laggy. Old ViewPager is cheating here - it only really scrolls one page. So if I wanted it to scroll from page 1 to page 80, it would really only scroll from 79 to 80. It was fast enough to fool the eye, though - and thus looked smoother.

## Environment:
```
val composeVersion = "1.1.0-beta02"
val accompanistVersion = "0.21.2-beta"
```

jari-kuipers:
I was trying this out and we are experiencing this issue as well. The first method was definitely superior and Google Maps does it that way too. We're still stuck on 0.18.0 because of these issues.

I hope we can migrate soon...

micHar:
Is there any update here?

micHar:
Well, copying 0.18 version is very easy and after a quick test I don't see any issues, so as a quick workaround it's fine.
https://github.com/google/accompanist/tree/v0.18.0/pager/src/main/java/com/google/accompanist/pager

jari-kuipers:
Well glad that helped. Have you tried the latest version already? Did it fix anything?

micHar:
No, I'm on `1.1.0-beta02` right now, but I don't think new version was fixed, @andkulikov would let us know or cross referenced a pull request if he started working on that. I hope it's going to be properly fixed at some point.

@andkulikov is there maybe a Compose / LazyRow blocker issue on google issue tracker that we could follow or you don't know yet how to tackle this issue?

andkulikov:
There is no clear strategy what to do with this issue yet. I will explore how that was handled in ViewPager2 which is based on RecyclerView which is pretty similar

PR #971, opened 2022-01-25
[Pager] Do not scroll more than 3 items as an optimization

andkulikov:
Ported the behavior from ViewPager2
Fixes https://github.com/google/accompanist/issues/876

### DIFF
--- a/pager/src/main/java/com/google/accompanist/pager/PagerState.kt
+++ b/pager/src/main/java/com/google/accompanist/pager/PagerState.kt
@@ -39,6 +39,7 @@ import androidx.compose.runtime.saveable.rememberSaveable
 import androidx.compose.runtime.setValue
 import io.github.aakira.napier.DebugAntilog
 import io.github.aakira.napier.Napier
+import kotlin.math.abs
 import kotlin.math.absoluteValue
 import kotlin.math.roundToInt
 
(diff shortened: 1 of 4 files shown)


pager/src/main/java/com/google/accompanist/pager/PagerState.kt
```diff
@@ -196,6 +197,13 @@ class PagerState(
         try {
             animationTargetPage = page
 
+            // pre-jump to nearby item for long jumps as an optimization
+            // the same trick is done in ViewPager2
+            val oldPage = lazyListState.firstVisibleItemIndex
+            if (abs(page - oldPage) > 3) {
```
chrisbanes, line 203: What if I need to jump 15 pages? Should this jump to (target - 3)?

andkulikov: I think this code is doing exactly that. so page is the target. oldPage is the current. am I missing something?

manuelvicnt (approved):

pager/src/main/java/com/google/accompanist/pager/PagerState.kt
chrisbanes: Oh you're right, totally misread 🚢

Issue #949, opened 2022-01-05
[System UI Controller] statusbar color changing but not darkIcons


jayesh83:
I've Xiaomi POCO X2 android 11, I'm facing issue in status bar icon color. I've applied transparent and set darkIcons=true but icons remain white which lead to complete white status status bar.

**Additional context**
The same method worked in Pixel device android 10

alexvanyo:
Hi, thanks for the report! Which version of Compose are you using and seeing this issue?

jayesh83:
> Hi, thanks for the report! Which version of ~Compose~ Accompanist are you using and seeing this issue?
> 
> Edit: Whoops, wrong library name 😄
I'm using 
implementation "com.google.accompanist:accompanist-systemuicontroller:0.22.0-rc"
implementation 'com.google.accompanist:accompanist-insets:0.22.0-rc'

costular:
> > Hi, thanks for the report! Which version of ~Compose~ Accompanist are you using and seeing this issue?
> > Edit: Whoops, wrong library name 😄
> 
> I'm using the latest implementation "com.google.accompanist:accompanist-systemuicontroller:0.22.0-rc" implementation 'com.google.accompanist:accompanist-insets:0.22.0-rc'

I'm experiencing a similar issue using same version `0.22.0-rc`

```kotlin
val systemUiController = rememberSystemUiController()
            val useDarkIcons = MaterialTheme.colors.isLight

            SideEffect {
                systemUiController.setSystemBarsColor(
                    Color.Transparent,
                    darkIcons = useDarkIcons
                )
            }
```

✅  Shows like this on light theme:
![Light theme](https://i.imgur.com/AeYt3bD.png)

🚫  and like this on dark theme:
![Dark theme](https://i.imgur.com/PdOZ10S.png)

jayesh83:
> > > Hi, thanks for the report! Which version of ~Compose~ Accompanist are you using and seeing this issue?
> > > Edit: Whoops, wrong library name 😄
> > 
> > 
> > I'm using the latest implementation "com.google.accompanist:accompanist-systemuicontroller:0.22.0-rc" implementation 'com.google.accompanist:accompanist-insets:0.22.0-rc'
> 
> I'm experiencing a similar issue using same version `0.22.0-rc`
> 
> ```kotlin
> val systemUiController = rememberSystemUiController()
>             val useDarkIcons = MaterialTheme.colors.isLight
> 
>             SideEffect {
[...]

PR #977, opened 2022-01-27
[SystemUiController] Add API 30 workaround for system bar appearance

### LINKED ISSUE #881: SystemUiController.setStatusbarColor not changed icons color in status bar

### LINKED ISSUE #918: [System UI Controller] Preview is not rendering
Preview started throwing an exception after commit bb866bf3a1e96eb14a746649dcc9bb39878542af

```
java.lang.IllegalArgumentException: The Compose View must be hosted in an Activity with a Window!  at
com.google.accompanist.systemuicontroller.AndroidSystemUiController.<init>(SystemUiController.kt:177)   at
com.google.accompanist.systemuicontroller.SystemUiControllerKt.rememberSystemUiController(SystemUiController.kt:165)
```

https://github.com/google/accompanist/blob/ce11b260f2d2e5d69d515833b4437cdb2c67e82d/systemuicontroller/src/main/java/com/google/accompanist/systemuicontroller/SystemUiController.kt#L177

alexvanyo:
This should actually fix #881 fully and #949, for the case where the theme has `android:windowLightStatusBar=true` and trying to set it to `false` (via `statusBarDarkContentEnabled = false` or `darkIcons = false`).

There was also a somewhat similar issue for the navigation bars, where the scrim wouldn't be updated while toggling the navigation bar dark icons on API 30, which I believe is https://issuetracker.google.com/issues/215832843 but hasn't been reported here yet as an issue.

The fix for both of these seems to be setting the corresponding `systemUiVisibility` flags anyway, in addition to calling the normal flags. I'm planning to upstream these changes as well, so that we don't have to use these workarounds with a future version of `androidx.core`.

This fix also made the `systemUiController` optional again, so that it will work when using it in a `@Preview` (and fixes #918)

BEFORE:

https://user-images.githubusercontent.com/4217560/151279075-a90b5dc7-0d8d-44fc-ab5a-47401fca5756.mp4

AFTER:

https://user-images.githubusercontent.com/4217560/151279079-54da2140-643a-45ba-a311-736d44a1a777.mp4

### DIFF
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -54,7 +54,7 @@ coil-gif = { module = "io.coil-kt:coil-gif", version.ref = "coil" }
 coil-compose = { module = "io.coil-kt:coil-compose", version.ref = "coil" }
 
 androidx-appcompat = "androidx.appcompat:appcompat:1.4.1"
-androidx-core = "androidx.core:core-ktx:1.7.0"
+androidx-core = "androidx.core:core-ktx:1.8.0-alpha03"
 androidx-activity-compose = "androidx.activity:activity-compose:1.4.0"
 androidx-fragment = "androidx.fragment:fragment-ktx:1.4.0"
 androidx-dynamicanimation = "androidx.dynamicanimation:dynamicanimation-ktx:1.0.0-alpha03"
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -1,6 +1,6 @@
 [versions]
 compose = "1.2.0-alpha01"
-composesnapshot = "-" # a single character = no snapshot
+composesnapshot = "8139767" # a single character = no snapshot
 
 # gradlePlugin and lint need to be updated together
 gradlePlugin = "7.1.0-rc01"
@@ -54,7 +54,7 @@ coil-gif = { module = "io.coil-kt:coil-gif", version.ref = "coil" }
 coil-compose = { module = "io.coil-kt:coil-compose", version.ref = "coil" }
 
 androidx-appcompat = "androidx.appcompat:appcompat:1.4.1"
-androidx-core = "androidx.core:core-ktx:1.8.0-alpha03"
+androidx-core = "androidx.core:core-ktx:1.8.0-SNAPSHOT"
 androidx-activity-compose = "androidx.activity:activity-compose:1.4.0"
 androidx-fragment = "androidx.fragment:fragment-ktx:1.4.0"
 androidx-dynamicanimation = "androidx.dynamicanimation:dynamicanimation-ktx:1.0.0-alpha03"
--- a/sample/src/main/AndroidManifest.xml
+++ b/sample/src/main/AndroidManifest.xml
@@ -302,7 +302,8 @@
         <activity
             android:name=".systemuicontroller.SystemBarsColorSample"
             android:label="@string/system_ui_controller_title_color"
-            android:exported="true">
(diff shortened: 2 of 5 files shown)


hrach:
Yeah, I was studying if the navigation bar issue is accompanist's, but ended up creating the issue for androidx.core :) 

Thanks for workarounding it here!

gradle/libs.versions.toml
```diff
@@ -54,7 +54,7 @@ coil-gif = { module = "io.coil-kt:coil-gif", version.ref = "coil" }
 coil-compose = { module = "io.coil-kt:coil-compose", version.ref = "coil" }
 
 androidx-appcompat = "androidx.appcompat:appcompat:1.4.1"
-androidx-core = "androidx.core:core-ktx:1.7.0"
+androidx-core = "androidx.core:core-ktx:1.8.0-alpha03"
```
bentrengrove, line 57: What change did you need to pull in the alpha version?

bentrengrove (dismissed):
LGTM. Just checking we actually need core-ktx alpha. Definitely agree we need to try and upstream this rather than having a workaround here.

gradle/libs.versions.toml
alexvanyo: Ah, this wasn't strictly necessary, but `ViewCompat.getWindowInsetsController` will be the single non-deprecated way to get the `WindowInsetsControllerCompat` in the upcoming release

alexvanyo:
@bentrengrove I just merged in the upstream fix in core-ktx, so it will be available in a snapshot build soon and probably in the next alpha: https://android-review.googlesource.com/c/platform/frameworks/support/+/1962461

Do we want to merge in the workaround here anyway? (applying the workaround twice should be harmless)
Or wait for it to be released in `core`?
And if it is released in `core`, do we want to transitively bump it via `systemuicontroller` (even if it is still `1.8.0-alphaXX`.

Part of this PR should still be merged at some point, both to fix `@Preview`s and to update to the non-deprecated `WindowInsetsControllerCompat` creation method.

bentrengrove:
Nice work upstreaming! In that case I think we wait for it to be released in core and then point to that version once it's out in alpha. I don't think we should be pointing to snapshot releases.

Do you want to close this one and open a new PR just with the partial changes you mentioned still needing to go in?

alexvanyo:
Agreed, was able to verify that the snapshot will fix it with the latest commit. Will close and open up a new PR.

prime-zs:
I am using accompinist version 0.26.2-beta, recently the ui is not updating with icon color change. I am testing it in android 9

alexvanyo:
@prime-zs please create a new bug if you're running into an issue with the latest version!

PR #988, opened 2022-02-03
Compose 1.1.0 snapshot

bentrengrove:


gradle/libs.versions.toml
```diff
@@ -1,7 +1,7 @@
 [versions]
-compose = "1.1.0-rc01"
-composesnapshot = "-" # a single character = no snapshot
-composeCompiler = "1.1.0-rc02"
+compose = "1.1.0-SNAPSHOT"
+composesnapshot = "8131182" # a single character = no snapshot
+composeCompiler = "1.1.0-SNAPSHOT"
```
manuelvicnt, line 4: Nit: `snapshot` shouldn't use `composeCompiler`

manuelvicnt (approved):

gradle/libs.versions.toml
bentrengrove: Agreed, that fix is on the 1.2 branch though and snapshot is for 1.1. I will fix it for the proper 1.1 release

PR #996, opened 2022-02-07
[Flow Layouts] Max lines support + fix max cross axis size processing

silverxcoins:
1. With changes from this PR its now possible to specify maximum lines constraint. Helpful for various scenarios, ex. maximum lines of tags for each item of user content list.

With 16 squares and spacing:
Before/after(maxLines = 2)
 

2.  Fixes unexpected children positions in case when content cross axis size is greater than max cross axis size

With 30 squares and spacing:
Before/after
 

\+ tests


flowlayout/src/sharedTest/kotlin/com/google/accompanist/flowlayout/FlowLayoutTest.kt
```diff
@@ -1097,6 +1100,300 @@ class FlowTest : LayoutTest() {
...
+            }
+        }
+
+        assertTrue(positionedLatch.await(1, TimeUnit.SECONDS))
+
+        assertEquals(
+            IntSize(width = 0, height = 0),
+            flowSize.value
+        )
+        for (i in 0 until requestedNumberOfSquares) {
```
jossiwolf, line 1234: While this behavior is semantically correct, is it really what we want? I'd argue that `maxLines < 0` is an illegal argument and would expect a runtime crash.

jossiwolf (changes requested):
Thank you for the contribution, looking very nice! I'm new to FlowLayout so I have some behavioral questions :)

Can I also ask you to update the API definitions using `./gradlew metalavaGenerateSignature`, please?
It'd also be awesome to have the new API showcased in the Flow layout samples :)

flowlayout/src/main/java/com/google/accompanist/flowlayout/Flow.kt
```diff
@@ -171,28 +176,38 @@ private fun Flow(
...
-        fun startNewSequence() {
-            if (sequences.isNotEmpty()) {
-                crossAxisSpace += crossAxisSpacing.roundToPx()
-            }
+        // Return whether current sequence was successfully added (or declined because of
+        // constraints excess).
+        fun startNewSequence(): Boolean {
+            val crossAxisOffset = if (sequences.isNotEmpty()) crossAxisSpacing.roundToPx() else 0
+
+            val canAddCurrentSequence = sequences.size < maxLines &&
```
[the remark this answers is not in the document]
silverxcoins: Yeah, I thought about it 🙂
It seems to be a rare case, when we want to constraint max cross axis size, still showing some items partially.
Besides, such behavior is easy to achieve, for example, by adding `Modifier.verticallScroll(…, enabled = false)`, while current behavior can be reached only with inner Flow logic(or unpretty workarounds).
Although its not perfect, it looks better to me than additional flag or something like that.
Only my opinion ofc) What do you think?

flowlayout/src/sharedTest/kotlin/com/google/accompanist/flowlayout/FlowLayoutTest.kt
silverxcoins: Fixed, thank you

silverxcoins:
Updated
- Nonnegative maxLines requirement(+ test removed)
- API definitions
- Max lines sample

\+ Flow layouts samples are now on single screen with settings
image

dev-gvs:
It would be nice to have something like `onLayout` lambda, which will pass `didOverflow` boolean.

I think it's commonly enough to display some kind of element when there would be overflow. Something like "Show more" button, which will change `maxLines` to maximum on click.

In standart library `Text` composable has `onTextLayout` with `TextLayoutResult`, which contains `didOverflowWidth`.

bentrengrove:
Thank you so much for this contribution. Sorry this has taken so long to get back to you. We have been working behind the scenes to start upstreaming this to Compose main as can now be seen on goo.gle/compose-roadmap. As such we are no longer accepting PRs to FlowLayout.

Sorry again we couldn't get this in.

PR #1000, opened 2022-02-08
[Webview] Make native webview fill parent WebView composable

ln-12:
I noticed that my web content does not fill the `WebView` composable completely. Instead, it wraps the content size. In my opinion the native webview should fill out the composable. So I simply added the `layoutParams`. What do you think?


bentrengrove (approved):
Good find. Thanks for the contribution!

PR #1023, opened 2022-02-15
[Permissions] Fix typo

ilkeraslan:
The `state` variable does not exist and might be confusing for the newcomers, thus it might be better to use `cameraPermissionState` in `when` statement.

### DIFF
--- a/docs/permissions.md
+++ b/docs/permissions.md
@@ -36,7 +36,7 @@ private fun FeatureThatRequiresCameraPermission() {
         android.Manifest.permission.CAMERA
     )
 
-    when (state.status) {
+    when (cameraPermissionState.status) {
         // If the camera permission is granted, then show screen with the feature enabled
         PermissionStatus.Granted -> {
             Text("Camera permission Granted")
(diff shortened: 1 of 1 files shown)


alexvanyo:
Ah, thanks for the catch! It looks like there's 2 other `state` usages too that wouldn't compile. Would you be able to change those too?

docs/permissions.md
```diff
@@ -54,7 +54,6 @@ private fun FeatureThatRequiresCameraPermission() {
                     "Camera permission required for this feature to be available. " +
                         "Please grant the permission"
                 }
-
                 Text(textToShow)
                 Button(onClick = { state.launchPermissionRequest() }) {
```
manuelvicnt, line 58:
```suggestion
                Button(onClick = { cameraPermissionState.launchPermissionRequest() }) {
```

manuelvicnt: And another mention in L46. Could you change those two? Thank you

ilkeraslan: Sure thing, thanks for the suggestion!

ilkeraslan:
> Ah, thanks for the catch! It looks like there's 2 other `state` usages too that wouldn't compile. Would you be able to change those too?

Sure thing, I guess I was able to correct the remaining `state` usages. Let me know if there is anything else I can do! :smile:

manuelvicnt (approved):

PR #1025, opened 2022-02-16
[Navigation Animation] Fix Nested NavHost animations

### LINKED ISSUE #953: [Navigation] Nested NavHost content disappears during animation
**Description**
I have created an app which contains a root NavHost, if user is logged in then Scaffold with BottomBar is shown. Here I use another NavHost. Once I'm navigating in a root NavHost to another destination, content in nested NavHost disappears before animation in root NavHost finishes.

**Steps to reproduce**
1. Create NavHost with `ScreenA` and `ScreenB`
2. `ScreenA` contains some UI elements and nested NavHost with at least one screen `ScreenC`
3. Navigate to `ScreenB`
4. Nested `ScreenC` disappears whilst `ScreenA` animates to `ScreenB`

**Expected behavior**
- `ScreenC` will remain visible until animation finishes  

**Additional context**
Video:
https://user-images.githubusercontent.com/12201961/149327432-a3ebaa6a-9585-4987-b4be-a4d052367fff.mp4

Attaching screenshots taken from the video where you can see the blank space:
| ScreenA with nested ScreenC | Animation progress (ScreenC is gone) | ScreenB |
|:-:|---|---|
| 01-Start | 02-Progress  |  03-End |

jbw0033:
There is an issue with Nested NavHost not being animated when their
parents are being navigated away from. This is caused by the Lifecycle
state of the Nested NavHost being below STARTED which is too low for
animation. The visibleEntries flow has access to CREATED entries, so by
using that we can animate the Nested NavHost while its parent still
exists.

Fixes #953


ianhanniballake (approved):

bmcpt:
I tried these changes changes out and the issue is still there. #1133 has a code example that is still broken even with these changes.

PR #1029, opened 2022-02-16
[Web] Add properties for page title and icon to WebViewState

sampengilly:
Add properties to the WebViewState class to supply any received title and favicon from the loaded page to the caller.

### DIFF
--- a/web/src/androidTest/kotlin/com/google/accompanist/web/WebTest.kt
+++ b/web/src/androidTest/kotlin/com/google/accompanist/web/WebTest.kt
@@ -26,6 +26,7 @@ import androidx.compose.ui.test.junit4.createComposeRule
 import androidx.compose.ui.test.onNodeWithTag
 import androidx.test.espresso.web.assertion.WebViewAssertions.webMatches
 import androidx.test.espresso.web.model.Atoms.getCurrentUrl
+import androidx.test.espresso.web.model.Atoms.getTitle
 import androidx.test.espresso.web.sugar.Web.onWebView
 import androidx.test.espresso.web.webdriver.DriverAtoms.findElement
 import androidx.test.espresso.web.webdriver.DriverAtoms.getText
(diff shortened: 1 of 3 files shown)


bentrengrove:
Thanks for the contribution. I'm just wondering if we have to worry about clearing that state on subsequent page loads. Do these callbacks always get called on page load so they will never be stale?

What about if you have loaded a url and then via a compose state change load a data string, would the title and favicon be set back to null?

sampengilly:
Good point, I'll update to reset the title and favicon on page started.

bentrengrove:
You will also need to update the API file which can be done by running `./gradlew metalavaGenerateSignature`

bentrengrove:
Sorry to be a pain, this new state should also have tests. You could add some test html and test it that way, there is already an example in the project of that I believe. Let me know if you need more info.

sampengilly:
I've added a test for the title state fine. However the favicon state is a real challenge. Since the bitmap isn't necessarily loaded by the time `onPageFinished()` is called (and the idling resource resumes)

web/src/androidTest/kotlin/com/google/accompanist/web/WebTest.kt
```diff
@@ -204,17 +206,48 @@ class WebTest {
...
+        lateinit var state: WebViewState
+
+        rule.setContent {
+            state = rememberWebViewStateWithHTMLData(data = TEST_TITLE_DATA)
+            WebTestContent(webViewState = state, idlingResource = idleResource)
+        }
+
+        // Wait for new title
+        onWebView()
+            .check(webMatches(getTitle(), equalTo("A Test Title")))
```
bentrengrove, line 220: Use your const here

bentrengrove (commented):
Nice, thank you! Don't worry about that other test for now, I'll see what I can come up with.

Just need to fix up the one small thing and this will be good to go.

web/src/androidTest/kotlin/com/google/accompanist/web/WebTest.kt
sampengilly: Done

bentrengrove (approved):
Awesome. Thanks for the contribution!

PR #1034, opened 2022-02-19
[Insets] Setup migration for accompanist/insets

alexvanyo:


sample/src/main/java/com/google/accompanist/sample/swiperefresh/SwipeRefreshContentPaddingSample.kt
```diff
@@ -76,68 +80,59 @@ private fun Sample() {
         systemUiController.setSystemBarsColor(Color.Transparent, darkIcons = useDarkIcons)
     }
 
-    Surface {
-        Box(Modifier.fillMaxSize()) {
-            // A state instance which allows us to track the size of the top app bar
-            var topAppBarSize by remember { mutableStateOf(0) }
```
alexvanyo, line 80: The new implementation here avoids a recomposition loop caused by the `.onSizeChanged`

insets/src/main/java/com/google/accompanist/insets/Insets.kt
```diff
@@ -28,6 +30,7 @@ import androidx.compose.runtime.setValue
  * offsets which describe changes to the four edges of a rectangle.
  */
 @Stable
+@Deprecated("accompanist/insets is deprecated")
```
bentrengrove, line 33: Would it be better in cases without a replaceWith to mention something like "it has been superseded by ..." so the user at least knows where to go and look?

sample/src/main/java/com/google/accompanist/sample/insets/EdgeToEdgeLazyColumn.kt
```diff
@@ -14,6 +14,8 @@
  * limitations under the License.
  */
 
+@file:Suppress("DEPRECATION")
```
bentrengrove, line 17: Please leave a comment on the sample uses of this with something like "to be replaced in a later PR to test the migration"

sample/src/main/java/com/google/accompanist/sample/swiperefresh/SwipeRefreshContentPaddingSample.kt
```diff
@@ -56,6 +58,10 @@ class SwipeRefreshContentPaddingSample : ComponentActivity() {
     override fun onCreate(savedInstanceState: Bundle?) {
         super.onCreate(savedInstanceState)
 
+        // Turn off the decor fitting system windows, which means we need to through handling
+        // insets
+        WindowCompat.setDecorFitsSystemWindows(window, false)
```
bentrengrove, line 63: Isn't this done by the systemUIController library? Or it was previously?

alexvanyo: No, system ui controller doesn't provide a way to do this.

I think in this particular sample, there was nothing doing this before? So it might have been a bit of a bug with the previous implementation.

insets/src/main/java/com/google/accompanist/insets/Insets.kt
alexvanyo: Done, I updated the deprecated descriptions to briefly explain the equivalents of methods and classes that don't have exact equivalents, and also linked to the migration doc.

sample/src/main/java/com/google/accompanist/sample/insets/EdgeToEdgeLazyColumn.kt
alexvanyo: I wasn't planning to publish those in-place to the repo, although I have a branch with those changes here working off of the migrations:
https://github.com/google/accompanist/tree/av/deprecated-insets-replace-samples

alexvanyo:
@bentrengrove and @simona-anomis this is ready for another look!

insets/src/main/java/com/google/accompanist/insets/WindowInsets.kt
```diff
@@ -509,4 +542,11 @@ annotation class ExperimentalAnimatedInsets
 /**
  * Composition local containing the current [WindowInsets].
  */
+@Deprecated(
+"""
+accompanist/insets is deprecated.
+The androidx.compose equivalent of LocalWindowInsets is the extensions on WindowInsets.
+For more migration information, please visit https://google.github.io/accompanist/insets/#migration
```
bentrengrove, line 549: Do we need to publish those docs first before merging this?

bentrengrove (commented):
This LGTM! Just the one comment, wondering if we should write the docs first before merging this and pointing people to docs that don't exist.

insets/src/main/java/com/google/accompanist/insets/WindowInsets.kt
alexvanyo: I'm not entirely sure what the doc publishing process is, will those go out immediately after publishing?

If so, the updated `docs/insets.md` in this PR should make this link real after the merge.

bentrengrove: Oh sorry! I completely missed that you had changed it because GitHub had collapsed it

docs/insets.md
```diff
@@ -2,6 +2,100 @@
 
 [![Maven Central](https://img.shields.io/maven-central/v/com.google.accompanist/accompanist-insets)](https://search.maven.org/search?q=g:com.google.accompanist)
 
+!!! warning
```
bentrengrove, line 5:
```suggestion
> :warning:
```

bentrengrove: Just thought this might look cleaner

alexvanyo: I'm not entirely familiar with the doc syntax, but plan was to getting a warning block similar to the permissions one:
https://github.com/google/accompanist/blob/d9ef702b78c0eca859fa9a1697fc13a6d0526aee/docs/permissions.md?plain=1#L7

bentrengrove: Neither am I to be honest, I was basing that on the markdown preview but I think you are right

bentrengrove (approved):

simona-anomis (approved):

Reeseb1222:
Run

Issue #1008, opened 2022-02-11
[Pager] paggerState.currentPage updates with delay (600-800 MS)


Festak:
Description:
If you have Horizontal Pager and scroll fast between the items, everything goes well and no problems, page updates immediately.
However, let's say, you scroll too slow and want to see all possible animations of custom pager indicators and etc. So you touch the item, scroll slowly to another one, see the second item, and put your finger up. The paggerState.currentPage update comes with a huge delay (it might come after 800 ms sometimes).

To reproduce:
Create the Pager,
Add logs to display the currentPage.
Scroll slowly from item 1 to item 2. 
When you can see item 2 (full item), put your finger up.
Check the logs.

Expected behavior:
The current page must be updated immediately: the same as for fast scrolling.

You can use my piece of code:
`

                val pagerState = rememberPagerState()

                val endPadding = 40.dp
                val startPadding = 40.dp
                val contentPadding = 20.dp

                HorizontalPager(
                    state = pagerState,
                    contentPadding = PaddingValues(
                        start = startPadding,
                        end = endPadding
                    ),
                    itemSpacing = contentPadding,
                    count = 6,
                    modifier = Modifier
                ) {
                    Card(
                        Modifier
                            .padding(all = 16.dp)
                            .fillMaxWidth()
                            .height(300.dp)
                    ) {

                        Text(
                            text = pagerState.currentPage.toString(),
                            textAlign = TextAlign.Center,
                            modifier = Modifier
                                .fillMaxSize()
                                .align(Alignment.CenterHorizontally)
                                .background(Color.Cyan)
                        )
                    }
                }
            }
`

andkulikov:
What version of accompanist do you use?

Festak:
@andkulikov Sorry mate.
Accompanist: 0.22.0-rc
Compose: 1.1.0-rc01

ThankRain:
My temporarily solution is taking the currentPageOffset into consideration.Before the currentPage updating,the currentPageOffset is always great than 1.0 or less than -1.0,
try to plus the currentPageOffset after rounding can get the correct currentPage

PR #1047, opened 2022-02-23
[Pager] Refactor how we update the current page

andkulikov:
We will now update it once we scrolled more than a half of the next page without waiting for the scrolling to fully settle
Fixes https://github.com/google/accompanist/issues/1008

### DIFF
--- a/pager/src/main/java/com/google/accompanist/pager/Pager.kt
+++ b/pager/src/main/java/com/google/accompanist/pager/Pager.kt
@@ -26,6 +26,7 @@ import androidx.compose.foundation.layout.Arrangement
 import androidx.compose.foundation.layout.Box
 import androidx.compose.foundation.layout.PaddingValues
 import androidx.compose.foundation.layout.calculateEndPadding
+import androidx.compose.foundation.layout.calculateStartPadding
 import androidx.compose.foundation.layout.wrapContentSize
 import androidx.compose.foundation.lazy.LazyColumn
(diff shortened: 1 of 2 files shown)


chrisbanes (approved):

bentrengrove (approved):

zoltish:
I believe this may be the reason for a regression Im seeing when using the latest release (0.24.3-alpha). 

Since the page is selected mid-swipe, if the events of that trigger a recomposition, the last bit of the swipe stutters a lot. In my case, both the pager and a bottom-bar is updated when the page changes. 

It has worked great in previous releases, so I guess Im wondering if youd consider any further tweaks to this, and if not - whether theres an alternative approach I can use?

andkulikov:
It really depends on your use case. Ideally you should try to understand why state change triggers ui junk as just postponing the page change doesn't remove the junk, it makes it less visible. However if you are sure you need the previous behavior you can do something like
```
    val state = rememberPagerState()
    var settledCurrentPage by remember { mutableStateOf(state.currentPage) }
    LaunchedEffect(state) {
        snapshotFlow {
            if (!state.isScrollInProgress) state.currentPage else settledCurrentPage
        }
            .collect {
                settledCurrentPage = it
            }
    }
```

zoltish:
Thanks @andkulikov! Youre right, from what I can tell its just **a lot** of composables being recomposed as a result of the page change, Ill still see what I can do - the new behavior is probably desirable if I can just get around the jank.

PR #1049, opened 2022-02-23
Update to Compose 1.2.0-alpha4

bentrengrove:


gradle.properties
```diff
@@ -33,7 +33,7 @@ systemProp.org.gradle.internal.http.socketTimeout=120000
 
 GROUP=com.google.accompanist
 # !! No longer need to update this manually when using a Compose SNAPSHOT
-VERSION_NAME=0.24.3-SNAPSHOT
+VERSION_NAME=0.24.3-alpha
```
JolandaVerhoef, line 36: Why does this need to be changed?

bentrengrove: This is the version number that gets sent to maven, it's always changed in these PRs and then quickly changed back to snapshot after release. That comment above it is for when dealing with making new snapshot builds.

IanGClifton (approved):

PR #1054, opened 2022-02-27
[Web] Expose progress as part of the loading state

sampengilly:
- Swap out the boolean for a `LoadingState` sealed class with a progress value
- Add test for loading state

MockWebServer was used to ensure an adequate turnaround time for the loading state to actually pass through the `Started` state. By using the same technique as other tests it never left `Finished`. I tried using HTML with meta tags to turn off any sort of caching but never saw the expected `Finished -> Started (*n) -> Finished` sequence.

Will likely conflict with https://github.com/google/accompanist/pull/1029 so I'll deal with that once things one or the other is merged.

### DIFF
--- a/web/src/main/java/com/google/accompanist/web/WebView.kt
+++ b/web/src/main/java/com/google/accompanist/web/WebView.kt
@@ -86,12 +86,18 @@ fun WebView(
                         super.onReceivedIcon(view, icon)
                         state.pageIcon = icon
                     }
+
+                    override fun onProgressChanged(view: WebView?, newProgress: Int) {
+                        super.onProgressChanged(view, newProgress)
+                        if (state.loadingState is LoadingState.Finished) return
+                        state.loadingState = LoadingState.Loading(newProgress / 100.0f)
+                    }
                 }
 
                 webViewClient = object : WebViewClient() {
                     override fun onPageStarted(view: WebView?, url: String?, favicon: Bitmap?) {
                         super.onPageStarted(view, url, favicon)
-                        state.isLoading = true
+                        state.loadingState = LoadingState.Loading(0.0f)
                         state.errorsForCurrentRequest.clear()
                         state.pageTitle = null
                         state.pageIcon = null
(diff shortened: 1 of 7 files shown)


web/src/main/java/com/google/accompanist/web/WebView.kt
```diff
@@ -180,9 +189,10 @@ class WebViewState(webContent: WebContent) {
     var content by mutableStateOf<WebContent>(webContent)
 
     /**
-     * Whether the WebView is currently loading data in its main frame
+     * Whether the WebView is currently [Started] loading data in its main frame (along with
+     * progress) or the data loading has [Finished]. See [LoadingState]
      */
-    var isLoading: Boolean by mutableStateOf(false)
```
bentrengrove, line 185: Let's keep this just as a computed property to make it easier to upgrade for people already using it. Probably something like

```
val isLoading: Boolean
   get() = loadingState !is Finished

File: web/src/main/java/com/google/accompanist/web/WebView.kt [hunk 2]
```diff
@@ -191,6 +201,23 @@ class WebViewState(webContent: WebContent) {
      * For more fine grained control use the OnError callback of the WebView.
      */
     val errorsForCurrentRequest = mutableStateListOf<WebViewError>()
+
+    /**
+     * Sealed class for constraining possible loading states.
+     * See [Started] and [Finished].
+     */
+    sealed class LoadingState
```
bentrengrove, line 209: Let's move this out of the `WebViewState` class and match the implementation of `WebContent` for consistency with the child classes being inside the sealed class. I think a better name as well might be `Loading` instead of `Started`. What do you think?

bentrengrove (changes requested):
Cool idea, just a few comments and we should be able to get this in

web/src/main/java/com/google/accompanist/web/WebView.kt
sampengilly: Yep, I had considered this, but wanted to put it up for discussion first

web/src/main/java/com/google/accompanist/web/WebView.kt
sampengilly: Yea, I was tossing up between Loading and Started. Initially it was Loading, but I changed it to better reflect where it's set from. Happy to change it back though.

sampengilly:
Hmm, looks like the test added for this might be a bit flaky across different machines.

bentrengrove:
Code looks good, I will try rerun the tests and we can see if it flakes again

bentrengrove (approved):

PR #1056, opened 2022-02-28
Remove duplicate gradle dependencies

lhoyong:
`libs.android.tools.build.gradle` is same for `libs.android.gradlePlugin`

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -21,7 +21,6 @@ buildscript {
     }
 
     dependencies {
-        classpath libs.android.tools.build.gradle
         classpath libs.android.gradlePlugin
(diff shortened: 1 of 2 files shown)


Goooler (approved):

lhoyong:
@Goooler  I resolved conflict.

bentrengrove:
This conflict resolution doesn't look right to me? We just changed that variable name to `lintMinCompose` and this would revert it

lhoyong:
> This conflict resolution doesn't look right to me? We just changed that variable name to `lintMinCompose` and this would revert it

@bentrengrove oh thanks. i fixed it

PR #1067, opened 2022-03-06
[Web] Navigation Controls

sampengilly:
Currently this is just an early proof of concept, implementing the ability to control web view navigation from outside the composable (e.g. for wiring up a `TopAppBar` "up" button to the WebView's `goBack()` method).

I figured it would be good to get some general feedback on this implementation in an early stage before going ahead with docs, tests, etc.

I figured that this stuff made more sense as its own class outside of `WebViewState` as it's a case that not everyone will have to override, and because it's more "action" oriented than "state" oriented. This implementation also seems a bit more along the lines of similar constructs elsewhere in Compose like `InteractionSource`

### DIFF
--- a/web/src/main/java/com/google/accompanist/web/WebView.kt
+++ b/web/src/main/java/com/google/accompanist/web/WebView.kt
@@ -26,14 +26,22 @@ import android.webkit.WebViewClient
 import androidx.activity.compose.BackHandler
 import androidx.compose.runtime.Composable
 import androidx.compose.runtime.Immutable
+import androidx.compose.runtime.LaunchedEffect
 import androidx.compose.runtime.Stable
 import androidx.compose.runtime.getValue
 import androidx.compose.runtime.mutableStateListOf
 import androidx.compose.runtime.mutableStateOf
 import androidx.compose.runtime.remember
+import androidx.compose.runtime.rememberCoroutineScope
 import androidx.compose.runtime.setValue
 import androidx.compose.ui.Modifier
 import androidx.compose.ui.viewinterop.AndroidView
+import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.flow.MutableSharedFlow
+import kotlinx.coroutines.flow.collect
+import kotlinx.coroutines.launch
+import kotlinx.coroutines.withContext
 
 /**
  * A wrapper around the Android View WebView to provide a basic WebView composable.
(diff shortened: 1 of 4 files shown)


web/src/main/java/com/google/accompanist/web/WebView.kt
```diff
@@ -220,6 +230,41 @@ class WebViewState(webContent: WebContent) {
     val errorsForCurrentRequest = mutableStateListOf<WebViewError>()
 }
 
+class WebViewNavigator(private val coroutineScope: CoroutineScope) {
+
+    private enum class NavigationEvent { BACK, FORWARD }
+    private val navigationEvents: MutableSharedFlow<NavigationEvent> = MutableSharedFlow()
```
bentrengrove, line 236: I don't think I am fully understanding the need for a shared flow. Are you worried about calls from multiple threads?

bentrengrove:
Cool idea, definitely something we can get in

web/src/main/java/com/google/accompanist/web/WebView.kt
sampengilly: It was mainly used as it's the simplest way to obtain a mutable flow instance that has the behavioural properties I needed (it just has an additional property, the "shared" part, that's unused). It could be done with Channels instead but it just seemed simpler. I can look into making that change.

bentrengrove: Sorry, I meant why not just call the webview functions directly, why pipe through a flow/channel?

sampengilly: I wasn't really sure that exposing the `WebView` itself to the navigator was the right thing to do, another option could be providing the navigator with references to the two methods on the `WebView`, but that also seemed like it would be a bit dangerous? The flow was something I went with in order to send instructions to the Composable in a decoupled way, without leaking references out from the Composable. I sort of looked at this as following the same sort of model as `MutableInteractionSource`.

bentrengrove: Cool, I understand it now. That decoupling does sound good. Let me go read how interaction source works so I can give you better feedback

sampengilly: Awesome, I'll work on cleaning it up, adding tests and docs. I'll probably also look at extending it slightly to include the "navigate back/forward by X steps" functions on WebView

sampengilly:
Pushed updated commit with cleaned up code, docs, tests, and rebased onto the latest `main`. The tests have a bit of overlap (e.g. forward using back to put the webview in a state where it can go forward), but I don't think there's an easy way around that.

sampengilly:
This time with spotless violations resolved 😅

bentrengrove (approved):
Great work!

PR #1086, opened 2022-03-15
[Pager Indicators] Add option to increase the size of the active page indicator

paul-jones-xero:
Adds an option to both the horizontal and vertical pager indicators to increase the scale of the active page indicator.

This is useful for accessibility enhancements to ensure that color is not the only indicator of page change (WCAG 2 SC 1.4.1)

Examples:

### DIFF
--- a/pager-indicators/api/current.api
+++ b/pager-indicators/api/current.api
@@ -2,8 +2,8 @@
 package com.google.accompanist.pager {
 
   public final class PagerIndicatorKt {
-    method @androidx.compose.runtime.Composable @com.google.accompanist.pager.ExperimentalPagerApi public static void HorizontalPagerIndicator(com.google.accompanist.pager.PagerState pagerState, optional androidx.compose.ui.Modifier modifier, optional long activeColor, optional long inactiveColor, optional float indicatorWidth, optional float indicatorHeight, optional float spacing, optional androidx.compose.ui.graphics.Shape indicatorShape);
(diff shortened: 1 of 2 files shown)


paul-jones-xero:
I wasn't sure whether or not I could merge this directly into the `compose-1.1` branch so that it can go out in a v0.23.x release? Just so we don't have to wait for Compose 1.2 to be stable. If I can, I'll just start a new PR into that branch instead :)

bentrengrove:
Unfortunately we are only rolling out new features to the Compose 1.2 branch

andkulikov:
Sorry, but unfortunately I don't think we would want to accept this contribution.
The reason is we are not yet sure in what api shape we will want to upstream page indicators in the main Compose UI once we upstream Pager.
With your API it could still be not scaleable enough as some other person could decide that just changing the scale is not enough and spacing between active and not active dots should be also configurable. So instead of trying to come up with the API which solves every possible combination you can think about our pager indicators as reference implementation, feel free to use them if they work for you, feel free to copy them in your project and tweak if not. The implementation is really simple and pretty small.
But thank you for your time!

paul-jones-xero:
No worries, can definitely see where you're coming from. We're happy to just copy and modify within our own project.

Just wondering, would you accept a PR that just hardcodes a larger size for the active indicator and applies to everyone, in order to improve accessibility/usability for every app that uses this library? That way we wouldn't have to add more options to the API.

andkulikov:
Sorry, but I don't think we should do it. This module was first released a while ago, which means there are a lot of its users. And if I remember correctly it is the first time someone asked to change the scale. It is not about accessibility. If we change it we could break the designs for all other users, maybe they internationally wanted all the dots to have the same size? We will also break the screen shot tests for them, as it is a noticeable visual change. So I still recommend you to fork the implementation. Thank you

PR #1090, opened 2022-03-17
[Insets] Update README to deprecate Insets

bentrengrove:
Moved to the bottom and marked as deprecated


alexvanyo (approved):

alexvanyo:
We may want to also mark https://github.com/google/accompanist/tree/main/insets as deprecated as well (and link to the migration docs)

bentrengrove:
Great point! Updated

alexvanyo (approved):

PR #1097, opened 2022-03-21
[Web] Add a method to expose WebViewClient and WebChromeClient

### LINKED ISSUE #1071: [Web] Feature request: add ability to prevent content loading
## Describe the feature request
As a user of this library I would like to be able to prevent the loading of a URL if matches a certain pattern.

## Additional context
It would be nice to be able to do something similar to what `WebViewClient.shouldOverrideUrlLoading()` allows to do. Currently it is only possible to monitor the `WebContent.Url` but not prevent the loading of a specific url.

### LINKED ISSUE #1095: [Web] Webview overrides developer WebViewClient
## Describe the bug

After `onCreated` is invoked, the compose function `WebView` overrides the `WebViewClient` I specified.

## To Reproduce

1. Use the compose function `WebView` to create a `WebView`.
2. In `onCreated` set a custom `WebViewClient`
3. Note that `WebViewClient` has been overridden 

## Expected behavior

`WebViewClient` set by the developer is retained.

## Proposal

Provide lambas that can be set to react to X function of `WebViewClient`, such as `onPageFinished`. 

```kotlin
fun WebView(
    state: WebViewState,
    captureBackPresses: Boolean = true,
    onCreated: (WebView) -> Unit = {},
    onError: (request: WebResourceRequest?, error: WebResourceError?) -> Unit = { _, _ -> },
    onPageFinished : (WebView, url: String?) -> Unit = { _, _ -> },
    ...
)
```

## Additional context

I understand the purpose of having a custom `WebViewClient` defined in compose, but this makes it difficult to add custom behavior to the WebView, such as reacting to when a page is finished.

bentrengrove:
Fixes #1095 
Fixes #1071 

Currently Accompanist Web sets it's own clients and doesn't provide a way to customise them. This limits the functionality and has created some obvious missing features. This is an attempt to expose the clients whilst maintaining the existing functionality. I did this via subclassing as I couldn't think of another way to do it, the clients themselves are classes not interfaces which limited the possibilities.

Open to other suggestions.

The other thing is I removed the `onError` callback as it was now duplicating functionality. I decided not to mark it as deprecated because this library was never released in a non-alpha version but I can create a deprecated version of it if needed.

### DIFF
--- a/sample/src/main/java/com/google/accompanist/sample/webview/BasicWebViewSample.kt
+++ b/sample/src/main/java/com/google/accompanist/sample/webview/BasicWebViewSample.kt
@@ -17,7 +17,10 @@
 package com.google.accompanist.sample.webview
 
 import android.annotation.SuppressLint
+import android.graphics.Bitmap
 import android.os.Bundle
+import android.util.Log
+import android.webkit.WebView
 import androidx.activity.ComponentActivity
 import androidx.activity.compose.setContent
 import androidx.compose.foundation.Image
@@ -44,6 +47,7 @@ import androidx.compose.ui.graphics.Color
 import androidx.compose.ui.graphics.ColorFilter
 import androidx.compose.ui.unit.dp
 import com.google.accompanist.sample.AccompanistSampleTheme
+import com.google.accompanist.web.AccompanistWebViewClient
 import com.google.accompanist.web.LoadingState
 import com.google.accompanist.web.WebContent
 import com.google.accompanist.web.WebView
@@ -115,13 +119,22 @@ class BasicWebViewSample : ComponentActivity() {
                         )
                     }
 
+                    // A custom WebViewClient and WebChromeClient can be provided via subclassing
+                    val webClient = object : AccompanistWebViewClient() {
+                        override fun onPageStarted(view: WebView?, url: String?, favicon: Bitmap?) {
+                            super.onPageStarted(view, url, favicon)
+                            Log.d("Accompanist WebView", "Page started loading for $url")
+                        }
+                    }
+
                     WebView(
                         state = state,
                         modifier = Modifier.weight(1f),
                         navigator = navigator,
                         onCreated = { webView ->
                             webView.settings.javaScriptEnabled = true
-                        }
+                        },
+                        client = webClient
                     )
                 }
             }
--- a/sample/src/main/java/com/google/accompanist/sample/webview/BasicWebViewSample.kt
+++ b/sample/src/main/java/com/google/accompanist/sample/webview/BasicWebViewSample.kt
@@ -120,10 +120,16 @@ class BasicWebViewSample : ComponentActivity() {
                     }
 
                     // A custom WebViewClient and WebChromeClient can be provided via subclassing
-                    val webClient = object : AccompanistWebViewClient() {
-                        override fun onPageStarted(view: WebView?, url: String?, favicon: Bitmap?) {
-                            super.onPageStarted(view, url, favicon)
-                            Log.d("Accompanist WebView", "Page started loading for $url")
+                    val webClient = remember {
+                        object : AccompanistWebViewClient() {
+                            override fun onPageStarted(
+                                view: WebView?,
+                                url: String?,
+                                favicon: Bitmap?
+                            ) {
+                                super.onPageStarted(view, url, favicon)
+                                Log.d("Accompanist WebView", "Page started loading for $url")
+                            }
                         }
                     }
 
--- a/web/src/androidTest/kotlin/com/google/accompanist/web/WebTest.kt
+++ b/web/src/androidTest/kotlin/com/google/accompanist/web/WebTest.kt
@@ -21,6 +21,7 @@ import android.webkit.WebView
 import androidx.compose.material.MaterialTheme
 import androidx.compose.runtime.Composable
 import androidx.compose.runtime.LaunchedEffect
+import androidx.compose.runtime.remember
 import androidx.compose.runtime.snapshotFlow
 import androidx.compose.ui.Modifier
 import androidx.compose.ui.platform.testTag
@@ -529,8 +530,8 @@ private fun WebTestContent(
     webViewState: WebViewState,
     idlingResource: WebViewIdlingResource,
     navigator: WebViewNavigator = rememberWebViewNavigator(),
-    client: AccompanistWebViewClient = AccompanistWebViewClient(),
-    chromeClient: AccompanistWebChromeClient = AccompanistWebChromeClient()
+    client: AccompanistWebViewClient = remember { AccompanistWebViewClient() },
+    chromeClient: AccompanistWebChromeClient = remember { AccompanistWebChromeClient() }
 ) {
     idlingResource.webviewLoading = webViewState.loadingState !is LoadingState.Finished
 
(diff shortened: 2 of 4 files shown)


sample/src/main/java/com/google/accompanist/sample/webview/BasicWebViewSample.kt
```diff
@@ -115,13 +119,22 @@ class BasicWebViewSample : ComponentActivity() {
                         )
                     }
 
+                    // A custom WebViewClient and WebChromeClient can be provided via subclassing
+                    val webClient = object : AccompanistWebViewClient() {
```
manuelvicnt, line 123: Nit: this `webClient` should be remembered

web/src/main/java/com/google/accompanist/web/WebView.kt
```diff
@@ -193,6 +122,107 @@ fun WebView(
...
 
+/**
+ * AccompanistWebViewClient
+ *
+ * A parent class implementation of WebViewClient that can be subclassed to add custom behaviour.
+ *
+ * As Accompanist Web needs to set its own web client to function, it provides this intermediary
+ * class that can be overriden if further custom behaviour is required.
+ */
+open class AccompanistWebViewClient : WebViewClient() {
```
manuelvicnt, line 133: Sad truth: I wish we could make these interfaces 😞

web/src/androidTest/kotlin/com/google/accompanist/web/WebTest.kt
```diff
@@ -467,6 +529,8 @@ private fun WebTestContent(
     webViewState: WebViewState,
     idlingResource: WebViewIdlingResource,
     navigator: WebViewNavigator = rememberWebViewNavigator(),
+    client: AccompanistWebViewClient = AccompanistWebViewClient(),
```
manuelvicnt, line 532: Please remember the default value

manuelvicnt, line 532: Please remember this client

web/src/androidTest/kotlin/com/google/accompanist/web/WebTest.kt
```diff
@@ -467,6 +529,8 @@ private fun WebTestContent(
...
+    chromeClient: AccompanistWebChromeClient = AccompanistWebChromeClient()
```
manuelvicnt, line 533: Ditto: remember

web/src/main/java/com/google/accompanist/web/WebView.kt
```diff
@@ -68,7 +68,8 @@ fun WebView(
     captureBackPresses: Boolean = true,
     navigator: WebViewNavigator = rememberWebViewNavigator(),
     onCreated: (WebView) -> Unit = {},
-    onError: (request: WebResourceRequest?, error: WebResourceError?) -> Unit = { _, _ -> }
+    client: AccompanistWebViewClient = AccompanistWebViewClient(),
```
manuelvicnt, line 71: Ditto: remember these two

web/src/androidTest/kotlin/com/google/accompanist/web/WebTest.kt
manuelvicnt, line 533: Ditto: remember

web/src/main/java/com/google/accompanist/web/WebView.kt
manuelvicnt, line 71: Ditto: remember these two

web/src/androidTest/kotlin/com/google/accompanist/web/WebTest.kt
manuelvicnt, line 532: Please remember this client

web/src/main/java/com/google/accompanist/web/WebView.kt
```diff
@@ -193,6 +122,107 @@ fun WebView(
...
+/**
+ * AccompanistWebViewClient
+ *
+ * A parent class implementation of WebViewClient that can be subclassed to add custom behaviour.
+ *
+ * As Accompanist Web needs to set its own web client to function, it provides this intermediary
+ * class that can be overriden if further custom behaviour is required.
+ */
+open class AccompanistWebViewClient : WebViewClient() {
+    internal lateinit var state: WebViewState
```
manuelvicnt, line 134: Make these two `open` so that they can be intercepted by another impl?

web/src/main/java/com/google/accompanist/web/WebView.kt
```diff
@@ -193,6 +122,107 @@ fun WebView(
...
+/**
+ * AccompanistWebChromeClient
+ *
+ * A parent class implementation of WebChromeClient that can be subclassed to add custom behaviour.
+ *
+ * As Accompanist Web needs to set its own web client to function, it provides this intermediary
+ * class that can be overriden if further custom behaviour is required.
+ */
+open class AccompanistWebChromeClient : WebChromeClient() {
+    internal lateinit var state: WebViewState
```
manuelvicnt, line 207: Ditto: consider `open` so that it can be intercepted

manuelvicnt (commented):
Some nits! Overall LGTM :) thanks!

web/src/main/java/com/google/accompanist/web/WebView.kt
bentrengrove: I really don't want people to be able to go `webViewClient.state = someOtherState` when using it so maybe
```
open lateinit var state: WebViewState
        internal set
```

sampengilly:
I think there's a question that comes up with opening this stuff up. Namely, should the compose state behaviour be guaranteed at all times, with the user only able to "hook-in" on top of that? Or should everything be able to be overridden by the user, including the default compose-like behaviours?

I also struggled a bit thinking about what the best way to expose the additional callbacks would be. The only other solution I had in mind outside of subclassing was to have something like the following:

```
data class WebChromeClientCallbacks(
    val onProgressStarted: (...) -> Unit,
    val onProgressFinished: (...) -> Unit,
    ...
)

@Composable
fun rememberWebChromeClientCallbacks(...) = remember(...) { WebChromeClientCallbacks(...) }

// Same for WebViewClient

@Composable
fun WebView(
    ...
    webChromeClientCallbacks: WebChromeClientCallbacks = rememberWebChromeClientCallbacks()
) {
   // When setting up the WebChromeClient and WebViewClient, call 
   // the appropriate callback after handling the more compose-like behaviours
}
```

That would fall into the "compose-like behaviours are guaranteed, user can hook-in for anything else on top of that" approach. Similarly if the subclassing method somehow required a `super` call.

manuelvicnt (approved):

bentrengrove:
The subclasses will be required to call super but that applies to when you use webview without Compose as well I think. I did think about doing the callbacks in a class idea as well, that's how Maps Compose did it, it just seemed like there were so many that we may as well match the implementation of webview.

I am happy with this approach for now but we can change it if we get more feature requests/bugs for things people can't do.

Doomsdayrs:
Could this not have been accomplished with a "configure webviewclient" function?

sampengilly:
Just had a thought, wouldn't this also need to set the clients in the `update` callback of `AndroidView()` to cover any changes over time?

PR #1104, opened 2022-03-23
[SystemUiController] Update androidx-core to 1.8.0-alpha06

alexvanyo:
Small version bump to `androidx.core`, which undeprecated the `WindowCompat.getInsetsController` method (which was previously deprecated).

This is a minor change on our side to use the non-deprecated version (and this won't break preview either!)

Upstream, the approach that promised to allow `ViewCompat.getWindowInsetsController` everywhere didn't work, as for some API versions direct access to the `Window` was required.

Therefore, `ViewCompat.getWindowInsetsController` was reverted to its original behavior but deprecated, since it would find the `Activity`'s `Window`. This wouldn't work when used from a `Dialog`, so `WindowCompat.getInsetsController` is the undeprecated version that requires explicitly passing the correct `Window`.

A related issue here is that `SystemUiController` wouldn't work in a `Dialog`, since it performs the similar logic of finding the `Activity`'s `Window` instead of the correct `Dialog` `Window`.

I filed https://issuetracker.google.com/issues/221889664 as a feature request for a `Window` `CompositionLocal` which could be overwritten in a `Dialog`.

### DIFF
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -54,7 +54,7 @@ coil-gif = { module = "io.coil-kt:coil-gif", version.ref = "coil" }
 coil-compose = { module = "io.coil-kt:coil-compose", version.ref = "coil" }
 
 androidx-appcompat = "androidx.appcompat:appcompat:1.4.1"
-androidx-core = "androidx.core:core-ktx:1.8.0-alpha04"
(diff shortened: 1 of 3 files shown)


bentrengrove (approved):
LGTM!

PR #1114, opened 2022-04-01
[SystemUiController] Add Dialog and custom Window support

### LINKED ISSUE #1110: [System UI Controller] Support using Window from Dialog
**Description**
Using `rememberSystemUiController` inside does not work well when in the `Dialog`.
I am trying to implement full screen dialog where dialog expands below system ui, but `SystemUiController` seems to refer the
`Window` of `Activity` below the `Dialog`.

**Steps to reproduce**
Use `rememberSystemUiController` inside any Dialog.

**Expected behavior** 
It should be able to control, for example, `statusBarDarkContentEnabled` etc.
``` kotlin
// code fragment from accompanist
 private fun Context.findWindow(): Window? {
        var context = this
        while (context is ContextWrapper) {
            if (context is Activity) return context.window
            context = context.baseContext
        }
        return null
    }
```
When calling `rememberSystemUiController()`, it lookup Window from `LocalView`. Unfortunately this only pulls Window from
Activity. [DialogLayout implements `DialogWindowProvider`](https://cs.android.com/androidx/platform/frameworks/support/+/androidx-main:compose/ui/ui/src/androidMain/kotlin/androidx/compose/ui/window/AndroidDialog.android.kt;l=203), but LocalView itself becomes `AndroidComposeView`, so in dialog you should do below to get Window.
``` kotlin
(LocalView.current.parent as DialogWindowProvider).window
```
But it would be dirty to do this and maybe quite fragile. Adding `window` parameter to `rememberSystemUiController` also seems to feasible. Support for dialog may not exist, (seems to be rare case) but at least one should be able to customize it, such as making `AndroidSystemUiController` public and adding window parameter to it.

**Additional context**
(none)

alexvanyo:
This PR adds an optional, nullable `Window` parameter to `rememberSystemUiController`.

The default value of this parameter is a multi-step fallback:

- First, the `Window` will be pulled from the `DialogWindowProvider` (if available), which will return the `Window` corresponding to a `Dialog` managed by `Compose`.
- Second, the `Window` will be attempted to be retrieved from the `Activity` (which matches current behavior).
- Third, return `null`.

If both methods fail, then no `Window` will be available, and `systemUiController` will degrade in behavior (which is useful for `@Preview` usage).

Additionally, a specific `Window` can be passed directly as an argument, which is useful in cases where the default behavior will fail (for example, if a `ComposeView` is being used inside of a `Dialog` or a `DialogFragment`)

I added a new `Dialog` sample to showcase the new behavior, and the remainder of this PR's content is tests.

This fixes #1110


bentrengrove (approved):
LGTM! Nice work

Issue #1070, opened 2022-03-09
[Navigation] isSkipHalfExpanded for bottomSheetNavigator


ThatsMeJavad:
**Description**
jetpack compose version 1.2.0-alpha01 now has ModalBottomSheetState that has an isSkipHalfExpanded flag. It can be either set through the constructor or updated later on by setting ModalBottomSheetState's isSkipHalfExpanded property to true.

I need that feature in the accompanist with bottomSheetNavigator.

does someone working on it or in the near future plan it is to be done?

PhilipDukhov:
This issue should be re-opened

PR #1119, opened 2022-04-04
[Navigation-Material] Added skipHalfExpanded to rememberBottomSheetNavigator

giacomozama:
Adds the skipHalfExpanded flag as an optional argument to rememberBottomSheetNavigator, as mentioned in [this](https://github.com/google/accompanist/issues/1070) issue.


jossiwolf:
Hi,

Thanks for opening. The API that we are aiming for will be per-destination instead of a global one - we had a contribution for the global flag that we decided to revert until we can provide the destination-level API. So we currently won't accept this contribution, sorry!

PR #1126, opened 2022-04-13
[Web] Add an initializing state to WebView

### LINKED ISSUE #1122: WebView - `loadingState` is initialized with the `LoadingState.Finished`
Considering this could be used for showing some spinner or animating WebView when the content is loaded for the first time, should you consider changing `loadingState` to be initialized with the `LoadingState.Loading` or `null`?

There is not much data I can provide to you, but my observation is visible just by looking at the initial value of `loadingState` property.

P.S. - I wanted to place this as an Open Discussion but GitHub is throwing a 404 when selecting that option.

Let me know what you think.

bentrengrove:
Fixes #1122

### DIFF
--- a/web/src/main/java/com/google/accompanist/web/WebView.kt
+++ b/web/src/main/java/com/google/accompanist/web/WebView.kt
@@ -262,7 +262,7 @@ sealed class LoadingState {
     data class Loading(val progress: Float) : LoadingState()
 
     /**
-     * Describes a webview that has finished loading content (or not started).
+     * Describes a webview that has finished loading content.
      */
     object Finished : LoadingState()
 }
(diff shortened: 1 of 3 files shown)


arriolac (approved):
LGTM

web/src/main/java/com/google/accompanist/web/WebView.kt
```diff
@@ -250,6 +250,10 @@ internal fun WebContent.withUrl(url: String) = when (this) {
  * See [Loading] and [Finished].
  */
 sealed class LoadingState {
+    /**
+     * Describes a WebView that has not yet loaded for the first time.
```
anvukelic, line 254: Don't forget to update `Finished` comment

bentrengrove: Good spotting, thank you!

PR #1141, opened 2022-04-25
[Insets] Deprecate IME nested scroll connection

alexvanyo:
The nested scroll connection is upstreamed, so we can now deprecated the accompanist version.

WIP until it is released in a non-snapshot version.


bentrengrove (approved):
LGTM

PR #1155, opened 2022-05-07
[Placeholder] Fix outline cache in placeholder

jeran:
In the original implementation of the placeholder modifier a `lastOutline` ref was used to cache the `Outline` returned by the `drawPlaceholder` method. An update was made that removed the assignment of the outline to the ref in this commit:

https://github.com/google/accompanist/commit/350f4b99fe9c04a074c9e39e25cf0eb7786ef2d7#r62458837

This change restores the ref assignment to make the outline cache function correctly.


bentrengrove (approved):
Thanks for the fix!

PR #1158, opened 2022-05-11
[pager-indicator] Support indicator of looping page with infinite page size

xiaozhikang0916:
Currently pager-indicator use `PagerState.pageCount` as the size of indicators. When I follow the guide of [HorizontalPagerLoopingSample](https://github.com/google/accompanist/blob/main/sample/src/main/java/com/google/accompanist/sample/pager/HorizontalPagerLoopingSample.kt), passing a `Int.MAX_VALUE` as page count, the indicator crashed, with no doubt.

I add a param `size` for user who needs to declare a specific page size to show indicators, and a `pageMapper` lambda to map from giving page index to a proper  index.


xiaozhikang0916:
Review needed

pager-indicators/src/main/java/com/google/accompanist/pager/PagerIndicator.kt
```diff
@@ -61,6 +66,8 @@ import androidx.compose.ui.unit.dp
 fun HorizontalPagerIndicator(
     pagerState: PagerState,
     modifier: Modifier = Modifier,
+    size: Int = pagerState.pageCount,
```
andkulikov, line 69: please rename "size" to pageCount
and "pageMapper" to "pageIndexMapping"

pager-indicators/src/main/java/com/google/accompanist/pager/PagerIndicator.kt
```diff
@@ -48,6 +50,9 @@ import androidx.compose.ui.unit.dp
  *
  * @param pagerState the state object of your [Pager] to be used to observe the list's state.
  * @param modifier the modifier to apply to this layout.
+ * @param size the size of indicators should be displayed, defaults to [PagerState.pageCount].
```
andkulikov, line 53: would be nice to also mention the use case when it is usually needed, I mean this looping scrollable pager

pager-indicators/src/main/java/com/google/accompanist/pager/PagerIndicator.kt
```diff
@@ -84,16 +91,20 @@ fun HorizontalPagerIndicator(
                 .size(width = indicatorWidth, height = indicatorHeight)
                 .background(color = inactiveColor, shape = indicatorShape)
 
-            repeat(pagerState.pageCount) {
+            repeat(size) {
                 Box(indicatorModifier)
             }
         }
 
+        val position = pageMapper(pagerState.currentPage)
```
andkulikov, line 99: please move all those calculations in the offset block. previously we only need to relayout when the currentPageOffset changes, now we also need to recompose. You can learn more about that here: https://youtu.be/EOQB8PTLkpY?t=434

andkulikov (commented):
Thank you! I left a few comments

pager-indicators/src/main/java/com/google/accompanist/pager/PagerIndicator.kt
xiaozhikang0916: Thanks, recompose is always a hard part for me to figure out.

andkulikov (approved):
Thanks

andkulikov:
The CI is complaining that there are some format violations in HorizontalPagerLoopingIndicatorSample file. Run './gradlew :sample:spotlessApply' to fix these violations.

xiaozhikang0916:
> The CI is complaining that there are some format violations in HorizontalPagerLoopingIndicatorSample file. Run './gradlew :sample:spotlessApply' to fix these violations.

This might be a bug of `spotlessApply`. It struggested to remove indents of the lambda block in `IconButton`, whose arguments were in one line.

I put the arguments in seperate lines, not following the change of `:sample:spotlessApply`, and check task passed in my local run.

PR #1165, opened 2022-05-16
[Web] Fix url in remember not causing recomposition

### LINKED ISSUE #1164: [WebView] WebViewState not updated within AccompanistWebViewClient
## Describe the bug
If a WebView composable is passed a new instance of WebViewState after the initial composition of the WebView, the AccompanistWebViewClient and AccompanistWebChromeClient instances passed in to the WebView composable do not get updated to use the new WebViewState instance. This results in the code calling the WebViewComposable using a different instance of WebViewState than what the clients are using. In my case, the LoadingState on the new instance of WebViewState is stuck as Loading and is never changed to Finished. From debugging I can see that this is because the AccompanistWebViewClient is updating the initial instance of the WebViewState instead of the new one. 

## To Reproduce
1. Instantiate a WebView with a WebViewState (created with the provided remember functions) and implementation of AccompanistWebViewClient passed in as parameters. 
2. After initial composition of the WebView, change the data for the WebViewState so that a new instance is created and passed to the WebView composable. 
3. Check the LoadingState of your new WebState and observe that it is out of sync with the one that the AccompanistWebViewClient is referencing. 

Here's a code snippet to illustrate my setup that lead to this as well:
```var webView: WebView? by remember { mutableStateOf(null) }
val webViewState = rememberWebViewStateWithHTMLData(
   data = internalHtml,
   baseUrl = webViewUrl
)

 WebView(
     state = webViewState,
     captureBackPresses = false,
     onCreated = {
                webView = it
                it.clearCache(true)
                it.clearHistory()
                it.settings.javaScriptEnabled = true
                it.addJavascriptInterface(webHandler, webHandler.interfaceName)
                Timber.d("Created WebView")
     },
     client = webViewClient,
     chromeClient = remember { MyWebChromeClient() },
)

// We change internalHtml at some point which results in the new webViewState we're referencing here no longer reflecting the LoadingState of the WebView accurately.
if(!webViewState.isLoading) {
   applyJavascript(webView)
}
```

## Expected behavior
I would expect that either the web view clients are updated with the new instance of WebViewState on each composition so the referenced states stay in sync or alternatively, the rememberWebViewState* functions could update the content field of the WebViewState instance instead of returning a new instance of WebViewState if that data changes. The latter behavior may be unexpected by consumers since remember functions typically return new instances when the passed-in data changes. 

## Environment:
 - Android OS version: Observed this on API 31, 30, and 29
 - Device: Observed on a Pixel 6, a Nexus 9 emulator, and a Pixel 5 emulator
 - Accompanist version: 0.24.7-alpha

## Additional context
I tried to include all of the same information here, but here's my conversation on this from the Kotlinlang Slack space: https://kotlinlang.slack.com/archives/CJLTWPH7S/p1652281163585679.

bentrengrove:
Fixes #1164 

The way the remember functions were written, their inputs could never be changed.

Switched to using `.apply {}` which the correct way to write these helper remember functions. See `rememberUpdatedSlingshot()` in this project for another example.

Added tests to cover this case for the future.

### DIFF
--- a/web/src/androidTest/kotlin/com/google/accompanist/web/WebTest.kt
+++ b/web/src/androidTest/kotlin/com/google/accompanist/web/WebTest.kt
@@ -21,6 +21,7 @@ import android.webkit.WebView
 import androidx.compose.material.MaterialTheme
 import androidx.compose.runtime.Composable
 import androidx.compose.runtime.LaunchedEffect
+import androidx.compose.runtime.mutableStateOf
 import androidx.compose.runtime.remember
 import androidx.compose.runtime.snapshotFlow
 import androidx.compose.ui.Modifier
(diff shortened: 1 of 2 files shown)


bentrengrove:
Just discovered this breaks redirects. Will reopen when ready

PR #1183, opened 2022-05-30
[Pager] PagerState's lazyListState public

cindy-at-fetch-rewards:
Make PagerState's lazyListState public.

I'd like to propose having the lazyListState of the PagerState to become public. A use case that I am needing the lazyListState is to check the `layoutInfo.visibleItemsInfo` of the lazyListState to check if an item is currently visible. I also saw the question as the comment, and my answer is Yes.

### DIFF
--- a/pager/src/main/java/com/google/accompanist/pager/PagerState.kt
+++ b/pager/src/main/java/com/google/accompanist/pager/PagerState.kt
@@ -74,8 +74,7 @@ fun rememberPagerState(
 class PagerState(
     @IntRange(from = 0) currentPage: Int = 0,
 ) : ScrollableState {
-    // Should this be public?
-    internal val lazyListState = LazyListState(firstVisibleItemIndex = currentPage)
+    val lazyListState = LazyListState(firstVisibleItemIndex = currentPage)
 
     private var _currentPage by mutableStateOf(currentPage)
 


andkulikov:
I think that the fact Pager works on top of LazyRow should stay an implementation detail. Making it public will expose this.
But could you please explain your use case better. Pager is designed in a way to have one main page and this page and its offset is exposed in the PagerState

cindy-at-fetch-rewards:
@andkulikov I agree with your comment.

Without having to go deep into context, I am looking to get the visible items in the ScrollableState. When the ScrollableState is used from a `LazyRow` or `LazyColumn`, I can leverage its `LazyListLayoutInfo.visibleItemsInfo`. What's nice about it is the ability to check its `key` and set a custom `key` when declaring the items in the LazyList. Now I'm looking for this alternative with the `PagerState`. Is it accurate that I am able to declare:

```
val visibleRange = IntRange((currentPage - currentPageOffset).toInt(), (currentPage + currentPageOffset).toInt())
```

to get the current visible items on screen? Then check the index of the current item to see if it's within the range to determin that it's actually visible on screen? My ViewPager, when snapped, will display partial view of the item to the left of the `currentPage` and the partial view of the item to the right of the `currentPage`. Would the same also work when the number of visible items on each side of the `currentPage` is greater than 1? For example, 2 items can be seen on either side of the `currentPage` thus making it 5 visible items on screen, so the IntRange will be a range of 5. Let me know if this makes sense. Thank you

andkulikov:
Yes, your calculation does make sense. However I do not believe Pager API allows to have 2 adjacent pages visible on side for the second example

PR #1186, opened 2022-06-01
Compose 1.2.0-beta03

bentrengrove:
Fixed an issue in navigation material. It was setting hide called back to false too soon, removing the set fixes the issue. The flag is set back to false thanks to the remember being keyed with the backStackEntry


navigation-material/src/main/java/com/google/accompanist/navigation/material/SheetContentHost.kt
```diff
@@ -141,11 +139,7 @@ internal fun SheetContentHost(
             onDispose {
                 scope.launch {
                     hideCalled = true
-                    try {
-                        sheetState.internalHide()
-                    } finally {
-                        hideCalled = false
-                    }
+                    sheetState.internalHide()
```
jossiwolf, line 142: Certainly not the ideal solution but if it works I can come back to it next week :)

jossiwolf (approved):
Thank you, Ben!

PR #1194, opened 2022-06-09
[Pager] Fix Modifier.pagerTabIndicatorOffset()

andkulikov:
After pagerState.currentPage was changed to be updated once we scrolled half of the page the indicator offsets were broken.
I also removed unneeded use of composed modifier and page scrolls causing recomposition.


bentrengrove (approved):
Is there an attached issue to close?

andkulikov:
> Is there an attached issue to close?
No, I just noticed this issue during my own testing

Issue #1178, opened 2022-05-27
[WebView] Suggestion for persistent WebView upon configuration change or navigating to/from WebView composable


jacobmichaelis:
## Describe the bug
Android Bug: When rotating the screen, or navigating to another composable and back, the WebView recomposes with no way to show an existing stored WebView. 

## To Reproduce
* Implement Accompanist WebView
* Show a web page on the mobile device
* Rotate the screen
* See the web page reload

## Expected behavior
The website shown in the WebView doesn't reload upon configuration change/navigating back to WebView.
Suggestion:
I am storing my WebView on the ViewModel, but there is no way to just show the existing cached WebView I have when recomposing the WebView by Accompanist. My suggestion is that you add an `existingWebView: WebView? = null` to the WebView composable and then change `var webView by remember { mutableStateOf<WebView?>(null) }` to be `mutableStateOf(existingWebView)`. Then in the AndroidView do `webView ?: WebView(context).apply { ...`. I have tested this in my project by copying your code and altering it and it worked as I had hoped.
This is just a suggestion, I'm sure you will have a much more brilliant way of doing what I suggested, but I hope it will be available in one way or another in the future.

## Environment:
 - Android OS version: 24+
 - Device: Pixel 4a Emulator
 - Accompanist version: 0.24.7-alpha (though I'm pretty sure this is still an issue in current versions)

bentrengrove:
Thanks for the report, this is definitely something we need to improve.

Just as a side note, I wouldn't store a view in the viewmodel, so be careful. That sounds like an almost certain memory leak on rotation as that view will be holding on to a context reference. The proper way to handle this is by saving and restoring state, there isn't a great way to do that right now unfortunately.

One thing that might help though, if you are in a fully Compose app you can pretty safely opt out of rotation by adding
```
android:configChanges="orientation|screenSize|screenLayout"
```

to your activity in your AndroidManifest. That should help this issue.

bharath1997:
Hey guys,
I wanted to avoid the reload of webview after navigation, I don't think storing it in ViewModel would be a proper solution too, but this is kind of a serious problem as we are currently developing a hybrid app similar to amazon(native and webview)completely with compose, the main content will be shown in webview, don't want to show it within bottom sheet or fullscreen dialog as it would be the main content to be shown to user after launch, can someone please recommend some other solution or a workaround for this if possible?

workspace:
[...]

PR #1221, opened 2022-06-30
[Web] Fix webview not destroyed

### LINKED ISSUE #1220: [Web] WebView not destroyed on back
## Describe the bug
The WebView is not destroyed on composable dispose. For example, if a sound is played in the webpage, the sound continue to play after going back.

## To Reproduce
1. Go to BasicWebViewSample
2. Load this url https://webbrowsertools.com/audio-test/
3. Choose an audio sample (like mono 44100Hz)
4. Press back

## Expected behavior
Everything should stop and be released

## Environment:
 - Android OS version: 12
 - Device: Pixel 3XL
 - Accompanist version: 0.24.14-SNAPSHOT

rlatapy-luna:
Fix #1220 

Give access to the webview during composable dispose run. By default, destroy the webview.

Use https://webbrowsertools.com/audio-test/ for testing (see #1220)


web/src/main/java/com/google/accompanist/web/WebView.kt
```diff
@@ -68,6 +70,7 @@ fun WebView(
     captureBackPresses: Boolean = true,
     navigator: WebViewNavigator = rememberWebViewNavigator(),
     onCreated: (WebView) -> Unit = {},
+    onDispose: (WebView) -> Unit = { it.destroy() },
```
bentrengrove, line 73: I don't mind exposing the callback but I think `it.destroy()` should be moved to onDispose and always called. Is there a valid usecase for not calling it?

bentrengrove (changes requested):
Thank you for the bug report and the PR to fix it!

Could you please also add a basic test to ensure onDispose is called?

bentrengrove:
You'll also have to run

`./gradlew metalavaGenerateSignature` to update the API file

web/src/main/java/com/google/accompanist/web/WebView.kt
rlatapy-luna: I was thinking about configuration changes. But it probably requires more work to handle it properly (cf https://github.com/google/accompanist/issues/1178).

bentrengrove:
That is a very good point about making that linked issue worse. This might actually be too big a breaking change to introduce in the rc stage. For now, let's remove the `it.destroy` from the callback and just expose the callback as an empty callback. 

Once we are back to alphas, we can investigate where is the best place to call it. Anyone on 1.2 can use the new callback to call it themselves manually if required.

rlatapy-luna:
@bentrengrove Tests are not failing on my side 🤷‍♂️

PR #1230, opened 2022-07-07
[navigation material] use result of assertThat

saied89:
result of `assertThat` is unused in `SheetContentHostTest`


bentrengrove (approved):
Good spot!

Issue #1235, opened 2022-07-08
SwipeRefresh fails to detect gestures when Webview is used with SwipeRefresh


wilinz:
```kotlin
                val webViewState = rememberWebViewState("https://github.com")
                var webView: WebView? = remember {
                    null
                }
                val swipeRefreshState = rememberSwipeRefreshState(isRefreshing = false)
                SwipeRefresh(state = swipeRefreshState, onRefresh = {
                    swipeRefreshState.isRefreshing = true
                    webView?.reload()
                    swipeRefreshState.isRefreshing = false
                }) {
                    WebView(
                        state = webViewState,
                        onCreated = { webView0 ->
                            webView = webView0
                            setWebView(webView0)
                        }
                    )
                }
```

samanaslani1992:
I have the same problem.

PR #1242, opened 2022-07-12
[Web] Add SwipeRefreshWebView

ParkWoocheol:
SwipeRefresh not detecting gestures when using WebView with SwipeRefresh
So I temporarily put WebView in SwipeRefreshLayout , as requested in https://github.com/google/accompanist/issues/1235


bentrengrove:
Thank you very much for the contribution. Unfortunately we can't introduce a new API just to workaround a bug. The proper fix is to make Accompanist SwipeToRefresh work with Accompanist WebView.

There is nothing wrong with your implementation though so I if you need it I would encourage you to use it until we do manage to fix this bug properly.

PR #1250, opened 2022-07-18
[Permissions] Add launch application detail setting

ParkWoocheol:
- If the user doesn't want to ask, they can go to the application settings screen when this dialog doesn't appear on the screen.


ParkWoocheol:
PTAL at the new changes.

I missed run  `./gradlew metalavaGenerateSignature`

manuelvicnt:
Showing the settings screen is not something that the Android permissions UX team considers a good user experience. Thanks for the PR but I'm afraid I need to close this as it's not something we'd want to provide. Thanks!

ParkWoocheol:
I understand your comment. Thanks.

PR #1275, opened 2022-08-03
[SystemUiController] Update SystemUiController to hoist dark theme

alexvanyo:
Reworks the primary docs entrance to the system ui controller module slightly.

First, it removes the reference to `MaterialTheme.colors.isLight`. This is M2 specific, and [doesn't exist](https://issuetracker.google.com/208860800) in M3 in favor of hoisting the decision directly.

Second, it replaces `SideEffect` with a `DisposableEffect` with keys. I could go either way here, but I'm less of a fan of the implicit nature in how `SideEffect` reruns. It's not super obvious that this wouldn't work when `useDarkIcons` changes:

```
val systemUiController = rememberSystemUiController()
var useDarkIcons by remember { mutableStateOf(false) }

SideEffect {
    systemUiController.setSystemBarsColor(
        color = Color.Transparent,
        darkIcons = useDarkIcons
    )
}
```

### DIFF
--- a/docs/systemuicontroller.md
+++ b/docs/systemuicontroller.md
@@ -12,9 +12,9 @@ In your layouts you can update the system bar colors like so:
 ``` kotlin
 // Remember a SystemUiController
 val systemUiController = rememberSystemUiController()
-val useDarkIcons = MaterialTheme.colors.isLight
+val useDarkIcons = !isSystemInDarkTheme()
 
-SideEffect {
+DisposableEffect(systemUiController, useDarkIcons) {
     // Update all of the system bar colors to be transparent, and use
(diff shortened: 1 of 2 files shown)


bentrengrove:
SideEffect runs on every recomposition, why wouldn't your above example work?

alexvanyo:
> SideEffect runs on every recomposition, why wouldn't your above example work?

That's true, but `SideEffect` will only recompose once.
Since `var useDarkIcons` is using a delegate, no state reads happen in the outer scope, so the `SideEffect` won't be recompose when `useDarkIcons` changes. Unless there are other state reads of it later.

bentrengrove (approved):

PR #1276, opened 2022-08-03
[SystemUiController] Remove @Stable from SystemUiController

alexvanyo:
`@Stable` has a strong contract:

```
 * When applied to a class or an interface, [Stable] indicates that the following must be true:
 *
 *   1) The result of [equals] will always return the same result for the same two instances.
 *   2) When a public property of the type changes, composition will be notified.
 *   3) All public property types are stable.
 ```
 
 The current implementation of `SystemUiController` definitely does not adhere to item 2, since properties are being backed by calls to platform APIs that aren't snapshot state aware and won't notify composition if they change.
 
 I'm not entirely sure what the ramifications are for having `@Stable` when the points above aren't satisfied, but it seems safest to remove `@Stable`?

### DIFF
--- a/systemuicontroller/api/current.api
+++ b/systemuicontroller/api/current.api
@@ -1,7 +1,7 @@
 // Signature format: 4.0
 package com.google.accompanist.systemuicontroller {
 
-  @androidx.compose.runtime.Stable public interface SystemUiController {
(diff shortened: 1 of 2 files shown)


bentrengrove:
Closing this for now, tracking in #1277

PR #1281, opened 2022-08-05
[Adaptive] Add accompanist/adaptive library with TwoPane implementation. 

alexvanyo:
Sets up an `adaptive` library, with the intention of adding a collection of adaptive layout utilities to help optimize UI for smaller and larger screens.

The first utility added is `TwoPane`, which allows placing two slots of content with their placement controlled by a specific strategy. These strategies can be composed, and the built-in entry point supports positioning the slots around folding features automatically. `TwoPane` itself does not support showing one slot or the other: if that behavior is desired (because there isn't enough room to show both slots), then that decision should be made a level above the `TwoPane`. 

Some of those more complicated layouts (list/detail, supporting panel) would then be using `TwoPane` as an implementation detail, and configuring the `TwoPane` strategy and what is passed to the slots appropriately.

This PR is using ~a snapshot~ the latest version of Compose for now to have access to `Placeable.coordinates`, since folding features are given in window coordinates, so `TwoPane` needs to know it's location relative to the window in order to know where to position its slots.

To provide information about the folding features, I've also set up ~a `WindowGeometry` object which contains information about the window. Right now, this includes the list of display features, as well as the window size class. The window size class isn't used anywhere right now, but I included it to have a single object containing the "top-level" window state that can be passed around.~ a calculation for the list of display features.

https://user-images.githubusercontent.com/4217560/182982301-0cf3321d-7819-4369-bf52-61b00e2ae2d6.mp4


adaptive/src/main/java/com/google/accompanist/adaptive/TwoPane.kt
```diff
@@ -0,0 +1,411 @@
...
+        )
+    )
+}
+
+/**
+ * Returns a [DelegateTwoPaneStrategy] that will delegate [TwoPaneStrategy.calculateSplitResult] to
+ * the [firstStrategy] when [useFirstStrategy] returns `true`. Otherwise, it will delegate to
+ * the [secondStrategy].
+ */
+public fun DelegateTwoPaneStrategy(
```
matvei-g, line 397: I assume that for the hinge usecase we will have to handle this manually, and for the rest of the usecases I didn't manage to find ones that would be worth of a separate API.

What's the usecase where this is needed?

adaptive/src/main/java/com/google/accompanist/adaptive/TwoPane.kt
```diff
@@ -0,0 +1,411 @@
...
+ */
+class SplitResult(
+
+    /**
+     * Whether the split is vertical or horizontal
+     */
+    val isHorizontalSplit: Boolean,
+
+    /**
+     * The area that is nether a `start` pane or an `end` pane, but a separation between those two.
```
matvei-g, line 175: Nice docs. We need to specify that this Rect is window-local, not layout local as well

matvei-g (commented):
Thanks Alex. Did a high level API pass, left a few comment.

Great work!

adaptive/src/main/java/com/google/accompanist/adaptive/TwoPane.kt
```diff
@@ -0,0 +1,411 @@
...
+    windowGeometry = windowGeometry
+)
+
+/**
+ * Returns a [TwoPaneStrategy] that will place the slots horizontally if there is a vertical fold.
+ *
+ * If there is no horizontal fold, then the [fallbackStrategy] will be used instead.
+ */
+public fun HorizontalTwoPaneStrategy(
+    fallbackStrategy: TwoPaneStrategy,
```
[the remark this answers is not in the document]
alexvanyo: Agreed, the "falllback" approach is nice from an implementation detail (check for folds first, then fallback to something else), but folding is not going to be the most common result of the strategy.

I think a `CompositionLocal` as mentioned above could help in this department, so that there's less of a lift required to opt-in to fold aware behavior.

adaptive/src/main/java/com/google/accompanist/adaptive/TwoPane.kt
alexvanyo: I included this as an example of how you can compose strategies that I use in the sample, but maybe this doesn't belong in the API: for more custom strategies, the implementation is fairly straightforward. I'll remove this from the API and show off the composability in the sample.

adaptive/src/main/java/com/google/accompanist/adaptive/TwoPane.kt
```diff
@@ -0,0 +1,411 @@
...
+    windowGeometry: WindowGeometry,
+): TwoPaneStrategy = TwoPaneStrategy { density, layoutDirection, layoutCoordinates ->
+    val verticalFold = windowGeometry.displayFeatures.find {
```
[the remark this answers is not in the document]
alexvanyo: Yep that's true and makes sense, I was trying to keep them together as similarly-scoped pieces of information. I think this relates to a larger `CompositionLocal` question:

If we keep them separate, then that's asking for more window-layout information to be passed around everywhere in the app through different layers. But we keep explicitness, and it's easy to replace in previews and tests.

If we instead provide them in a `CompositionLocal`, then we lose the explicitness,and previews and testing is more difficult (needing to override a `CompositionLocal` instead of passing it manually). But, we avoid having to pass them through everywhere, and we can also make it easier to have the strategies be fold-aware by default.

adaptive/src/main/java/com/google/accompanist/adaptive/TwoPane.kt
alexvanyo: Done!

adaptive/src/main/java/com/google/accompanist/adaptive/TwoPane.kt
alexvanyo: Removed!

adaptive/src/main/java/com/google/accompanist/adaptive/TwoPane.kt
```diff
@@ -0,0 +1,411 @@
...
+
+/**
+ * Returns a [TwoPaneStrategy] that will place the slots horizontally.
+ *
+ * The split will be placed at the given [splitFraction] from start, with the given [splitWidth].
+ *
+ * This strategy is _not_ fold aware.
+ */
+public fun FractionHorizontalTwoPaneStrategy(
+    splitFraction: Float,
```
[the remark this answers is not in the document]
alexvanyo: Agreed, updated "split" to always be used as a verb, and "gap" to refer to the thing created by splitting.

I also renamed to `gapBounds` instead of `gapArea`, I think that's a little bit more inline from some of the other APIs.

adaptive/src/main/java/com/google/accompanist/adaptive/TwoPane.kt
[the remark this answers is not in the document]
alexvanyo: Let me know what you think of the latest iteration. I removed the "wrapping" behavior in the implementation in favor of a list of conditional strategies that ends with one that is not conditional.

I also removed all public entry points except those that support fold-aware behavior by default for the fraction and offset entry points.

For the more general entry points, I renamed the argument from `fallbackStrategy` to `defaultStrategy`.

I think the new API does make it much easier to have foldable support by default, at the cost of being a bit stranger.

adaptive/src/main/java/com/google/accompanist/adaptive/TwoPane.kt
```diff
@@ -0,0 +1,545 @@
...
+/**
+ * Returns the specification for where to place a split in [TwoPane] as a result of
+ * [TwoPaneStrategy.calculateSplitResult]
+ */
+public class SplitResult(
+
+    /**
+     * Whether the split is vertical or horizontal
+     */
+    public val isHorizontalSplit: Boolean,
```
matvei-g, line 155: Now if we steer away from the noun `split`, we have to rephrase it. Perhaps, `splitOrientation: Orientation` might be a good idea?

adaptive/src/main/java/com/google/accompanist/adaptive/TwoPane.kt
```diff
@@ -0,0 +1,545 @@
...
+
+/**
+ * Returns a [TwoPaneStrategy] that will place the slots vertically or horizontally if there is a
+ * horizontal or vertical fold respectively.
+ *
+ * If there is no fold, then the [defaultStrategy] will be used instead.
+ */
+public fun TwoPaneStrategy(
+    windowGeometry: WindowGeometry,
+    defaultStrategy: TwoPaneStrategy,
```
matvei-g, line 217: I like the name `default` more. What I was wondering if we need this wrapping at all? Having to create a strategy and provide a default one for the non-fold seems like an unnecessary step for developers. They might (should) use the other ones, but this function take the best name in the space. 

What do you think if we have just remove this from public api?

adaptive/src/main/java/com/google/accompanist/adaptive/TwoPane.kt
```diff
@@ -0,0 +1,545 @@
...
+
+/**
+ * Returns a [TwoPaneStrategy] that will place the slots horizontally.
+ *
+ * If there is a vertical fold, then the gap will be placed along the fold.
+ *
+ * Otherwise, the gap will be placed at the given [splitFraction] from start, with the given
+ * [gapWidth].
+ */
+public fun HorizontalTwoPaneStrategy(
```
matvei-g, line 245: You can probably half the number of APIs if you provide an orientation: Orientation parameter.

adaptive/src/main/java/com/google/accompanist/adaptive/TwoPane.kt
```diff
@@ -0,0 +1,545 @@
...
+ *
+ * If there is a vertical fold, then the gap will be placed along the fold.
+ *
+ * Otherwise, the gap will be placed at [splitOffset] either from the start or end based on
+ * [offsetFromStart], with the given [gapWidth].
+ */
+public fun HorizontalTwoPaneStrategy(
+    windowGeometry: WindowGeometry,
+    splitOffset: Dp,
+    offsetFromStart: Boolean,
```
matvei-g, line 268: Seems like this might have a reasonable default, perhaps, `true`? 

Also, I cannot understand why, but having this boolean makes me want to split the functions into `FixedStartTwoPaneStrategy(orientation:Orientation, ...)` and `FixedEndTwoPaneStrategy`.

adaptive/src/main/java/com/google/accompanist/adaptive/TwoPane.kt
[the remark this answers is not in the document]
matvei-g: Regarding the split, I think splitting them we don't necessarily ask for more movements, but it might feel like because we operate with two entities instead of one. With that in mind, we might notice that we don't propagate WSC where it is not needed, so in that matter we actually have better-SRP-ed API that asks only for what it needs.

CompositionLocal, as you mentioned, adds some implicit-y and we can always fall back to this option if we want to. Do you think it would make sense to try without it? If developers want, they can do it themselves easy-peasy. Also if you want to read this composition local, you will have to deal with the @composables everywhere here in the API, which I'm not sure is a good idea for such a flexible API surface as we have right now.

Could you try to split them and we can see how it goes? Maybe then it would warrant for a composition local for the display features or the object that contains them? And we can keep it as a parameter on the two pane so that we can default to the composition local, and allow to override for previews and tests.

matvei-g (commented):
Thanks, we are getting close! A few more commentarinos

adaptive/src/main/java/com/google/accompanist/adaptive/TwoPane.kt
alexvanyo: If we remove this one from the public API, then I think we would need to expose how to create the fold strategies to allow for the same behavior, which would also mean exposing the idea of a strategy that may not apply. Not sure if we'll be able to completely avoid having a `default`.

adaptive/src/main/java/com/google/accompanist/adaptive/TwoPane.kt
alexvanyo: Updated! I went with `gapOrientation`, and used the `Orientation` that already exists in Compose (might want to have a custom one?)

adaptive/src/main/java/com/google/accompanist/adaptive/TwoPane.kt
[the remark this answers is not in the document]
alexvanyo: Updated to split up `WindowGeometry`, and depend just on the list of `DisplayFeature`s!

adaptive/src/main/java/com/google/accompanist/adaptive/TwoPane.kt
alexvanyo: Updated to have default of `true`!

adaptive/src/main/java/com/google/accompanist/adaptive/TwoPane.kt
alexvanyo: In the latest commit, I was able to keep current functionality intact while removing the direct fallback support from the public API. Let me know what you think!

adaptive/src/main/java/com/google/accompanist/adaptive/TwoPane.kt
alexvanyo: I think I'd prefer to keep it is as-is. Having the name split means the parameter names can be orientation specific (`gapWidth` vs `gapHeight`) and clearer behavior when it comes to LTR and RTL.

adaptive/src/main/java/com/google/accompanist/adaptive/TwoPane.kt
matvei-g: Nah, I think the Orientation we have covers exactly what we want

adaptive/src/main/java/com/google/accompanist/adaptive/TwoPane.kt
matvei-g: ack, let's roll with it!

adaptive/src/main/java/com/google/accompanist/adaptive/TwoPane.kt
```diff
@@ -0,0 +1,598 @@
...
+ * @param modifier an optional modifier for the layout
+ */
+@Composable
+public fun TwoPane(
+    first: @Composable () -> Unit,
+    second: @Composable () -> Unit,
+    strategy: TwoPaneStrategy,
+    displayFeatures: List<DisplayFeature>,
+    modifier: Modifier = Modifier,
+    foldAwareConfiguration: FoldAwareConfiguration = FoldAwareConfiguration.AllFolds,
```
matvei-g, line 76: Curious why this is not on the public TwoPaneStrategies? Do you think it would be a better API?

matvei-g (approved):
Thanks Alex, LGTM given the comments are addressed, I think this is a good API to try out!

I took a look at the layouting logic impl-wise, seems pretty straighforward. If you want me to take a look at some other implementation bits - let me know

adaptive/src/main/java/com/google/accompanist/adaptive/TwoPane.kt
alexvanyo: If this was on each of the `TwoPaneStrategy` s, it would be a bit more awkward to define a custom `TwoPaneStrategy` since it would have to be fold aware. Placing it on the single `TwoPane` entry point means that at least one type of fold has to be handled, and the rest of the strategies don't need to directly care about folds.

bentrengrove (approved):
Nice work!

PR #1285, opened 2022-08-10
Update to Compose 1.3-alpha03

bentrengrove:


appcompat-theme/build.gradle
```diff
@@ -25,12 +25,12 @@ kotlin {
 }
 
 android {
-    compileSdkVersion 32
+    compileSdkVersion 33
```
alexvanyo, line 28: The latest API version we test with in CI is 29, we should probably update that at some point to get closer to latest.

alexvanyo (approved):

appcompat-theme/build.gradle
bentrengrove: Great point. I will do that post upgrade

PR #1309, opened 2022-08-25
[Pager] Upgrade to Snapper 0.3.0

chrisbanes:
Removes the need to explicitly pass in the `endContentPadding` as we can rely on the necessary functionality in Lazy* (added in Compose 1.2.0).

Also removed some of the old deprecated overloads for `PagerDefaults.flingBehavior()`. This is a breaking change, but no one should be using them now anyway.

### DIFF
--- a/pager/src/main/java/com/google/accompanist/pager/Pager.kt
+++ b/pager/src/main/java/com/google/accompanist/pager/Pager.kt
@@ -44,7 +44,6 @@ import androidx.compose.ui.input.nestedscroll.nestedScroll
 import androidx.compose.ui.platform.LocalDensity
 import androidx.compose.ui.platform.LocalLayoutDirection
 import androidx.compose.ui.unit.Dp
-import androidx.compose.ui.unit.LayoutDirection
 import androidx.compose.ui.unit.Velocity
 import androidx.compose.ui.unit.dp
 import dev.chrisbanes.snapper.ExperimentalSnapperApi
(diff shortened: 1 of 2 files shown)


pager/src/main/java/com/google/accompanist/pager/Pager.kt
```diff
@@ -153,50 +100,35 @@ object PagerDefaults {
         state: PagerState,
         decayAnimationSpec: DecayAnimationSpec<Float> = rememberSplineBasedDecay(),
         snapAnimationSpec: AnimationSpec<Float> = SnapperFlingBehaviorDefaults.SpringAnimationSpec,
-        endContentPadding: Dp = 0.dp,
-        snapIndex: (SnapperLayoutInfo, startIndex: Int, targetIndex: Int) -> Int,
```
andkulikov, line 157: in the previous api we had two overloads, the one without snapIndex and the one with snapIndex with no default value. in the new impl you left only one overload where snapIndex has a default value. it is not binary compatible change unfortunately because of how default params are generated in jvm. could you please keep both of the previous variants as deprecated, and then add two similar overloads(with and without snapIndex) in the new api shape without endContentPadding. this should help us in future to nicely migrate to the new snapping implementation we are currently working on in compose foundation. thanks!

yschimke:
Is this PR defunct now that snapper is archived? Any plans to move to Compose inbuilt functionality?

chrisbanes:
I'll let @andkulikov comment on migrating to `SnapFlingBehavior`. I haven't been able to migrate in Tivi as its fling behavior doesn't quite meet my requirements, but it might work better for a pager. I'll close this PR

yschimke:
Raised https://github.com/google/accompanist/issues/1419 for tracking since this is closed.

PR #1322, opened 2022-09-05
[DrawablePainter] Fix overwriting bounds of drawable when drawing it via DrawablePainter

int02h:
This PR fixes the bug when `DrawablePainter` overwrites bounds of the drawable instance when drawing it. It leads to incorrect behaviour of content scaling of `Image`.

### DIFF
--- a/drawablepainter/build.gradle
+++ b/drawablepainter/build.gradle
@@ -66,6 +66,32 @@ android {
 dependencies {
     implementation libs.compose.ui.ui
     implementation libs.kotlin.coroutines.android
+
+    // ======================
+    // Test dependencies
+    // ======================
+
+    androidTestImplementation libs.compose.foundation.foundation
+
+    androidTestImplementation project(':internal-testutils')
+    testImplementation project(':internal-testutils')
+
+    androidTestImplementation libs.junit
+    testImplementation libs.junit
+
+    androidTestImplementation libs.truth
+    testImplementation libs.truth
+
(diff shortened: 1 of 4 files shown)


andkulikov (approved):

brunopiovan-dominos:
when the drawable is AdaptiveIconDrawable, it gets rendered too big, ignoring its bounds or size passed through a modifier, to work around this issue I convert it to a BitmapDrawable like so:

```
iconDrawable = service.loadIcon(pm).let {
    if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.O && it is AdaptiveIconDrawable) {
        BitmapDrawable(context.resources, it.toBitmap())
    } else {
        it
    }
},
```

not sure if this issue should be considered in the library

natanfudge:
After 0.26.3-beta (presumably this commit), gifs no longer expand to fit when using ContentScale.Fit.

gajicm93:
> After 0.26.3-beta (presumably this commit), some images no longer expand to fit when using ContentScale.Fit.
> Surely if I specify ContentScale.Fit, the library **should** overwrite the bounds of the drawable to be the same as the bounds of component?

I also moved from v0.25.1 now to v0.26.5-rc, and DrawablePainter now completely ignores BOTH ContentScale and Alignment set on the Image.. It only follows its own intrinsic size.

andkulikov:
The change was reverted

PR #1329, opened 2022-09-09
[Pager indicators]Don't draw active dot indicators for empty pagers.

### LINKED ISSUE #1319: [Pager] HorizontalPagerIndicator displays indicator even if pager is empty
**Description**

I'm not sure if it is a bug or an intended behavior/simplification.
If there are no pages then I would expect that HorizontalPagerIndicator displays nothing rather than a single dot.

**Steps to reproduce**
Use the following code snippet:

```kotlin
val pagerState = rememberPagerState()
    HorizontalPager(
        count = 0,
        state = pagerState,
    ) { }

    HorizontalPagerIndicator(pagerState = pagerState)
```

**Expected behavior** 

HorizontalPagerIndicator displays a single dot.

**Additional context**

Nothing is displayed.

**Environment**

        composeCompiler          version 1.3.0
        compose                   version 1.2.1
        accompanist              version 0.25.1

koral--:
Fixes #1319

Before:
Screenshot_20220909_180010

After:
Screenshot_20220909_175923

Analogous change for vertical pager indicator.

### DIFF
--- a/pager-indicators/src/main/java/com/google/accompanist/pager/PagerIndicator.kt
+++ b/pager-indicators/src/main/java/com/google/accompanist/pager/PagerIndicator.kt
@@ -113,9 +113,12 @@ fun HorizontalPagerIndicator(
                     )
                 }
                 .size(width = indicatorWidth, height = indicatorHeight)
-                .background(
-                    color = activeColor,
-                    shape = indicatorShape,
+                .then(
+                    if (pageCount > 0) Modifier.background(
+                        color = activeColor,
+                        shape = indicatorShape,
+                    )
+                    else Modifier
                 )
         )
     }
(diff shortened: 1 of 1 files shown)


pager-indicators/src/main/java/com/google/accompanist/pager/PagerIndicator.kt
```diff
@@ -113,10 +113,14 @@ fun HorizontalPagerIndicator(
                     )
                 }
                 .size(width = indicatorWidth, height = indicatorHeight)
-                .background(
-                    color = activeColor,
-                    shape = indicatorShape,
-                )
+                .apply {
```
andkulikov, line 116: Hey. It is wrong to use apply{} here unfortunately as Modifiers are build via the builder pattern, you always need to use the object returned by the next builder function and apply returns "this", not the new object. Instead you need something like
.then(if (pageCount > 0) Modifier.backround() else Modifier)

andkulikov (approved):

PR #1332, opened 2022-09-11
[Pager] Make Pager able to scroll by flinging a nested scroll child

wcaokaze:
Currently, Pager cannot scroll by flinging a nested scroll child.

https://user-images.githubusercontent.com/14194909/189514161-8326d628-2135-425c-abe5-df7bc64f7ac1.mp4

The reason is that Pager consumes ALL flings from children to prevent #347

I think that Pager should not consume flings when it is already scrolling

### DIFF
--- a/pager/src/main/java/com/google/accompanist/pager/Pager.kt
+++ b/pager/src/main/java/com/google/accompanist/pager/Pager.kt
@@ -385,6 +385,7 @@ internal fun Pager(
         ConsumeFlingNestedScrollConnection(
             consumeHorizontal = !isVertical,
             consumeVertical = isVertical,
+            pagerState = state,
         )
     }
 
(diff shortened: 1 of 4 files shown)


pager/src/main/java/com/google/accompanist/pager/Pager.kt
```diff
@@ -465,9 +468,15 @@ private class ConsumeFlingNestedScrollConnection(
     }
 
     override suspend fun onPostFling(consumed: Velocity, available: Velocity): Velocity {
-        // We can consume all post fling velocity on the main-axis
-        // so that it doesn't propagate up to the Pager
-        return available.consume(consumeHorizontal, consumeVertical)
+        return if (pagerState.currentPageLayoutInfo?.offset != 0) {
```
Levi-Moreira, line 471: Hi, do you think we could use currentPageOffset here? It's already public and if currentPageLayoutInfo.offset == 0, then this will also be 0.

wcaokaze: Ah, you are right. That's cool!
I will commit it soon.

Levi-Moreira (approved):

PR #1341, opened 2022-09-17
[Pager] Add contentType support

chibatching:
Add `contentType` parameter support for Pager library.
`contentType` is introduced in Compose 1.2.0 to improve item compositions on LazyColumn/Row.

I think it is helpful to improve pager performance if we can use this parameter.


andkulikov:
Thank you. We actually not sure if we want to add it in the Accompanist implementation. Instead we filed a feature request to add it in the new implementation of Pager we are working on in Compose UI: https://b.corp.google.com/issues/251153733

PR #1342, opened 2022-09-18
[Navigation Material] Add `skipHalfExpanded` parameter with default value equal to `false`.

jstarczewski:
Hey, 
I added `skipHalfExpanded` boolean to the parameters of `rememberBottomSheetNavigator` function, because I think it is a popular use case to make it skip half expanded state as a default behaviour, especially when it is treated as separate navigation destination.

### DIFF
--- a/navigation-material/src/main/java/com/google/accompanist/navigation/material/BottomSheetNavigator.kt
+++ b/navigation-material/src/main/java/com/google/accompanist/navigation/material/BottomSheetNavigator.kt
@@ -102,11 +102,13 @@ public class BottomSheetNavigatorSheetState(private val sheetState: ModalBottomS
 @OptIn(ExperimentalMaterialApi::class)
 @Composable
 public fun rememberBottomSheetNavigator(
-    animationSpec: AnimationSpec<Float> = SwipeableDefaults.AnimationSpec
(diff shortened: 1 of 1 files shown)


jossiwolf:
Hi, thanks for the PR. This is a change we're consciously not making as we rather want to support a destination-level flag. Due to ongoing work in Material's `ModalBottomSheetLayout` I don't believe we can make this work yet though. Feel free to prototype a destination-level flag, I might be wrong :)

The concerns are about recreating the state in time when the value changes as the destination changes, but maybe it'd work. We'd want some good test coverage that tests
- Navigating from a `skipHalfExpanded = false` to a `skipHalfExpanded = false` destination
- Navigating from a `skipHalfExpanded = false` to a `skipHalfExpanded = true` destination
- Navigating from a `skipHalfExpanded = true` to a `skipHalfExpanded = false` destination
and ensures that the sheet opens up to the correct state.

Let me know if you'd be interested in contributing this change!

jossiwolf:
Awesome, thank you!

amanv8060:
Hey, there hasn't been any progress on this since 1 year, can I go ahead and create a new pr?

jstarczewski:
@amanv8060 Yes go on please, I unfortunetley was not able to find time to adress it properly :(

ianhanniballake:

PR #1344, opened 2022-09-19
[Web] Fix webview blocking js reload

### LINKED ISSUE #1335: [WebView] window.location.reload() broken by shouldOverrideUrlLoading
AccompanistWebViewClient overrides `shouldOverrideUrlLoading` with 

```
 override fun shouldOverrideUrlLoading(
        view: WebView?,
        request: WebResourceRequest?
    ): Boolean {
        // Override all url loads to make the single source of truth
        // of the URL the state holder Url
        request?.let {
            state.content = state.content.withUrl(it.url.toString())
        }
        return true
    }
```
Returning `true` ensures that the load is always cancelled here though since it's setting `state.content` it will normally result in the URL being loaded. However 

```
                if (url.isNotEmpty() && url != view.url) {
                    view.loadUrl(url, content.additionalHttpHeaders.toMutableMap())
                }
```
in com.google.accompanist.web.WebView explicitly skips the load if the URL has not changed. This means that javascript like `window.location.reload()` will not reload the page.

## To Reproduce
I've attached [AccompanistTest.zip](https://github.com/google/accompanist/files/9576647/AccompanistTest.zip) with a sample project demonstrating the issue.

This loads a local HTML page with a single button that should reload the page with `window.location.reload()`. On page load, `********PAGE LOADED********` should be written to console and is logged to logcat. 

This only shows up on initial page load though, the button doesn't do anything.  

Uncommenting the inelegant lines 50-52 in the sample project's MainActivity fixes the issue by avoiding the AccompanistWebViewClient.shouldOverrideUrlLoading call if the URL matches the existing URL:
```
                                if (view?.url == request?.url.toString()) {
                                    return false
                                }
```

## Expected behavior
Page should reload on `window.location.reload()` and other similar calls that normally cause pages to reload.

## Environment:
 - Android OS version: Android 13
 - Device: P6Pro
 - Accompanist version: 0.25.1

bentrengrove:
Fixes #1335

### DIFF
--- a/web/src/main/java/com/google/accompanist/web/WebView.kt
+++ b/web/src/main/java/com/google/accompanist/web/WebView.kt
@@ -204,6 +204,12 @@ open class AccompanistWebViewClient : WebViewClient() {
         view: WebView?,
         request: WebResourceRequest?
     ): Boolean {
+        // If the url hasn't changed, this is probably an internal event like
+        // a javascript reload. We should let it happen.
+        if (view?.url == request?.url.toString()) {
+            return false
+        }
+
         // Override all url loads to make the single source of truth
         // of the URL the state holder Url
         request?.let {
--- a/web/src/androidTest/kotlin/com/google/accompanist/web/WebTest.kt
+++ b/web/src/androidTest/kotlin/com/google/accompanist/web/WebTest.kt
@@ -60,8 +60,8 @@ import java.util.concurrent.TimeUnit
 @RunWith(AndroidJUnit4::class)
 // Emulator image doesn't have a WebView until API 26
 // Google API emulator image seems to be really flaky before 28 so currently we will set these tests
-// to min 29
-@SdkSuppress(minSdkVersion = 28)
+// to min 29 and max 30. 31/32 image is also really flaky
+@SdkSuppress(minSdkVersion = 28, maxSdkVersion = 30)
 class WebTest {
     @get:Rule
     val rule = createComposeRule()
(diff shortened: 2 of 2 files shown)


web/src/main/java/com/google/accompanist/web/WebView.kt
```diff
@@ -204,6 +204,12 @@ open class AccompanistWebViewClient : WebViewClient() {
         view: WebView?,
         request: WebResourceRequest?
     ): Boolean {
+        // If the url hasn't changed, this is probably an internal event like
+        // a javascript reload. We should let it happen.
+        if (view?.url == request?.url.toString()) {
```
alexvanyo, line 209: If I'm understanding it correctly, this solution is to allow internal `WebView` reload requests to pass through without overriding loading.

I'm wondering if there's an alternative solution using some sort of unique ID to compare against instead of `url != view.url` to call `view.loadUrl`.

Then, instead of returning `false` here, we'd update the unique ID representing the particular request which would trigger a reload.

That might keep the number of codepaths down, and be closer to a more single-source-of-truth?
Not sure if that would change how we would want to do normal reloading.

bentrengrove: Not that I could find, you get barely any information about the request. I couldn't even tell it was from a js event in this callback.

alexvanyo: I was thinking about something like

```kotlin
    request?.let {
        state.content = state.content.withUrl(it.url.toString())
        state.requestId = UUID.randomUUID()
    }
```

And then that `requestId` could act like a `key()` to trigger a reload.

alexvanyo (approved):

PR #1352, opened 2022-09-24
[WebView] Extend the WebView composable to allow for customized (subclasses of) WebViews

morrisseyai:
This PR introduces a `factory` parameter to the WebView composable, unlocking two use cases:
1. Allows using a subclass of WebView (a very common use case in my experience)
2. Allows further control over the instantiation of the underlying WebView

I've also added a test, updated the documentation, and created a second commit which fixes a 404 linking to the documentation for WebView.

Let me know if there's anything I've missed and I can make the required changes!


bentrengrove (approved):
This looks great, thanks for the contribution!

Issue #1160, opened 2022-05-11
[Navigation Animation] Control composable route zIndex


PhilipDukhov:
I'm trying to implement modal transition: new screen slides in from bottom and slides out back.

Expected behaviour:
<img src="https://user-images.githubusercontent.com/6103621/168486905-29efe27f-ee2c-411e-895f-5ff79b5017dc.gif" width="300"></img>

Slide in animation works fine, but slide out animation doesn't - because appearing view seems to have higher `zIndex`, no matter wether it is pushing or popping.

Actual behaviour:
<img src="https://user-images.githubusercontent.com/6103621/168486988-21aaa825-34b5-4948-b673-1d7a9f2faf52.gif" width="300"></img>

I've tried specifying it with a modifier, it didn't worked out. In my case I need `zIndex` to be equal to backstack index, in other words during pop disappearing view should have higher zIndex, than the appearing one.
```kotlin
object Destinations {
    const val Login = "login"
    const val Home = "home"
}

@Composable
fun NavTestScreen(
) {
    val navController = rememberAnimatedNavController()
    AnimatedNavHost(navController = navController,
        startDestination = Destinations.Login,
        modifier = Modifier.fillMaxSize()) {
        composable(
            Destinations.Login,
            enterTransition = { NavigationTransition.IdentityEnter },
            exitTransition = { NavigationTransition.IdentityExit },
        ) {
            Button(onClick = {
                navController.navigate(Destinations.Home)
            }) {
                Text(text = "Next")
            }
        }
        composable(
            route = Destinations.Home,
            enterTransition = { NavigationTransition.slideInBottomAnimation },
            exitTransition = { NavigationTransition.slideOutBottomAnimation },
        ) {
            Button(
                onClick = {
                    navController.popBackStack()
                },
                colors = ButtonDefaults.buttonColors(backgroundColor = Color.Yellow),
                modifier = Modifier.zIndex(100f)
            ) {
                Text(text = "label")
            }
        }
    }
}

object NavigationTransition {
    private val animation: FiniteAnimationSpec<IntOffset> = tween(
        easing = LinearOutSlowInEasing,
        durationMillis = 2000,
    )

    val IdentityEnter = slideInVertically(
        initialOffsetY = {
            -1 //fix for https://github.com/google/accompanist/issues/1159
        },
        animationSpec = animation
    )

    val IdentityExit = slideOutVertically(
        targetOffsetY = {
            -1 //fix for https://github.com/google/accompanist/issues/1159
[...]

PR #1384, opened 2022-11-02
Base z-order on visibleEntries

jbw0033:
NavHost should respect the z-order established by visibleEntries. This ensures that incoming Composables will be on top of exiting ones, except in the case of a pop it is expected that it is reversed.

Fixes: #1160


alitahabundle:
@ianhanniballake any updates?

morganbovi:
@ianhanniballake bump

ianhanniballake (approved):

PR #1392, opened 2022-11-10
[Test Harness] Add test harness library

alexvanyo:
Adds a new testing library, `testharness` for testing Compose layouts.


yschimke:
Want to add isRound? :)

bentrengrove (changes requested):
Please add an issue template to .github/ISSUE_TEMPLATE and also an entry for the docs generation into mkdocs.yml

Also, is there no sample for this project?

alexvanyo:
> Want to add isRound? :)

For `isRound`, I'd want to do more investigation with how the APIs handle rendering outside of the round viewport on a circular watch.

For example: if you render something in the very corner of a physically round watch (such that it isn't actually visible), do the testing APIs treat it as being visible?

If they do, the `isRound` should be just a matter of flipping the `Configuration` flag, otherwise we could inset the content within the round viewport.

alexvanyo:
> Please add an issue template to .github/ISSUE_TEMPLATE and also an entry for the docs generation into mkdocs.yml

Added, along with a first draft of documentation as well.

> Also, is there no sample for this project?

There are the tests of the `TestHarness` itself, and I converted some tests in `pager` and `adaptive` where it makes sense to use `TestHarness`, but agreed there should be more.

Do you think it makes sense to add some tests to `sample` that make use of it? I was thinking a navigation rail versus navigation bar test setup would be a good showcase of it.

bentrengrove (approved):
LGTM! Amazing work.

For the sample app, if you don't want to add a test in the sample app, just a note in your docs about where to see a sample in the repo would be good.

yschimke:
> > Want to add isRound? :)
> 
> For `isRound`, I'd want to do more investigation with how the APIs handle rendering outside of the round viewport on a circular watch.
> 
> For example: if you render something in the very corner of a physically round watch (such that it isn't actually visible), do the testing APIs treat it as being visible?
> 
> If they do, the `isRound` should be just a matter of flipping the `Configuration` flag, otherwise we could inset the content within the round viewport.

Yep, I can follow up.

The answer is yes, they can draw outside the bounds and screenshot tools show content outside.  So for Paparazzi I change both the configuration as you suggest, but also apply a circular crop.

JoseAlcerreca:
Given the sample is a catalog, we could add a testharness sample (with a big warning so that it's not used in prod)

image

Once I fix the darkMode issue...

Thoughts?

JoseAlcerreca:
I've pushed it anyway, but happy to revert.

yschimke:
I tried this to see if it could replace my bespoke code.  Not sure if it's AGP or paparazzi related (I doubt given it's debug code, not test). But seeing

```
java.lang.NullPointerException: Cannot invoke "android.content.Context.getResources()" because "resContext" is null   at android.view.ContextThemeWrapper.getResourcesInternal(ContextThemeWrapper.java:137)   at android.view.ContextThemeWrapper.getResources(ContextThemeWrapper.java:128)   at androidx.compose.ui.res.Resources_androidKt.resources(Resources.android.kt:33)   at androidx.compose.ui.res.StringResources_androidKt.stringResource(StringResources.android.kt:35)
```

Hopefully some typo in my copy/paste/simplification. But wanted to flag.

JoseAlcerreca:
> Hopefully some typo in my copy/paste/simplification. But wanted to flag.

I can't think of anything that the harness might be doing that might be affecting resource loading. Can you share your code?

yschimke:
> > Hopefully some typo in my copy/paste/simplification. But wanted to flag.
> 
> I can't think of anything that the harness might be doing that might be affecting resource loading. Can you share your code?

https://github.com/yschimke/horologist/commit/155971e206ce01afa69238ba9ced67e44397c719

Don't block on this, mainly flagging in case it's a more general problem.

yschimke:
For completeness I would suggest one of the tests being to change the locale and call stringResource() from a composable.

JoseAlcerreca:
Done 
image

alexvanyo:
I was trying out `TestHarness` in a `@Preview`, and ran into a similar looking error.

I believe the issue is the same, caused by [this line](https://cs.android.com/android/platform/superproject/+/master:frameworks/layoutlib/bridge/src/com/android/layoutlib/bridge/android/BridgeContext.java;l=1380;drc=857b1f380d02a7d1dab011850257d472a27370cc) in layoutlib.

`ContextThemeWrapper` [uses](https://cs.android.com/android/platform/superproject/+/master:frameworks/base/core/java/android/view/ContextThemeWrapper.java;l=136;drc=ad9254cb1d45844e0a3b7470292b77d34d31e962) `Context.createConfigurationContext` to apply the override, and it looks like this isn't supported in layoutlib directly, and instead, `null` is returned which leads to the crash.

A proper fix would likely be adding some functionality to layoutlib, to support `ContextThemeWrapper`? https://issuetracker.google.com/issues/200845657 seems similar.

alexvanyo:
List of updates since your approval @bentrengrove :
- Added test harness sample
- Added locale overriding to sample and tests
- Documented current `@Preview` limitation

JoseAlcerreca:
Got the Trengrove :rocket: of approval offline, merging!

testharness/src/main/java/com/google/accompanist/testharness/TestHarness.kt
```diff
@@ -0,0 +1,209 @@
...
+                    Configuration.UI_MODE_NIGHT_YES
+                } else {
+                    Configuration.UI_MODE_NIGHT_NO
+                }
+                // Update the locale list
+                if (Build.VERSION.SDK_INT >= 24) {
+                    setLocales(LocaleList.forLanguageTags(locales.toLanguageTags()))
+                } else {
+                    setLocale(locales[0])
+                }
```
alashow, line 109: @alexvanyo `java.util.Locale.setDefault(locales[0])` would be also useful here since java formatters won't know about the locale update otherwise

alexvanyo: I'd be interested to hear your use case here, since I'd intentionally prefer to avoid calling `java.util.Locale.setDefault(locales[0])` if possible.

`java.util.Locale.getDefault()` isn't observable by Compose, so if you're calling that from a `@Composable` function, you might not be notified if the locale changes in the general case (for instance, if opting out of activity recreation for locale changes)

If possible, can you retrieve the locale from `LocalConfiguration.current.locales` instead to create the formatter?

Issue #1277, opened 2022-08-03
[System UI Controller] SystemUiController is annotated with @Stable but doesn't adhere to the contract


alexvanyo:
`SystemUiController` is annotated with `@Stable`, but doesn't adhere to the contract. Specifically, it has public properties that aren't state aware and won't notify composition when they change, since they're backed by direct platform API calls.

Nek-12:
@bentrengrove I like how people are having to remove the stale label every month, but you keep refusing to disable that bot. This time, it seems to be a record - 7 times with no other activity.

Nek-12:
...

PR #1394, opened 2022-11-13
[SystemUiController] Provide current status bar and navigation bar colors

### LINKED ISSUE #1102: [System UI Controller] Request - Provide current statusBarColor, navigationBarColor
**Description**
It would be useful if SystemUiController could be used to get current values statusBarColor, navigationBarColor. 
One usecase would be user moving from some main screen which has already set e.g. statusBarColor to a different screen which wants to set a different statusBar color and have possibility of reverting back to orig value

**Steps to reproduce**

**Expected behavior** 

**Additional context**

svenjacobs:
Fix #1102

### DIFF
--- a/systemuicontroller/api/current.api
+++ b/systemuicontroller/api/current.api
@@ -2,7 +2,9 @@
 package com.google.accompanist.systemuicontroller {
 
   @androidx.compose.runtime.Stable public interface SystemUiController {
+    method public androidx.compose.ui.graphics.Color? getNavigationBarColor();
     method public boolean getNavigationBarDarkContentEnabled();
+    method public androidx.compose.ui.graphics.Color? getStatusBarColor();
     method public boolean getStatusBarDarkContentEnabled();
     method public int getSystemBarsBehavior();
     method public default boolean getSystemBarsDarkContentEnabled();
@@ -25,7 +27,9 @@ package com.google.accompanist.systemuicontroller {
     property public abstract boolean isNavigationBarVisible;
     property public abstract boolean isStatusBarVisible;
     property public default boolean isSystemBarsVisible;
+    property public abstract androidx.compose.ui.graphics.Color? navigationBarColor;
     property public abstract boolean navigationBarDarkContentEnabled;
+    property public abstract androidx.compose.ui.graphics.Color? statusBarColor;
     property public abstract boolean statusBarDarkContentEnabled;
     property public abstract int systemBarsBehavior;
     property public default boolean systemBarsDarkContentEnabled;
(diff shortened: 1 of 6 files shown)


alexvanyo:
The code here looks good, my only concern is that this is a further violation of #1277 .
In other words, even though `SystemUiController` promises that it is `@Stable`, there isn't a mechanism to notify Compose when the system bar colors changes.

@bentrengrove do you think we should merge as is, considering that other parameters also have that issue? Ultimately we'll need a proper solution to solve #1277 in a way that doesn't break the contract.

svenjacobs:
Use there and platform API that notifies when status bar or navigation bar colors change? If so, then these properties could be changed to State?

bentrengrove:
Yeah I am a bit worried that people would expect these to update like Compose state. I don't believe that there is an API you could hook in to but @alexvanyo might know.

svenjacobs:
> Yeah I am a bit worried that people would expect these to update like Compose state.

We could explicitly document this ("this is not a state").

alexvanyo:
I'm not aware of such an API either.

Unfortunately `@Stable` is already a piece of documentation and telling the compiler that it would be state observable to Compose, so adding additional documentation doesn't really help that much.

There's probably an API evolution here where we make `SystemUiController` use Compose observable state that syncs to the window, but where it is only valid if there is a single instance of the `SystemUiController` that is making changes to the window attributes. Otherwise, the `SystemUiController` may get out of sync with the underlying window attributes.

alexvanyo:
We've decided to close this due to the issue of the system UI controller not being able to adhere to the `@Stable` contract for the colors (#1277).

As-is, there would be the expectation as enforced by the compiler that the following would work and be updated whenever the status bar color changes:
```kotlin
fun Test(systemUiController: SystemUiController) {
    Text("Status bar color: ${systemUiController.statusBarColor}")
}
```

There are already properties in system UI controller that also violate the contract, but we don't want to violate the contract further in the library if we can help it.

As a workaround, it's fairly straightforward to get the status bar color directly from `WindowInsetsControllerCompat`, using the same method that system ui controller uses, that is more clear that this is a one-time retrieval:

```kotlin
@Composable
fun MyApp() {
    val window = findWindow()
    val view = LocalView.current
    val windowInsetsControllerCompat = remember(view, window) {
        window?.let {
            WindowCompat.getInsetsController(it, view)
        }
    }
    // Use WindowInsetsControllerCompat methods
}

@Composable
private fun findWindow(): Window? =
    (LocalView.current.parent as? DialogWindowProvider)?.window
        ?: LocalView.current.context.findWindow()

private tailrec fun Context.findWindow(): Window? =
    when (this) {
        is Activity -> window
        is ContextWrapper -> baseContext.findWindow()
        else -> null
    }
```

Thanks for helping us work through this, and I'll keep #1102 open because it is a reasonable request for a form of system UI controller that acts as a singular manager of the window parameters as part of a larger refactor.

PR #1397, opened 2022-11-14
[Web] Fix WebView always filling max available space

### LINKED ISSUE #1297: WebView created with html data always takes up the whole screen
## Describe the bug
As of right now the LayoutParams are created inside the WebView composable and cannot be changed. The layoutParams are MATCH_PARENT, so the result is that no matter how big is the html content is, it will always take up the whole screen.

## To Reproduce
1. Embed WebView inside ModalBottomSheetLayout
2. Create state by using rememberWebViewStateWithHTMLData with data equal to some small html content and pass the state to the WebVIew
3. Run the project, open the bottom sheet with webview and see that it takes up the whole screen

## Expected behavior
The bottom sheet should be the height of the html content that was passed into the WebView. 

## Environment:
 - Android OS version: Android 11.0
 - Device: Emulator
 - Accompanist version: 0.24.13-rc

## Additional context
My proposal is to pass the LayoutParams as the argument to the WebView, so the programmers would be able to configure it according to their needs.

bentrengrove:
Fixes #1297

WebView changes its layout engine based on its layoutParams. This PR converts from a Compose modifier into view layoutParams in order to support this. I thought this was a better integration than simply exposing the layoutParams parameter.

Screenshot_20221114_133703

### DIFF
--- a/sample/src/main/AndroidManifest.xml
+++ b/sample/src/main/AndroidManifest.xml
@@ -16,6 +16,7 @@
   -->
 
 <manifest xmlns:android="http://schemas.android.com/apk/res/android"
+    xmlns:tools="http://schemas.android.com/tools"
     package="com.google.accompanist.sample">
 
     <!-- Used for loading images from the network -->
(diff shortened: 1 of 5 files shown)


arriolac (approved):
LGTM

PR #1398, opened 2022-11-15
Refactor webview to not need to override url loading

### LINKED ISSUE #1433: [WebView] When using WebContent.Data , mimeType and encoding cannot be specified.
```
            is WebContent.Data -> {
                view.loadDataWithBaseURL(content.baseUrl, content.data, null, "utf-8", null)
            }
```

When using WebContent.Data , mimeType and encoding cannot be specified. Is this the intended code?

### LINKED ISSUE #1476: [WebView] with input type file dont work
**Description**
When using webview with form that have files field(for upload) nothing happend when pressing on the input(should open the image picker), There is a way to fix this?

bentrengrove:
Refactors WebView to not have to use `overrideUrlLoading`. This was leading to a lot of navigation edge cases that weren't properly handled. The new wrapping method should provide a more stable foundation to build upon.

| Deprecated | Replacement |
|---|---|
| **state.content.getCurrentUrl()** - **state.content** will also no longer track what url the webview is showing | **state.lastLoadedUrl** - Compose state object for the last url loaded by the WebView |

Fixes not following all redirects correctly
Fixes #1433
Fixes #1476

### DIFF
--- a/sample/src/main/java/com/google/accompanist/sample/webview/BasicWebViewSample.kt
+++ b/sample/src/main/java/com/google/accompanist/sample/webview/BasicWebViewSample.kt
@@ -59,16 +59,16 @@ import com.google.accompanist.web.rememberWebViewNavigator
 import com.google.accompanist.web.rememberWebViewState
 
 class BasicWebViewSample : ComponentActivity() {
+    val initialUrl = "https://google.com"
     @SuppressLint("SetJavaScriptEnabled")
     override fun onCreate(savedInstanceState: Bundle?) {
         super.onCreate(savedInstanceState)
         setContent {
             AccompanistSampleTheme {
-                var url by remember { mutableStateOf("https://google.com") }
-                val state = rememberWebViewState(url = url)
+                val state = rememberWebViewState(url = initialUrl)
                 val navigator = rememberWebViewNavigator()
-                var textFieldValue by remember(state.content.getCurrentUrl()) {
-                    mutableStateOf(state.content.getCurrentUrl() ?: "")
+                var textFieldValue by remember(state.lastLoadedUrl) {
+                    mutableStateOf(state.lastLoadedUrl)
                 }
 
                 Column {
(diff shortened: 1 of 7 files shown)


web/src/androidTest/kotlin/com/google/accompanist/web/WebTest.kt
```diff
@@ -71,7 +72,7 @@ import java.util.concurrent.TimeUnit
 // Emulator image doesn't have a WebView until API 26
 // Google API emulator image seems to be really flaky before 28 so currently we will set these tests
 // to min 29 and max 30. 31/32 image is also really flaky
```
arriolac, line 74: Update the comment here since `maxSdkVersion` was removed?

web/src/androidTest/kotlin/com/google/accompanist/web/WebTest.kt
```diff
@@ -465,19 +467,18 @@ class WebTest {
 
         rule.waitForIdle()
 
-        onWebView()
-            .withElement(findElement(Locator.ID, "link"))
-            .perform(webClick())
+        // HACK: EspressoWeb webClick doesn't track back navigation for some reason
+        // so manually click the view
+        rule.onNodeWithTag(WebViewTag).performClick()
```
arriolac, line 472: How about making an extension method for this and having the comment visible once there?

arriolac (approved):
LGTM just minor comments

sample/src/main/java/com/google/accompanist/sample/webview/BasicWebViewSample.kt
```diff
@@ -100,15 +100,17 @@ class BasicWebViewSample : ComponentActivity() {
                             }
 
                             OutlinedTextField(
-                                value = textFieldValue,
+                                value = textFieldValue ?: "",
```
kerembalaban, line 103: you can use textFieldValue.orEmpty() instead of textFieldValue ?: ""

PR #1414, opened 2022-11-23
[Test Harness] Add custom ContextThemeWrapper to improve test harness behavior in previews

alexvanyo:
Adds a custom `ContextThemeWrapper` to work around the issue that currently causes `TestHarness` to fail in previews.

The cause of the crash is that `Context.createConfigurationContext` returns `null` for `BridgeContext` in layoutlib: https://cs.android.com/android/platform/superproject/+/master:frameworks/layoutlib/bridge/src/com/android/layoutlib/bridge/android/BridgeContext.java;l=1380;drc=58bd418e161b5d4fdde39b46523d3e49269117e3

`ContextThemeWrapper` uses this `Context.createConfigurationContext` internally and assumes it is not null, which causes the crash.

The `androidx.appcompat.view.ContextThemeWrapper` has a fallback mechanism for older API versions using a deprecated `Resources` constructor, so this PR adds a custom `ContextThemeWrapper` based on the appcompat one that uses that fallback mechanism in case `Context.createConfigurationContext` is null.

This improves behavior, preventing the crash, but doesn't remove all of the limitations. In particular, alternate resources are not loaded (such as the string not translated in the below image), which is probably related to how layoutlib manages resources.

Screen Shot 2022-11-22 at 12 05 11 PM

### DIFF
--- a/testharness/src/main/java/com/google/accompanist/testharness/ContextThemeWrapper.java
+++ b/testharness/src/main/java/com/google/accompanist/testharness/ContextThemeWrapper.java
@@ -1,0 +1,238 @@
+/*
+ * Copyright 2022 The Android Open Source Project
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.google.accompanist.testharness;
+
+import android.content.Context;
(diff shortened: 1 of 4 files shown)


testharness/src/main/java/com/google/accompanist/testharness/ContextThemeWrapper.java
```diff
@@ -0,0 +1,238 @@
...
+import androidx.annotation.RequiresApi;
+import androidx.annotation.StyleRes;
+
+/**
+ * A context wrapper that allows you to modify or replace the theme of the wrapped context.
+ * <p>
+ * This is a modification of androidx.appcompat.view.ContextThemeWrapper to allow the
+ * TestHarness to work in previews.
+ */
+class ContextThemeWrapper extends ContextWrapper {
```
alexvanyo, line 36: I kept this class in `.java`, to be closer to the appcompat `ContextThemeWrapper`. I could convert and simplify, if desired.

bentrengrove: If you are happy to maintain Java code, fine by me!

bentrengrove (approved):

PR #1429, opened 2022-12-01
[testharness] Add support for Wear to test harness

yschimke:
Add support for Wear to test harness.  

To replace existing test libraries like https://github.com/google/horologist/blob/1ab282cf951b0386650acfff3ee8ed8acaf2ea2d/compose-tools/src/main/java/com/google/android/horologist/compose/tools/RoundPreview.kt

### DIFF
--- a/adaptive/src/sharedTest/kotlin/com/google/accompanist/adaptive/WearTest.kt
+++ b/adaptive/src/sharedTest/kotlin/com/google/accompanist/adaptive/WearTest.kt
@@ -16,15 +16,13 @@
 
 package com.google.accompanist.adaptive
 
-import android.os.Build
 import androidx.activity.ComponentActivity
 import androidx.compose.ui.platform.LocalConfiguration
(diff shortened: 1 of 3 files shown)


adaptive/src/sharedTest/kotlin/com/google/accompanist/adaptive/WearTest.kt
```diff
@@ -0,0 +1,60 @@
...
+
+@RunWith(AndroidJUnit4::class)
+class WearTest {
+
+    @get:Rule
+    val composeTestRule = createAndroidComposeRule<ComponentActivity>()
+
+    @Test
+    fun is_round_for_wear() {
+        assumeTrue(Build.VERSION.SDK_INT >= 23)
```
alexvanyo, line 40: Elsewhere in the project we've been using `@SdkSuppress`.

`assume` should be equivalent functionality, but maybe it'd be best to stick with `@SdkSuppress` for consistency?

alexvanyo (approved):

PR #1437, opened 2022-12-08
[Navigation Material] Make BottomSheetNavigatorSheetState can manually show or hide

Omico:
Currently, I cannot find a method to show or hide the bottom sheet manually.

Usage:
```
val bottomSheetNavigator: BottomSheetNavigator = rememberBottomSheetNavigator()
val coroutineScope: CoroutineScope = rememberCoroutineScope()
Button(
    onClick = {
        coroutineScope.launch {
            bottomSheetNavigator.navigatorSheetState.show() // or hide()
        }
    },
    content = { Text(text = "Button") },
)
``

### DIFF
--- a/navigation-material/api/current.api
+++ b/navigation-material/api/current.api
@@ -30,7 +30,9 @@ package com.google.accompanist.navigation.material {
     method public androidx.compose.runtime.State<java.lang.Float> getOverflow();
     method public androidx.compose.material.SwipeProgress<androidx.compose.material.ModalBottomSheetValue> getProgress();
     method public androidx.compose.material.ModalBottomSheetValue getTargetValue();
+    method public suspend Object? hide(kotlin.coroutines.Continuation<? super kotlin.Unit> p);
     method public boolean isVisible();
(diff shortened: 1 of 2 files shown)


ianhanniballake:
You show and hide a bottom sheet destination by, respectively, navigating to a bottom sheet destination or popping a bottom sheet destination by using the `NavController` APIs. It is expected that you don't directly access `show()` or `hide()`.

Omico:
Hello @ianhanniballake, if we use `NavController` APIs, it won't trigger `ModalBottomSheetState.hide()`; thus, there is no animation.

jossiwolf:
We currently don't have a hide animation due to upstream bugs. Snapping to the hidden state is a workaround. These issues will be fixed with the next release of Compose. Exposing this will only work for some cases, so until we have a new release we will keep this workaround and then improve. It does not warrant exposing and maintaining a new API :)

ianhanniballake:
Feel free to follow along on the transition from `snapTo` to `hide()` in #724

PR #1438, opened 2022-12-08
Add minSdk info to AndroidManifest

bentrengrove:
This makes it easier to import Accompanist into non-gradle build environments


arriolac (approved):
LGTM. I counted 30+ `AndroidManifest.xml` files in the repo though—do we need to do the same for other files?

bentrengrove:
Only needed for the actual main library implementations

PR #1444, opened 2022-12-09
Convert to namespace

### LINKED ISSUE #1443: Convert all modules to use namespace instead of package in AndroidManifest.xml
This will be required in AGP 8.0, so converting earlier will keep us up to date and also make it easier to test out newer releases of AGP (for testing GMD fixes, for instance)

alexvanyo:
Converts projects from using packages in the `AndroidManifest.xml`s to using `namespace` in the build.gradle files, to fix #1443


bentrengrove (approved):
This is great, thanks for doing this.

Issue #772, opened 2021-10-05
BottomSheet route is not displayed


nglauber:
## Describe the bug
Bottom sheet route is not displayed using a combination of `LaunchedEffect`, `flow.collectAsState()` and `Text`. 

## To Reproduce
1. Declare a `bottomSheet` route inside of the `AnimatedNavHost`
```
ModalBottomSheetLayout(bottomSheetNavigator) {
    AnimatedNavHost(...) {
        bottomSheet("bottomSheetNav") { BottomSheetViaRoute(it) }
    }
}
```
2. Navigate to this route using:
```
navController.navigate("bottomSheetNav")
```
3. The composable function `BottomSheetViaRoute` should be like this:
```
@Composable
fun BottomSheetViaRoute(
    navBackStackEntry: NavBackStackEntry
) {
    val viewModel: MyViewModel = viewModel(navBackStackEntry)
    val list by viewModel.listFlow.collectAsState(emptyList())
    LaunchedEffect(Unit) {
        viewModel.loadList()
    }
    Column(Modifier.verticalScroll(rememberScrollState())) {
        //Text(text = "BottomSheet via route", style = MaterialTheme.typography.h3)
        list.forEach {
            Text(
                text = it,
                Modifier
                    .padding(16.dp)
                    .fillMaxWidth()
            )
        }
    }
}
```
4. Notice the `Text` is commented...
5. Run this code and it should work as expected.
6. Now, uncomment the call to the `Text` function.

## Expected behavior
The bottom sheet should be displayed with the `Text` on top of the items.

## Current behavior
Nothing happens... :/ 

## Screenshots?
N/A

## Environment:
 - Android OS version: 8.0
 - Device: Motorola moto Z2 Play
 - Accompanist version: 0.19.0

## Additional context
The full code of the `BottomSheetViaRoute` function can be found [here](https://github.com/nglauber/JetpackComposePlayground/blob/master/app/src/main/java/br/com/nglauber/jetpackcomposeplayground/screens/BottomSheetViaRoute.kt).
The full code of the `MyViewModel` class can be found [here](https://github.com/nglauber/JetpackComposePlayground/blob/master/app/src/main/java/br/com/nglauber/jetpackcomposeplayground/screens/MyViewModel.kt).

I also tried comment the `list.forEach` call. And it worked... I mean, only `Text` and only `list.forEach` work, but not both...

jossiwolf:
[...]

Issue #1356, opened 2022-09-28
[Navigation Material] Unexpected behavior when navigate to bottom sheet and hide keyboard


imholynx:
**Description**
What I have:

- Home Screen with TextField
- Bottom sheet
- `android:windowSoftInputMode="adjustResize"` for activity
- `LocalSoftwareKeyboardController.current` to hide keyboard

**Steps to reproduce**

1. Tap on TextField to open keyboard
2. Navigate to bottom sheet and hide keyboard in same time (onTap)
3. Keyboard is hided but Bottom sheet is invisible

**Expected behavior** 

Keyboard is hidden
Bottom sheet is visible

**Additional context**
    ```implementation "com.google.accompanist:accompanist-navigation-material:0.23.1"```

Sample: https://github.com/imholynx/AccompanistNavigationBugSample

Source code to reproduce:
```
class MainActivity : ComponentActivity() {
    @OptIn(ExperimentalMaterialNavigationApi::class)
    override fun onCreate(savedInstanceState: Bundle?) {
        super.onCreate(savedInstanceState)
        setContent {
            MyApplicationTheme() {
                val bottomSheetNavigator = rememberBottomSheetNavigator()
                val navController = rememberNavController(bottomSheetNavigator)

                ModalBottomSheetLayout(bottomSheetNavigator) {
                    NavHost(navController, "home") {
                        composable("home") { Home(navController) }
                        bottomSheet("bottom_sheet") { BottomSheet() }
                    }
                }
            }
        }
    }
}

@OptIn(ExperimentalComposeUiApi::class)
@Composable
private fun Home(navController: NavController) {
    val keyboardController = LocalSoftwareKeyboardController.current

    LazyColumn {
        item { TextField(value = "", onValueChange = {}) }
        repeat(99) {
            item {
                Text(text = it.toString(),
                    modifier = Modifier
                        .fillMaxWidth()
                        .padding(16.dp)
                        .clickable {
                            keyboardController?.hide()
                            navController.navigate("bottom_sheet")
                        })
            }
        }
    }
}

@Composable
private fun BottomSheet() {
    LazyColumn {
        repeat(9) {
            item { Text(text = it.toString()) }
        }
    }
}
```

https://user-images.githubusercontent.com/1606106/192773388-c7e5e47f-2cfa-4d54-9c8b-c3d85214d835.mp4

jossiwolf:
[...]

Issue #1147, opened 2022-04-30
[Navigation Material] App crashes on rotation when using Scaffold and a "tall" bottom sheet


raamcosta:
**Description**
If we have a bottom sheet tall enough (I'm guessing more than half) and we wrap our NavHost in a Scaffold, then when rotating the device while the bottom sheet is the active destination, the app crashes.

<details>
  <summary>Click to expand crash stack trace</summary>

```
2022-04-29 16:37:55.046 30964-30964/com.ramcosta.samples.playground E/AndroidRuntime: FATAL EXCEPTION: main
    Process: com.ramcosta.samples.playground, PID: 30964
    java.lang.IllegalArgumentException: The initial value must have an associated anchor.
        at androidx.compose.material.SwipeableState.ensureInit$material_release(Swipeable.kt:138)
        at androidx.compose.material.SwipeableKt$swipeable$3.invoke(Swipeable.kt:594)
        at androidx.compose.material.SwipeableKt$swipeable$3.invoke(Swipeable.kt:573)
        at androidx.compose.ui.ComposedModifierKt$materialize$result$1.invoke(ComposedModifier.kt:263)
        at androidx.compose.ui.ComposedModifierKt$materialize$result$1.invoke(ComposedModifier.kt:258)
        at androidx.compose.ui.Modifier$Element$DefaultImpls.foldIn(Modifier.kt:107)
        at androidx.compose.ui.ComposedModifier.foldIn(ComposedModifier.kt:172)
        at androidx.compose.ui.CombinedModifier.foldIn(Modifier.kt:149)
        at androidx.compose.ui.CombinedModifier.foldIn(Modifier.kt:149)
        at androidx.compose.ui.CombinedModifier.foldIn(Modifier.kt:149)
        at androidx.compose.ui.CombinedModifier.foldIn(Modifier.kt:149)
        at androidx.compose.ui.CombinedModifier.foldIn(Modifier.kt:149)
        at androidx.compose.ui.CombinedModifier.foldIn(Modifier.kt:149)
        at androidx.compose.ui.CombinedModifier.foldIn(Modifier.kt:149)
        at androidx.compose.ui.CombinedModifier.foldIn(Modifier.kt:149)
        at androidx.compose.ui.CombinedModifier.foldIn(Modifier.kt:149)
        at androidx.compose.ui.ComposedModifierKt.materialize(ComposedModifier.kt:258)
[...]

Issue #1031, opened 2022-02-16
[Navigation] sheet destination not popped when rapidly dismissing sheet 


ind1go:
**Description**

I have a layout where a FAB triggers the bottom sheet, and then disappears so as not to cover it. When the sheet is dismissed, the button reappears (see the video under 'expected behaviour').

This works most of the time, however if the sheet is dismissed at a particular moment shortly after it's started, it seems that the the `currentBackStackEntryAsState` state isn't updated back to the prior destination, and therefore the action to show the button isn't run. There appears to be a race condition.

Video of the failing case:

https://user-images.githubusercontent.com/1038350/154358665-7326cd6e-6a35-485c-b0c3-65fbf9313a2b.mp4

**Steps to reproduce**

1. Press the button to switch to the sheet and hide the button.
2. Press the scrim, just about at the time when the sheet animation stops.
3. Observe that sheet is dismissed, but the button has not reappeared.

Also, in terms of a bit of troubleshooting...

4. If I add a `println` as follows:
    ```kotlin
    val navBackStackEntry by navController.currentBackStackEntryAsState() // existing line
    println("Route: ${navBackStackEntry?.destination?.route}") // new println
    ```
    within my `MainLayout`, then in the erroneous case it ends on `SHEET`. The `HOME` navigation does not appear to be triggered.

Full code below.

```kotlin
import android.os.Bundle
import androidx.activity.ComponentActivity
import androidx.activity.compose.setContent
import androidx.compose.material.icons.Icons
import androidx.compose.material.icons.filled.Star
import androidx.compose.material3.*
import androidx.compose.runtime.*
import androidx.compose.ui.graphics.vector.rememberVectorPainter
import androidx.compose.ui.unit.sp
import androidx.navigation.NavHostController
import androidx.navigation.compose.NavHost
import androidx.navigation.compose.composable
import androidx.navigation.compose.currentBackStackEntryAsState
import androidx.navigation.compose.rememberNavController
import com.google.accompanist.navigation.material.*

class TestActivity2 : ComponentActivity() {

    @OptIn(ExperimentalMaterialNavigationApi::class, ExperimentalMaterial3Api::class)
    override fun onCreate(savedInstanceState: Bundle?) {
        super.onCreate(savedInstanceState)
        setContent {
            val bottomSheetNavigator = rememberBottomSheetNavigator()
            val navController = rememberNavController(bottomSheetNavigator)
            var fabVisible by remember { mutableStateOf(true) }

            Scaffold(
[...]

PR #1465, opened 2023-01-05
[Navigation Material] Update Navigation Material after ModalBottomSheetLayout bug fixes

jossiwolf:
This PR refactors our Navigation Material code after upstream updates to ModalBottomSheetLayout that address one-frame consistency and anchor update + reconciliation logic.

We now retain the back stack entry until all animations are completed.

### API Changes
|API|Status|
|---|---|
|public ctor BottomSheetNavigatorSheetState|internal ctor|
|BottomSheetNavigatorSheetState#offset|Deprecated|
|BottomSheetNavigatorSheetState#direction|Deprecated|
|BottomSheetNavigatorSheetState#progress|Deprecated|

### Other changes
- BottomSheetNavigator now uses moveable content under the hood
- The strategy for retaining bottom sheet back stack entries has changed to only dispose of them after the sheet has been hidden

Fixes: #772, #813, #910, #937, #1031, #1147, #1356

### DIFF
--- a/navigation-material/src/androidTest/java/com/google/accompanist/navigation.material/BottomSheetNavigatorTest.kt
+++ b/navigation-material/src/androidTest/java/com/google/accompanist/navigation.material/BottomSheetNavigatorTest.kt
@@ -19,9 +19,13 @@ package com.google.accompanist.navigation.material
 import android.os.Bundle
 import androidx.activity.OnBackPressedDispatcher
 import androidx.activity.compose.LocalOnBackPressedDispatcherOwner
+import androidx.compose.animation.core.tween
+import androidx.compose.foundation.background
 import androidx.compose.foundation.layout.Box
 import androidx.compose.foundation.layout.Column
 import androidx.compose.foundation.layout.fillMaxSize
+import androidx.compose.foundation.layout.fillMaxWidth
+import androidx.compose.foundation.layout.height
 import androidx.compose.foundation.layout.size
 import androidx.compose.material.Button
 import androidx.compose.material.ExperimentalMaterialApi
@@ -34,19 +38,25 @@ import androidx.compose.runtime.mutableStateOf
 import androidx.compose.runtime.remember
 import androidx.compose.runtime.setValue
 import androidx.compose.ui.Modifier
+import androidx.compose.ui.graphics.Color
 import androidx.compose.ui.platform.testTag
+import androidx.compose.ui.test.click
+import androidx.compose.ui.test.getUnclippedBoundsInRoot
 import androidx.compose.ui.test.junit4.createComposeRule
 import androidx.compose.ui.test.onNodeWithTag
 import androidx.compose.ui.test.onNodeWithText
+import androidx.compose.ui.test.onRoot
 import androidx.compose.ui.test.performClick
+import androidx.compose.ui.test.performTouchInput
+import androidx.compose.ui.unit.Dp
 import androidx.compose.ui.unit.dp
+import androidx.compose.ui.unit.height
 import androidx.lifecycle.Lifecycle
 import androidx.navigation.NavBackStackEntry
 import androidx.navigation.NavHostController
 import androidx.navigation.compose.NavHost
 import androidx.navigation.compose.composable
 import androidx.navigation.compose.rememberNavController
-import androidx.navigation.plusAssign
 import androidx.navigation.testing.TestNavigatorState
 import androidx.test.ext.junit.runners.AndroidJUnit4
 import androidx.test.filters.LargeTest
@@ -56,6 +66,7 @@ import kotlinx.coroutines.runBlocking
 import org.junit.Rule
 import org.junit.Test
 import org.junit.runner.RunWith
+import kotlin.math.roundToLong
 
 @LargeTest
 @RunWith(AndroidJUnit4::class)
@@ -72,7 +83,7 @@ internal class BottomSheetNavigatorTest {
         val navigator = BottomSheetNavigator(sheetState)
 
         navigator.onAttach(navigatorState)
-        val entry = navigatorState.createBackStackEntry(navigator.createDestination(), null)
+        val entry = navigatorState.createBackStackEntry(navigator.createFakeDestination(), null)
         navigator.navigate(listOf(entry), null, null)
 
         assertWithMessage("The back stack entry has been added to the back stack")
@@ -87,8 +98,8 @@ internal class BottomSheetNavigatorTest {
         val navigatorState = TestNavigatorState()
 
         navigator.onAttach(navigatorState)
-        val firstEntry = navigatorState.createBackStackEntry(navigator.createDestination(), null)
-        val secondEntry = navigatorState.createBackStackEntry(navigator.createDestination(), null)
+        val firstEntry = navigatorState.createBackStackEntry(navigator.createFakeDestination(), null)
+        val secondEntry = navigatorState.createBackStackEntry(navigator.createFakeDestination(), null)
 
         navigator.navigate(listOf(firstEntry), null, null)
         assertWithMessage("The first entry has been added to the back stack")
@@ -134,7 +145,11 @@ internal class BottomSheetNavigatorTest {
                 secondDestinationCompositions++
                 onDispose { secondDestinationCompositions = 0 }
             }
-            Box(Modifier.size(64.dp).testTag(secondDestinationContentTag))
+            Box(
+                Modifier
+                    .size(64.dp)
+                    .testTag(secondDestinationContentTag)
+            )
         }
         val secondEntry = navigatorState.createBackStackEntry(secondDestination, null)
 
@@ -171,11 +186,20 @@ internal class BottomSheetNavigatorTest {
         composeTestRule.setContent {
             ModalBottomSheetLayout(
                 bottomSheetNavigator = navigator,
-                content = { Box(Modifier.fillMaxSize().testTag(bodyContentTag)) }
+                content = {
+                    Box(
+                        Modifier
+                            .fillMaxSize()
+                            .testTag(bodyContentTag)
+                    )
+                }
             )
         }
 
-        val destination = navigator.createFakeDestination()
+        val destination = BottomSheetNavigator.Destination(
+            navigator = navigator,
+            content = { Box(Modifier.height(20.dp)) }
+        )
         val backStackEntry = navigatorState.createBackStackEntry(destination, null)
         navigator.navigate(listOf(backStackEntry), null, null)
         composeTestRule.awaitIdle()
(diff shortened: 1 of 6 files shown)


navigation-material/src/main/java/com/google/accompanist/navigation/material/BottomSheetNavigator.kt
```diff
@@ -165,63 +177,78 @@ public class BottomSheetNavigator(
...
-        DisposableEffect(backStackEntries) {
-            transitionsInProgressEntries.forEach {
-                if (it != latestEntry) state.markTransitionComplete(it)
+            val retainedEntry by produceState<NavBackStackEntry?>(initialValue = null) {
+                backStack
+                    .map { backStackEntries ->
+                        backStackEntries.lastOrNull { entry ->
+                            // We might have entries in the back stack that aren't started
+                            // currently, so filter these
+                            entry.lifecycle.currentState.isAtLeast(Lifecycle.State.INITIALIZED)
```
jossiwolf, line 197: @jbw0033 in tests (specifically, `testTapOnScrimDismissesSheetAndPopsBackStack`) the entry added would only be `INITIALIZED`. All is fine in the sample. Any ideas what I'm missing? Do we even need to filter?

jossiwolf:
@jbw0033 `testSheetShownAfterNavControllerRestoresState` also seems to fail now - looks like the sheet state doesn't get moved to Hidden. Do you have an idea?

sample/src/main/java/com/google/accompanist/sample/navigation/material/BottomSheetNavSample.kt
```diff
@@ -101,7 +116,17 @@ private fun HomeScreen(showSheet: () -> Unit, showFeed: () -> Unit) {
 
 @Composable
 private fun BottomSheet(showFeed: () -> Unit, showAnotherSheet: () -> Unit, arg: String) {
+    var items by remember { mutableStateOf(2) }
```
jossiwolf, line 119: Will undo the sample changes before merging :)

jbw0033: ACK

jbw0033: Is it worth having multiple samples to test specific cases?

navigation-material/src/main/java/com/google/accompanist/navigation/material/BottomSheetNavigator.kt
```diff
@@ -165,63 +177,78 @@ public class BottomSheetNavigator(
...
-                // Sheet dismissal can be started through popBackStack in which case we have a
-                // transition that we'll want to complete
-                if (transitionsInProgressEntries.contains(backStackEntry)) {
-                    state.markTransitionComplete(backStackEntry)
-                } else {
-                    state.pop(popUpTo = backStackEntry, saveState = false)
+            // Mark all of the entries' transitions as complete, except for the entry we are
+            // currently displaying because it will have its transition completed when the sheet's
+            // animation has completed
+            DisposableEffect(transitionsInProgressEntries) {
```
jbw0033, line 211: Can we still change the system so that we always do the show/hide so that we get callbacks in all cases. And then the `onSheetShown` and `onSheetDismissed` callbacks can just do:

```
transitionsInProgressEntries.forEach {
    state.markTransitionComplete(it)
}
```

And this big of code should no longer be needed I believe.

jossiwolf: Done

gradle/libs.versions.toml
```diff
@@ -1,9 +1,9 @@
 [versions]
 
-compose = "1.3.1"
+compose = "1.4.0-SNAPSHOT"
```
jossiwolf, line 3: Will revert this change after the next release before merging

sample/src/main/java/com/google/accompanist/sample/navigation/material/BottomSheetNavSample.kt
jossiwolf: Not specifically samples I think, but I expanded tests to cover this (`testSheetContentSizeChangeDuringAnimation_opensSheet_shortSheetToShortSheet`, `testSheetContentSizeChangeDuringAnimation_opensSheet_shortSheetToTallSheet`, `testSheetContentSizeChangeDuringAnimation_opensSheet_tallSheetToTallSheet`, `testSheetContentSizeChangeDuringAnimation_opensSheet_tallSheetToShortSheet`)

navigation-material/src/androidTest/java/com/google/accompanist/navigation.material/BottomSheetNavigatorTest.kt
```diff
@@ -273,9 +296,6 @@ internal class BottomSheetNavigatorTest {
 
         assertThat(navigatorState.transitionsInProgress.value)
             .doesNotContain(backStackEntry1)
-
-        assertThat(navigatorState.transitionsInProgress.value)
```
ianhanniballake, line 277: Is there a reason this was removed? I would think we still want to ensure after `awaitIdle` completes that `backStackEntry2` has been removed from the transitioning set and moved all the way to `RESUMED`.

navigation-material/src/androidTest/java/com/google/accompanist/navigation.material/BottomSheetNavigatorTest.kt
```diff
@@ -368,6 +388,379 @@ internal class BottomSheetNavigatorTest {
...
+        val animationDuration = 2000
+        val animationSpec = tween<Float>(animationDuration)
+        lateinit var navigator: BottomSheetNavigator
+        lateinit var navController: NavHostController
+        var height: Dp by mutableStateOf(20.dp)
+
+        composeTestRule.setContent {
+            navigator = rememberBottomSheetNavigator(animationSpec)
+            navController = rememberNavController(navigator)
+            navController.currentBackStackEntryAsState().value
```
ianhanniballake, line 402: Why do we need to read the value here? This should work whether or not we add extra recompositions, right?

navigation-material/src/main/java/com/google/accompanist/navigation/material/BottomSheetNavigator.kt
```diff
@@ -132,9 +141,9 @@ public fun rememberBottomSheetNavigator(
 @ExperimentalMaterialNavigationApi
 @OptIn(ExperimentalMaterialApi::class)
 @Navigator.Name("BottomSheetNavigator")
-public class BottomSheetNavigator(
+class BottomSheetNavigator internal constructor(
```
ianhanniballake, line 144: Can you explain why this is in an `internal` constructor now? If this is intentional, make sure to add a comment stating that should only be constructed via `rememberBottomSheetNavigator` to the KDocs if that is indeed the only entry point supported.

navigation-material/src/main/java/com/google/accompanist/navigation/material/BottomSheetNavigator.kt
```diff
@@ -165,63 +174,71 @@ public class BottomSheetNavigator(
...
             }
-        )
-    }
+
+            SheetContentHost(
+                backStackEntry = retainedEntry,
+                sheetState = sheetState,
+                saveableStateHolder = saveableStateHolder,
+                onSheetShown = {
+                    transitionsInProgressEntries.forEach(state::markTransitionComplete)
```
ianhanniballake, line 214: How does this work when moving between multiple screens (i.e., from Sheet A then `navigate(SheetB)`, `navigate(SheetC)` - do we move directly to `SheetC` (in which case this makes sense - finishing the show of `SheetC` should mark both `SheetB` and `SheetC` as completed) or do we see `SheetB`, then subsequently see `SheetC`?

navigation-material/src/androidTest/java/com/google/accompanist/navigation.material/BottomSheetNavigatorTest.kt
```diff
@@ -368,6 +388,379 @@ internal class BottomSheetNavigatorTest {
...
+        }
+
+        composeTestRule.mainClock.autoAdvance = true
+
+        composeTestRule.waitForIdle()
+        composeTestRule.runOnUiThread {
+            navController.navigate("sheet")
+        }
+        composeTestRule.mainClock.advanceTimeBy(100)
+        height = (composeTestRule.onRoot().getUnclippedBoundsInRoot().height) / 3
```
ianhanniballake, line 422: Can we verify that the sheet is in `transitionsInProgress` here while the sheet is being opened?

Similarly, it would be good to verify that it has been removed from that set after the animation finishes, here and in the other tests as well.

navigation-material/src/androidTest/java/com/google/accompanist/navigation.material/BottomSheetNavigatorTest.kt
jossiwolf: Yep, good catch!

navigation-material/src/androidTest/java/com/google/accompanist/navigation.material/BottomSheetNavigatorTest.kt
jossiwolf: Done

navigation-material/src/androidTest/java/com/google/accompanist/navigation.material/BottomSheetNavigatorTest.kt
jossiwolf: Yep, not needed. Done

navigation-material/src/main/java/com/google/accompanist/navigation/material/BottomSheetNavigator.kt
jossiwolf: Good call, will point it out and softly deprecate it for now. Folks are constructing navigators with custom sheet states which is not something we want to support, and `skipHalfExpanded` support will likely require changing the constructor so I'd like to not have to maintain binary compatibility for this in the future.

navigation-material/src/androidTest/java/com/google/accompanist/navigation.material/BottomSheetNavigatorTest.kt
jossiwolf: Ack, not sure why I removed it.

ianhanniballake (approved):

navigation-material/src/main/java/com/google/accompanist/navigation/material/BottomSheetNavigator.kt
jossiwolf: If we're being consistent with our UX we'd probably want to first see `SheetB` and then `SheetC`. I'm going to leave addressing this for a follow-up PR, I believe this behavior is okay for now.

navigation-material/src/main/java/com/google/accompanist/navigation/material/BottomSheetNavigator.kt
jossiwolf: Decided not to deprecate it for the time being. I saw some folks pass a sheet state with `skipHalfExpanded = true` which is a valid use case imo

jbw0033 (approved):

PR #1470, opened 2023-01-10
[Pager] Migrate from Snapper to SnapFlingBehavior; Support Snap on Mouse Scroll

cotysaxman:
- Removes dependency on Snapper in favor of SnapFlingBehavior, as per Snapper's deprecation.

- Adds support for snapping when scrolling with a mouse (in other words, fixes scrolling on desktop with Compose Multiplatform).

- Fixes a small compiler warning in a modified file.

Tested the Pager code in isolation in my Compose Multiplatform project (MacOS + Android), and tested the Android sample.

Background: I needed to use Pager in a Compose Multiplatform project, and made these two changes in order to accommodate that. I'm not sure what else is needed for Accompanist to support compose-jb, but this fixes Pager.

### DIFF
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -33,8 +33,6 @@ compose-material-iconsext = { module = "androidx.compose.material:material-icons
 compose-material3-material3 = { module = "androidx.compose.material3:material3", version.ref = "composeMaterial3" }
 compose-animation-animation = { module = "androidx.compose.animation:animation", version.ref = "compose" }
 
-snapper = "dev.chrisbanes.snapper:snapper:0.2.2"
-
 android-gradlePlugin = { module = "com.android.tools.build:gradle", version.ref = "gradlePlugin" }
 gradleMavenPublishPlugin = "com.vanniktech:gradle-maven-publish-plugin:0.17.0"
 metalavaGradle = "me.tylerbwong.gradle:metalava-gradle:0.1.9"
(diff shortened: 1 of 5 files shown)


Levi-Moreira (commented):
Hi @cotysaxman I had a look in your changes and I think that the best approach for you to have both your additional mouse support and SnapFlingBehavior would be to adopt foundation's Pager API instead of accompanist's. In the future we indent to move all development of Pager there and that would guaranteed the best support for you. Looking into your changes it seems that all you need is publicly available in PagerState.

cotysaxman:
@Levi-Moreira 
Thank you for reviewing!

The full context leading up to my pull request is that I needed a Date Picker for a Compose Desktop application. This led me to adopting an open-source Material UI Calendar View meant for Android, which depended on Accompanist. SnapFlingBehavior being incompatible with mouse-based click-and-drag seems to be a known blocker for Compose Multiplatform compatibility (I believe it was mentioned on a Snapper issue before deprecation). 

For my own uses, I cloned only Pager and PagerState, and developed these changes in isolation. SnapFlingBehavior is available on multiplatform with Compose 1.3.0+ (currently available in RC), so a local copy of Pager/PagerState along with compose-jb 1.3.0-RC was my best solution.

Adding this mouse solution to Foundation's Pager seems like a great idea, but it will only help multiplatform users when Compose 1.4.x is made available for compose-jb.

I believe Compose on Android should want mouse support, so if you think Foundation Pager is a better target for these changes I can test them against that target when I have the chance.

bentrengrove:
Thank you for this. I'm going to close it as we are currently in the process of deprecating Pager in accompanist and are not accepting any new PRs for it.

Issue #1422, opened 2022-11-29
[Pager] Zoomable Content In Pager


wakaztahir:
**Description**
Not a bug but mostly Horizontal pager is used to display images and I want them to be zoomable individually inside the horizontal pager which led me to ~research and I found this solution

https://stackoverflow.com/questions/71156016/how-to-use-detecttransformgestures-but-not-consuming-all-pointer-event

which is like a hack to make pager zoomable with each image in it. The problem is some solutions cut some of the image and some mess with the way pager scrolls and make it worse.

I would love it If you guys could provide a proper solution on how to use pager with zooming. To disable scrolling a state is being used and some are using screen width to prevent image from getting out of the bounds , I think these solutions could be improved , Could you please provide a solution for this.

wakaztahir:
Not stale

wakaztahir:
Alo Hamora

bentrengrove:
Accompanist Pager has now been deprecated as we have upstreamed it to the main Compose library and so I am closing this request sorry. We won't be updating the documentation here but you can file a request for this at [goo.gle/compose-feedback](https://goo.gle/compose-feedback) if it's still needed when using the Compose Foundation Pager.

PR #1471, opened 2023-01-11
[Pager] Add sample for zoomable content in HorizontalPager

fornewid:
- #1422
- Add sample code that shows zoomable content in pager using Pager's `userScrollEnabled` slot.
   (There are no changes to the Pager API.)

https://user-images.githubusercontent.com/3405740/211869459-0f640486-00a0-4bac-87ac-977486e52e69.mp4

### DIFF
--- a/sample/src/main/AndroidManifest.xml
+++ b/sample/src/main/AndroidManifest.xml
@@ -192,6 +192,16 @@
         </activity>
 
         <activity
+            android:name=".pager.HorizontalPagerWithZoomableContentSample"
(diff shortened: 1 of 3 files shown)


andkulikov:
Hello. Thanks for contributing. We recently added a new Pager components right into compose.foundation library and are now recommend using it instead of Accompanist. Because of that we do not want to add new samples here, however we will make sure to add a sample with zoomable content for the new Pager. Thanks

fornewid:
Thanks for answering!
I'm exciting for new Pager composable in compose.foundation. :)

Issue #1115, opened 2022-04-02
[DrawablePainter] rememberDrawablePainter drops pixel density info of bitmap drawable resource


xiaozhikang0916:
## Describe the bug

When using `rememberDrawablePainter` to show a bitmap drawable resource, which does not fit current screen density, it will show the drawable with a wrong scale.

## To Reproduce

1. Put a bitmap resource in a specified density directory, e.g. drawable-hdpi
2. Use `rememberDrawablePainter` to show that drawable
3. Run the app in a device with xx-h density
4. The drawable is shown as 0.5 scaled

I push a sample page in [my fork repo](https://github.com/xiaozhikang0916/accompanist/tree/sample/drawable-density), branch `sample/drawable-density`, shown in the first sample item `Drawable Painter: Density`

## Expected behavior

The drawable should be displayed with correctly scaled

## Screenshots?

![image](https://user-images.githubusercontent.com/8273026/161386253-f5ffe9d4-7959-4258-a462-07b93b9e4ab9.png)
With the same drawable, current accompanist version in the left, correct scaled in the right.

## Environment:
 - Android Api level 30
 - Device: samsung SM-G981U
 - Accompanist version: v0.24.4-alpha

## Some ideas?

`DrawablePainter` in accompanist can draw the bitmap drawable correctlly, so we can simply remove the `is BitmapDrawable` branch in `rememberDrawablePainter`.

But draw with `BitmapPainter` may has a better performance (don't know, maybe?), I try scaled the bitmap before the BitmapPainter is created, in a extend funtion `BitmapDrawable.asScaledImageBitmap`. You can check it in my repo linked previously.

xiaozhikang0916:
Any info on this? Is it behaved by designed?

PR #1477, opened 2023-01-13
[DrawablePainter] Fix rememberDrawablePainter for BitmapDrawable

hqzxzwb:
This PR removes special case handling for BitmapDrawable. The previous handling was problematic in several ways:
1. bitmap in BitmapDrawable can be null. This was not handled.
2. BitmapDrawable handles density scaling internally. The density information was dropped in previous code, and the returned BitmapPainter would have wrong intrinsic size if specified density for drawable resource was different from device density, which is a common case in resource density fallback.
3. Mutation of BitmapDrawable after creating the Painter is ignored. But mutation of drawables with non-BitmapDrawable types are reflected in the returned DrawablePainter. This behavior difference seems not appropriate to me.

The ColorDrawable special case suffers from 3rd point but not 1st or 2nd point. If removal for ColorDrawable special case is wanted, I'd be glad to do that job too.

### DIFF
--- a/drawablepainter/src/main/java/com/google/accompanist/drawablepainter/DrawablePainter.kt
+++ b/drawablepainter/src/main/java/com/google/accompanist/drawablepainter/DrawablePainter.kt
@@ -17,7 +17,6 @@
 package com.google.accompanist.drawablepainter
 
 import android.graphics.drawable.Animatable
-import android.graphics.drawable.BitmapDrawable
 import android.graphics.drawable.ColorDrawable
 import android.graphics.drawable.Drawable
 import android.os.Build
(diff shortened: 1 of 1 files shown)


hqzxzwb:
And this fixes https://github.com/google/accompanist/issues/1115 .

hqzxzwb:
Hi @bentrengrove , sorry for the disturb if any. But it has been a month since this PR was created. Is there anything I can do to get it reviewed?

bentrengrove:
Yes, I have been chasing the person who originally added this code for you. Still trying!

nadewad (approved):

bentrengrove (approved):

PR #1483, opened 2023-01-17
[Web] Wrap webview in a viewgroup to workaround rare crash

bentrengrove:
Fixes https://issuetracker.google.com/243567497


IanGClifton (approved):

KevinnZou:
This change may lead to another issue, the nested scroll does not work now. Since the AndroidViewHolder use `view?.isNestedScrollingEnabled` to determine whether the child view support nested scroll. It works well before this change because Webview supports nested scroll but failed now since FrameLayout does not. I am wondering if there is any workaround for that problem.

bentrengrove:
Please open an issue for this and we can investigate

PR #1485, opened 2023-01-18
[Pager Indicators] Update Indicators to also be usable with Foundation Pager

Levi-Moreira:
Allow PagerState from Foundation to be fed into the indicators Libraries.


pager-indicators/src/main/java/com/google/accompanist/pager/PagerIndicator.kt
```diff
@@ -77,6 +80,105 @@ fun HorizontalPagerIndicator(
...
+ * @param indicatorWidth the width of each indicator in [Dp].
+ * @param indicatorHeight the height of each indicator in [Dp]. Defaults to [indicatorWidth].
+ * @param spacing the spacing between each indicator in [Dp].
+ * @param indicatorShape the shape representing each indicator. This defaults to [CircleShape].
+ */
+@OptIn(ExperimentalFoundationApi::class)
+@ExperimentalPagerApi
+@Composable
+fun HorizontalPagerIndicator(
+    pagerState: FoundationPagerState,
```
andkulikov, line 135: it is ok to have such aliases in test files, but in the api and public function it is better to fully spell it out. our user can see it and think that there is a class with such name and be confused

pager-indicators/src/main/java/com/google/accompanist/pager/PagerIndicator.kt
```diff
@@ -77,6 +80,105 @@ fun HorizontalPagerIndicator(
...
+        inactiveColor = inactiveColor,
+        indicatorHeight = indicatorHeight,
+        indicatorWidth = indicatorWidth,
+        spacing = spacing,
+        indicatorShape = indicatorShape
+    )
+}
+
+/**
+ * An horizontally laid out indicator for a [HorizontalPager] or [VerticalPager], representing
```
andkulikov, line 107: I think "n" is not needed in "An"

pager-indicators/src/main/java/com/google/accompanist/pager/PagerIndicator.kt
```diff
@@ -162,6 +269,103 @@ fun VerticalPagerIndicator(
...
+}
+
+/**
+ * An vertically laid out indicator for a [VerticalPager] or [HorizontalPager], representing
+ * the currently active page and total pages drawn using a [Shape].
+ *
+ * This element allows the setting of the [indicatorShape], which defines how the
+ * indicator is visually represented.
+ *
+ * @param pagerState The Compose Foundation PagerState object of your [Pager] to be used to observe
```
andkulikov, line 302: [Pager] is not a public api, shouldn't be in brackets

pager-indicators/src/main/java/com/google/accompanist/pager/PagerIndicator.kt
```diff
@@ -162,6 +269,103 @@ fun VerticalPagerIndicator(
...
+        inactiveColor = inactiveColor,
+        indicatorHeight = indicatorHeight,
+        indicatorWidth = indicatorWidth,
+        spacing = spacing,
+        indicatorShape = indicatorShape
+    )
+}
+
+/**
+ * An vertically laid out indicator for a [VerticalPager] or [HorizontalPager], representing
```
andkulikov, line 296: probably [VerticalPager] or [HorizontalPager] also should be fully spelled out with the package here

pager-indicators/src/main/java/com/google/accompanist/pager/PagerIndicator.kt
andkulikov, line 302: maybe instead of "The Compose Foundation PagerState" we will have a [androidx.compose.foundation.pager.PagerState]

andkulikov (approved):

PR #1500, opened 2023-01-27
[Adaptive] Add FoldAwareColumn component

khalp:
### [Adaptive] Add FoldAwareColumn component
 
Adds current state of [FoldAwareColumn](https://github.com/khalp/fold-aware-column) to Accompanist Adaptive.

I put all of the copied RowColumn logic in the `FoldAwareColumn.kt` file, but I can extract that to helper files if that's preferred.


adaptive/src/main/java/com/google/accompanist/adaptive/FoldAwareColumn.kt
```diff
@@ -0,0 +1,321 @@
...
+                        left = 0f,
+                        top = placeableY.toFloat(),
+                        right = placeable.width.toFloat(),
+                        bottom = (placeableY + placeable.height).toFloat()
+                    )
+                    val absoluteBounds =
+                        relativeBounds.translate(layoutBounds.left, layoutBounds.top)
+
+                    // If placeable overlaps fold, push placeable below
+                    if (foldBoundsPx?.let { absoluteBounds.overlaps(it) } == true &&
```
khalp, line 276: When testing, found that `overlaps` will return false if a placeable has a width of zero (ex: `Spacer` with only height specified) - do we want to change the logic to just check if the bounds overlap vertically, or do we want to keep the current logic?

adaptive/src/main/java/com/google/accompanist/adaptive/FoldAwareColumn.kt
```diff
@@ -0,0 +1,321 @@
...
+        val topPaddingPx = foldPadding.calculateTopPadding().roundToPx()
+        val bottomPaddingPx = foldPadding.calculateBottomPadding().roundToPx()
+
+        fold?.bounds?.toComposeRect()?.let {
+            Rect(
+                top = it.top - topPaddingPx,
+                left = it.left,
+                right = it.right,
+                bottom = it.bottom + bottomPaddingPx
+            )
```
khalp, line 84: Do we prefer using `Rect` /`roundToPx` or `RectF` /`toPx`? Since these bounds are only used to determine if children overlap the fold, we're not limited to int values

adaptive/src/main/java/com/google/accompanist/adaptive/FoldAwareColumn.kt
```diff
@@ -0,0 +1,1247 @@
...
+import com.google.accompanist.adaptive.LayoutOrientation.Vertical
+import kotlin.math.max
+import kotlin.math.min
+import kotlin.math.roundToInt
+import kotlin.math.sign
+
+/**
+ * A simplified version of [Column] that places children in a fold-aware manner.
+ *
+ * @param displayFeatures the list of known display features to automatically avoid
```
uchennafokoye, line 53: I would change the param description to: a list of display features the device currently has, as that appears to be what is passed in. 

We will avoid the ones that are relevant to our design not all of them. In the logic, we only avoid Horizontal Folds. Any other display features are not considered, so the description is not fully accurate. 

I would just say it is a list of display features that the device currently has. 

However, I also think this might be an unnecessary parameter. It should be done within the code logic without requiring the developer to get this information. However, I think for Accompanist we can have this here.

khalp: Great catch, just updated the parameter description!

And that's a good point about whether the parameter is necessary - I'll leave it for now like you said since it matches the `TwoPane` constructor for Accompanist, but we can think about how to remove it going forward.

jossiwolf (commented):
Thanks! Looking great, here's a first round of review (finally!)

adaptive/src/main/java/com/google/accompanist/adaptive/FoldAwareColumn.kt
```diff
@@ -0,0 +1,321 @@
...
+        }
+
+    override fun equals(other: Any?): Boolean {
+        if (this === other) return true
+        if (other !is IgnoreFoldModifier) return false
+        return true
+    }
+
+    override fun hashCode(): Int {
+        return hashCode()
```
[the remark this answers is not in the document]
khalp: I don't think we have to actually call `hashCode()`, but we do need to override the `hashCode` method - is there another int value you would recommend returning?
image

adaptive/src/main/java/com/google/accompanist/adaptive/FoldAwareColumn.kt
```diff
@@ -0,0 +1,321 @@
...
+                right = it.right,
+                bottom = it.bottom + bottomPaddingPx
+            )
+        }
+    }
+
+    Layout(
+        modifier = modifier,
+        measurePolicy = foldAwareColumnMeasurePolicy(
+            verticalArrangement = Arrangement.Top,
```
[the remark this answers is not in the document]
khalp: I think for now we are leaving out weight/arrangement implementations due to complexity, but this is a great idea for future improvements!

adaptive/src/main/java/com/google/accompanist/adaptive/RowColumnImpl.kt
```diff
@@ -0,0 +1,760 @@
...
+import androidx.compose.ui.unit.Dp
+import androidx.compose.ui.unit.LayoutDirection
+import androidx.compose.ui.util.fastForEach
+import com.google.accompanist.adaptive.LayoutOrientation.Horizontal
+import com.google.accompanist.adaptive.LayoutOrientation.Vertical
+import kotlin.math.max
+import kotlin.math.min
+import kotlin.math.roundToInt
+
+internal fun rowColumnMeasurePolicy(
```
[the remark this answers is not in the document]
khalp: Great point, I'll add that!

adaptive/src/main/java/com/google/accompanist/adaptive/FoldAwareColumn.kt
```diff
@@ -0,0 +1,321 @@
...
+import androidx.compose.ui.unit.Constraints
+import androidx.compose.ui.unit.Density
+import androidx.compose.ui.unit.Dp
+import androidx.compose.ui.unit.LayoutDirection
+import androidx.window.layout.DisplayFeature
+import androidx.window.layout.FoldingFeature
+import kotlin.math.roundToInt
+
+/**
+ * A simplified version of [Column] that places children in a fold-aware manner.
```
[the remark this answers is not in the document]
khalp: Yes, just added some more descriptive comments - let me know if that makes more sense!

adaptive/src/sharedTest/kotlin/com/google/accompanist/adaptive/FoldAwareColumnTest.kt
```diff
@@ -0,0 +1,232 @@
...
+    @Test
+    fun layout_bounds_align_with_child_bounds_when_pushed_below_separating_fold() {
+        setUp()
+
+        val layoutBottomPx = composeTestRule.onRoot().fetchSemanticsNode().boundsInWindow.bottom
+
+        assertEquals(layoutBottomPx, secondSpacerBoundsPx.bottom)
+    }
+
+    private fun setUp(
```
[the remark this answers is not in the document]
khalp: Great point! I ended up splitting up the `setUp` function because I think it was trying to do too much in one spot - instead of making it a `@Before` method, we now just call set test content and simulate a fold as necessary.

Does the new format look good?

adaptive/src/main/java/com/google/accompanist/adaptive/FoldAwareColumn.kt
khalp: Agreed - just changed the logic to check only if the top/bottom ranges of the Rects overlap.

This changed some of the logic for calculating the bounds of a Rect in the current window, so let me know if the updates still look ok!

adaptive/src/main/java/com/google/accompanist/adaptive/FoldAwareColumn.kt
```diff
@@ -0,0 +1,383 @@
...
+        ),
+        content = { FoldAwareColumnScopeInstance.content() }
+    )
+}
+
+/**
+ * FoldAwareColumn version of [rowColumnMeasurePolicy] that uses [FoldAwareColumnMeasurementHelper.foldAwarePlaceHelper]
+ * method instead of [RowColumnMeasurementHelper.placeHelper]
+ *
+ * TODO: change from internal to private once metalava issue is solved https://partnerissuetracker.corp.google.com/issues/271539608
```
alexvanyo, line 114:
```suggestion
 * TODO: change from internal to private once metalava issue is solved https://issuetracker.google.com/issues/271539608
```

Small nit here to use the publicly visible issue URL.

adaptive/src/main/java/com/google/accompanist/adaptive/FoldAwareColumn.kt
```diff
@@ -0,0 +1,383 @@
...
+            if (orientation == LayoutOrientation.Horizontal) {
+                layoutWidth = measureResult.mainAxisSize
+                layoutHeight = measureResult.crossAxisSize
+            } else {
+                layoutWidth = measureResult.crossAxisSize
+                layoutHeight = measureResult.mainAxisSize
+            }
+
+            // We only know how much padding is added inside the placement scope, so just add fold height
+            // and height of the largest child when laying out to cover the maximum possible height
```
alexvanyo, line 166: 👍

adaptive/src/main/java/com/google/accompanist/adaptive/FoldAwareColumn.kt
```diff
@@ -0,0 +1,383 @@
...
+
+/**
+ * Copy of original [LayoutCoordinates.boundsInWindow], but without the nonzero dimension check.
+ *
+ * Instead of returning [Rect.Zero] for a layout with zero width/height, this method will still
+ * return a Rect with the layout's bounds.
+ */
+@VisibleForTesting
+internal fun LayoutCoordinates.trueBoundsInWindow(): Rect {
+    this.boundsInWindow()
```
alexvanyo, line 313: I think this line is extraneous?

adaptive/src/main/java/com/google/accompanist/adaptive/RowColumnImpl.kt
```diff
@@ -0,0 +1,675 @@
...
+import androidx.compose.ui.unit.LayoutDirection
+import androidx.compose.ui.util.fastForEach
+import com.google.accompanist.adaptive.LayoutOrientation.Horizontal
+import com.google.accompanist.adaptive.LayoutOrientation.Vertical
+import kotlin.math.max
+import kotlin.math.min
+import kotlin.math.roundToInt
+
+/**
+ * Copied from:
```
alexvanyo, line 40: This is a bit unfortunate from a maintenance standpoint, but I don't think there's much to do about that.

adaptive/src/main/java/com/google/accompanist/adaptive/FoldAwareColumn.kt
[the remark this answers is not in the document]
alexvanyo: I think this is a case where a `0` hashcode would be appropriate. Every `IgnoreFoldModifier` will be equal to each other, so `0` hashcode should be correct?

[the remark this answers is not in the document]
khalp: Yes that makes sense, thanks!

adaptive/src/main/java/com/google/accompanist/adaptive/FoldAwareColumn.kt
khalp: Ah yes good catch!

jossiwolf:
@alexvanyo looks like [`d338347`](https://github.com/google/accompanist/pull/1500/commits/d338347129b5c1ec8ee281039a309446e3746a16) was committed from your personal email which doesn't have the CLA signed. Do you mind signing the CLA or updating the email on the commit?

adaptive/src/sharedTest/kotlin/com/google/accompanist/adaptive/FoldAwareColumnTest.kt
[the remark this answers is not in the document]
jossiwolf: Yep!

jossiwolf (approved):

adaptive/src/main/java/com/google/accompanist/adaptive/ExperimentalAdaptiveApi.kt
```diff
@@ -0,0 +1,21 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.google.accompanist.adaptive
+
+@RequiresOptIn(message = "This APIs are experimental and may change in the future.")
+@Retention(AnnotationRetention.BINARY)
+annotation class ExperimentalAdaptiveApi
```
bentrengrove, line 21: I'm hesitant to add this at all because Accompanist doesn't make binary compat guarantees anyway and the whole repo should be considered experimental. Feedback is that this extra annotation is just confusing

khalp: I see what you mean - I'll discuss with Jossi about potentially removing the annotation!

jossiwolf: I'm not opinionated here. We do have it on most other APIs, but we might not need it since it's a new API. Kristen, let's remove it for now?

jossiwolf (approved):
Yay :)

adaptive/src/main/java/com/google/accompanist/adaptive/ExperimentalAdaptiveApi.kt
khalp: sounds good, just updated!

jossiwolf (approved):
LGTM, thanks Kristen!

bentrengrove (approved):

PR #1528, opened 2023-02-27
[Mdc3Theme] Upgrade MDC to 1.8 and map outline variant

### LINKED ISSUE #1526: [Mdc3Theme] Map outline variant, it is available in Material 1.8
Not sure if already is a ticket for this TODO.

This is available in Material 1.8

https://github.com/google/accompanist/blob/4f29fdb3c3c8c603c31e0e913d8cea7cf67c5c87/themeadapter-material3/src/main/java/com/google/accompanist/themeadapter/material3/Mdc3Theme.kt#L259

nicbell:
- Upgrade to MDC `1.8.0` which has been available since 24 Jan 2023.
- `1.8.0` includes outline variant so we should map it https://github.com/material-components/material-components-android/commit/ac6f13c0bec818568bf413883a24358bfa33dd3a

Fixes #1526


bentrengrove (approved):
Thank you for the contribution!

PR #1540, opened 2023-03-09
Switch to gradle build action for CI

bentrengrove:


arriolac (approved):

.github/workflows/build.yml
```diff
@@ -30,25 +30,18 @@ jobs:
...
+        uses: actions/setup-java@v3
         with:
+          distribution: temurin
           java-version: 11
 
       - name: Decrypt secrets
         run: release/signing-setup.sh ${{ secrets.ENCRYPT_KEY }}
 
-      - name: Generate cache key
-        run: ./checksum.sh checksum.txt
```
alexvanyo, line 42: Do we need `checksum.sh` anymore then? Or are other workflows using it still?

bentrengrove: Other workflows are still using it, I want to test this works first

PR #1554, opened 2023-03-24
Rollback navigation upgrade to latest stable

### LINKED ISSUE #1552: [Navigation Material] Current release 0.30.0 ships navigation-compose alpha
**Description**
Current accompanist stable release 0.30.0 ships the navigation-compose:2.6.0-alpha07. We have an internal policy that we dont want to alpha versions without an important reason. This includes also transitive dependencies.

Was this commit intended and what was the reason?
https://github.com/google/accompanist/commit/c1ac4cb0a1bf4d2e01e7def6a3479204805f48d0

**Steps to reproduce**

**Expected behavior** 
Stable releases only ships stable transitive dependencies or has a valid reason to depend on beta/alpha versions.

**Additional context**

bentrengrove:
Fixes #1552

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -19,7 +19,7 @@ jobs:
     if: "!contains(github.event.head_commit.message, 'skip ci')"
 
     runs-on: ubuntu-latest
-    timeout-minutes: 30
+    timeout-minutes: 45
 
     steps:
       - uses: actions/checkout@v2
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -16,7 +16,7 @@ okhttp = "3.12.13"
 coil = "1.3.2"
 
 androidxtest = "1.4.0"
-androidxnavigation = "2.6.0-alpha07"
+androidxnavigation = "2.5.3"
 androidxWindow = "1.0.0"
 
 [libraries]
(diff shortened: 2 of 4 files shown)


bentrengrove:
Looks like that gradle issue still exists. @jossiwolf, you need 2.6.0 for your work right?

jossiwolf:
@bentrengrove I don't think we need any APIs or features specifically, but there have been quite a few important bugfixes. Tbh I think it's fine to depend on an alpha version in Accompanist, we do the same with other AndroidX dependencies. Do you think we should depend on a stable Nav version instead?

cc @jbw0033

G00fY2:
@jossiwolf I heavily disagree on using transitive unstable dependencies in stable releases. Thats also not the case for any other official library I am aware of (also not AndroidX deps).
The consumers should decide wether the bugfixes are important and justify using the alpha release.

E.g. we use [this check](https://gist.github.com/G00fY2/c852322a18b187f40bb884a85065c57f) (inspired by [this](https://discuss.gradle.org/t/failonversionconflict-only-for-certain-dependencies/29375)  Gradle discussion). Currently Accompanist is the only dependency which violates the check.

JohannesPtaszyk:
Have to agree with @G00fY2. Transitive unstable dependencies in stable releases should be disallowed.

serhii-pokrovskyi:
> We will be rolling back the navigation dependency to the last stable version and doing another Accompanist stable release.

any chances that this will fix?
I think that they are related
#1558

jossiwolf (approved):

PR #1557, opened 2023-03-24
[WebView] Support saving and loading state and custom layout params

### LINKED ISSUE #1178: [WebView] Suggestion for persistent WebView upon configuration change or navigating to/from WebView composable

### LINKED ISSUE #1407: [WebView] pageTitle is not restored when reload
**Description**

After `WebViewNavigator.reload()` is called the `WebViewState.pageTitle` is reset to null with `AccompanistWebViewClient.onPageStarted()` but `AccompanistWebChromeClient.onReceivedTitle()` is not called with `WebView.reload()` so pageTitle stay null.

**Steps to reproduce**

```kotlin
@Composable
fun AccompanistWebSample() {
    val state = rememberWebViewState(url = "https://google.github.io/accompanist/web/")
    val navigator = rememberWebViewNavigator()

    Column {
        TopAppBar(
            title = {
                Text(text = state.pageTitle ?: "")
            },
            actions = {
                IconButton(onClick = { navigator.reload() }) {
                    Icon(imageVector = Icons.Default.Refresh, contentDescription = "reload")
                }
            }
        )

        WebView(
            state = state,
            navigator = navigator,
            modifier = Modifier.fillMaxSize()
        )
    }
}
```

**Expected behavior** 

pageTitle should stay same string when reload.

### LINKED ISSUE #1435: [WebView] Pager height miscalculation when restoring
**Description**

Within a pager, with each page being a webview, the previous webview height will miscalculate to 0.

**Steps to reproduce**

1. Create Pager, with pages of webviews, each using HTML from a string
2. Go forward, how every many you like
3. Go backwards, notice height miscalculation in layout inspector

**Expected behavior** 

Height calculates correctly

**Additional context**

The current solution, is to override onPageFinished and use the following hack
```kotlin
view?.layoutParams = ViewGroup.LayoutParams(
	ViewGroup.LayoutParams.MATCH_PARENT,
	ViewGroup.LayoutParams.WRAP_CONTENT
)
```

### LINKED ISSUE #1563: [WebView] Double page loading during first composition
**Description**
The WebView composable seems to not be taking proper care of _when_ the launched effects run. This causes launched effects to be run multiple times, with same values, causing multiple loads of the initial page.

**Steps to reproduce**
1. Create a screen with the WebView composable that just loads a simple URL:
`WebView(state = WebViewState(WebContent.Url(url = "https://example.com")))`

**Expected behavior** 
1. The URL is only loaded once;
2. The launched effects are executed with right values; 

**Actual behvior**
Please see the logs below. You'll notice a few things:
1. Composition of the WebView starts; - OK
2. Composition of AndroidView and creation of WebView - OK;
3. Since everything has been on main thread, the **launched effect** coroutines start up **only after the WebView is created**. This causes the issue, due to the captured values: both `LaunchedEffects` run now, after the `webView` as been instantiated, so the `webView` is not null inside anymore, even though they were "composed" when the `webView` was `null`.
4. Problem is that: we attach to the 2 flows (both the `navigator.navigationEvents` and the `state.content`) twice and load the URL twice.

**Proposed solution**
Only compose the `LaunchedEffect` if the `webView` is not null _and_ capture & use correct value:
```
    webView?.let { wv ->
        LaunchedEffect(wv, navigator) {
            with(navigator) { wv.handleNavigationEvents() }
        }
    }
```
or at least use the captured value
```
val currentWebView = webView
    LaunchedEffect(currentWebView, navigator) {
        with(navigator) { currentWebView?.handleNavigationEvents() }
    }
```

**Additional context**
Here's are logs with certain events. Please notice that the LaunchedEffects run _after_ the WebView has been initialized, so they run with the same arguments twice. 
```
14:51:33.386 32001-32001 Composing AccompanistWebView
14:51:33.395 32001-32001 Composing AndroidView
14:51:33.398 32001-32001 Creating WebView via Factory...
...
14:51:33.421 32001-32001 Created WebView: android.webkit.WebView{3aeafb5 VFEDHVC.. ......I. 0,0-0,0}
14:51:33.441 32001-32001 WebView+Navigator LaunchedEffect: android.webkit.WebView{3aeafb5 VFEDHVC.. ......ID 0,0-1080,0} io.shortcut.wrappkit.core.webview.WebViewNavigator@a6f533
14:51:33.443 32001-32001 WebView+State LaunchedEffect: android.webkit.WebView{3aeafb5 VFEDHVC.. ......ID 0,0-1080,0} io.shortcut.wrappkit.core.webview.WebViewState@e3e20f0
14:51:33.446 32001-32001 Loading url: Url(url=https://example.com, additionalHttpHeaders={})
14:51:33.454 32001-32001 Composing AccompanistWebView
14:51:33.462 32001-32001 OnProgressChanged: 10
14:51:33.463 32001-32001 WebView+Navigator LaunchedEffect: android.webkit.WebView{3aeafb5 VFEDHVC.. ......ID 0,0-1080,0} io.shortcut.wrappkit.core.webview.WebViewNavigator@a6f533
14:51:33.464 32001-32001 WebView+State LaunchedEffect: android.webkit.WebView{3aeafb5 VFEDHVC.. ......ID 0,0-1080,0} io.shortcut.wrappkit.core.webview.WebViewState@e3e20f0
14:51:33.465 32001-32001 Loading url: Url(url=https://example.com, additionalHttpHeaders={})
14:51:33.472 32001-32001 Composing AndroidView
14:51:33.479 32001-32001 OnPageFinished: https://example.com/
14:51:33.564 32001-32001 OnPageStarted: https://example.com/
14:51:33.568 32001-32001 OnProgressChanged: 100
14:51:33.568 32001-32001 OnProgressChanged: 100
14:51:33.569 32001-32001 OnPageFinished: https://example.com/
```

### LINKED ISSUE #1564: [WebView] Too many null checks in cases where it doesn't make sense
**Description**
There are too many null checks in situations that don't really make sense. For example, the methods in `WebViewClient` and `WebChromeClient` all have nullable arguments, while that is not correct, since they'll never be called with a null web view. This is likely due to missing of nullable annotations on the Java WebView.

### LINKED ISSUE #1573: [WebView] Do not require an URL on state creation
**Description**
Right now, when the WebView state is created it requires a URL string to be passed in. This is fine in most occasions, but sometimes the URL is retrieved asynchronously and might not be known when the state is created. This can easily be worked around by loading "about:blank" first and loading the correct url afterwards.  
However, this workaround adds the "about:blank" page to the back stack, and when pressing the back button we get a blank screen.

I think there should be the option to not load any url at the start, and let the developer load it when it finds it convenient, like the normal view allows us to do.

Regards,
Iscle

bentrengrove:
Enables the ability to save and restore WebView state when leaving and reentering composition. 

In order to restore state you must use the new `rememberSaveableWebViewState()` API that doesn't take an initial URL.
Then you can check if any view state exists before loading a URL in the WebView. This also allows for state to not be restored if required.

```
val webViewState = rememberSaveableWebViewState()
    val navigator = rememberWebViewNavigator()

    LaunchedEffect(navigator) {
        val bundle = webViewState.viewState
        if (bundle == null) {
            // This is the first time load, so load the home page.
            navigator.loadUrl("https://bbc.com")
        }
    }

    WebView(
        state = webViewState,
        navigator = navigator,
        modifier = Modifier.fillMaxSize()
    )
 ```

As part of this refactor, a new API was introduced that allows custom LayoutParams to be passed into the WebView composable. There are now two entry points, one that attempts to size itself based on Modifier constraints and one where you must pass in the appropriate LayoutParams to size the WebView correctly.

Fixes #1178 
Fixes #1407 
Fixes #1435
Fixes #1563
Fixes #1573
Fixes #1564

### CODE REVIEW

File: sample/src/main/java/com/google/accompanist/sample/webview/WebViewSaveStateSample.kt [hunk 1]
```diff
@@ -0,0 +1,114 @@
...
+                }
+            }
+        }
+    }
+}
+
+@Composable
+private fun Home() {
+    val webViewState = rememberSaveableWebViewState()
+    val webViewBundleState = rememberSaveable { mutableStateOf<Bundle?>(null) }
```
alexvanyo, line 82: I think it'd be better if this state was handled internally by the `WebView` wrapper, and then it'd function like other components that have some sort of `rememberSaveable` internal state. A wrapping `key` would be able to reset the state if desired.

web/src/main/java/com/google/accompanist/web/WebView.kt
```diff
@@ -80,45 +88,120 @@ fun WebView(
...
+            with(navigator) {
+                wv.handleNavigationEvents()
             }
         }
-    }
 
-    val currentOnDispose by rememberUpdatedState(onDispose)
+        LaunchedEffect(wv, state) {
+            state.viewState?.let {
+                wv.restoreState(it)
```
alexvanyo, line 181: Is it possible to call this as part of the `AndroidView` factory? As is, I think we will have about a 2-composition delay before this is called:

- First composition, `webView` is `null`
- Second composition, `LaunchedEffect` is called
- After second composition, this line will run

sample/src/main/java/com/google/accompanist/sample/webview/WrappedContentWebViewSample.kt
```diff
@@ -73,24 +73,30 @@ class WrappedContentWebViewSample : ComponentActivity() {
...
- * A sample WebView that is wrapping it's content inside of a bottom sheet.
+ * A sample WebView that is wrapping it's content height.
  * The sheet should be the size of the rendered content and not unbounded.
  */
 @Composable
-fun WebContent() {
+fun WrappingWebContent(
+    body: String,
+    onCreated: (WebView) -> Unit = {},
+    onDispose: (WebView) -> Unit = {}
```
alexvanyo, line 83: Nit: is it still necessary to expose these?

sample/src/main/java/com/google/accompanist/sample/webview/WebViewSaveStateSample.kt
```diff
@@ -0,0 +1,105 @@
...
+    val webViewState = rememberSaveableWebViewState()
+    val navigator = rememberWebViewNavigator()
+
+    LaunchedEffect(navigator) {
+        val bundle = webViewState.viewState
+        if (bundle == null) {
+            // This is the first time load, so load the home page.
+            navigator.loadUrl("https://bbc.com")
+        }
+    }
```
alexvanyo, line 88: This is probably best for a future exploration, but it might be nice if we could do something like

```
val navigator = rememberWebViewNavigator(
    initialUrl = "https://bbc.com"
)
```

Which would only load the URL if there wasn't saved state.

bentrengrove: Yeah I thought that as well but trying to do it reveals a circular dependency between the navigator and state classes which makes me think a full revamp of the API would be needed to implement.

Definitely the direction this is going but this PR has already crept too far

web/src/main/java/com/google/accompanist/web/WebView.kt
bentrengrove: Great idea, the current place felt wrong to me as well

navigation-animation/api/current.api
```diff
@@ -11,13 +11,13 @@ package com.google.accompanist.navigation.animation {
...
...ndroidx.compose.runtime.Composable public static void AnimatedNavHost(androidx.navigation.NavHostController navController, String startDestination, optional androidx.compose.ui.Modifier modifier, optional androidx.compose.ui.Alignment contentAlignment, optional String? route, optional kotlin.jvm.functions.Function1<? super androidx.compose.animation.AnimatedContentScope<androidx.navigation.NavBackStackEntry>,? extends androidx.compose.animation.EnterTransition> enterTransition, optional kotlin.jvm.functions.Function1<? super androidx.compose.animation.AnimatedContentScope<androidx.navigation.NavBackStackEntry>,? extends androidx.compose.animation.ExitTransition> exitTransition, optional kotlin.jvm.functions.Function1<? super androidx.compose.animation.AnimatedContentScope<androidx.navigation.NavBackStackEntry>,? extends androidx.compose.animation.EnterTransition> popEnterTransition, optional kotlin.jvm.functions.Function1<? super androidx.compose.animation.AnimatedContentScope<androidx.navigation.NavBackStackEntry>,? extends androidx.compose.animation.ExitTransition> popExitTransition, kotlin.jvm.functions.Function1<? super androidx.navigation.NavGraphBuilder,kotlin.Unit> builder);
```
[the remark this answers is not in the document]
bentrengrove: I don't have any local changes to those files. This makes me think maybe the CI is broken and not really checking it properly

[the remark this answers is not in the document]
alexvanyo: Huh, that's really weird. The diff also doesn't look like it is a real diff with any changes too.

alexvanyo (approved):
LGTM!

bentrengrove:
It will be in the next release in approximately two weeks time

bentrengrove:
This has just been released to maven in 0.31.1-alpha

R4M5E5:
Is it somehow possible to hoist the `webViewState` and the `navigator` above the NavHost? So for example in the `WebViewSaveStateSample.kt` to call the `rememberSaveableWebViewState()` and the `rememberWebViewNavigator()` just right after the `rememberNavController()` call and then pass down the `webViewState` and the `navigator` to the `Home` composable via arguments?

So something like:
```kotlin
val navController = rememberNavController()
val webViewState = rememberSaveableWebViewState()
val navigator = rememberWebViewNavigator()

LaunchedEffect(navigator) {
   val bundle = webViewState.viewState
   if (bundle == null) {
      // This is the first time load, so load the home page.
      navigator.loadUrl("https://bbc.com")
   }
}


NavHost(navController = navController, startDestination = "home") {
   composable("home") {
      Home(webViewState, navigator)
    }
   composable("detail") {
       Detail()
    }
}                   
```

Because for me this doesn't work. With this approach the webview is not keeping the state after navigating. Any idea how to make this work?

rafakob:
Is there a way to remember WebView state which is in a LazyColumn item? I'm testing the `rememberSaveableWebViewState` but when I scroll away from that item and go back to it - WebView refreshes the page.

YugeCse:
I put a webview into a ScrollableControlView,  Then I click one button, and the page navigates to annother page. Then, I back to the webview page, the webview refreshes the page. It loss its scroll position.

Box(Modifier.fillMaxSize().verticalScroll(rememberScrollState())){
   Button(Modifier.click(gotoAnnotherPage)){  Text('NEXT PAGE') }
   WebView(Modifier.fillMaxWidth().wrapContentHeight())
   Text('Hello , this is a test!')
}

Is there any way to remember the scroll position? Thank you!

nickboris:
Thank you @bentrengrove for adding this functionality, I think it's a great step in making webviews easier to manage with compose. I wanted to second the use case that @rafakob mentioned above. I have a small webview element being displayed in a lazy column that I can't revert to a non-recycling column due to performance concerns, and the reloads are quite jarring from a UX perspective. Will be following this thread to see if there are any updates/plans to help support this use case in the future.

PR #1561, opened 2023-03-27
[Pager] Fix typo in pager.md

barnard-b:
Removes a space which caused the markdown to render incorrectly.


bentrengrove (approved):
Thank you!

PR #1565, opened 2023-03-31
[Pager] Fix formatting and improve pager documentation

MaximillianLeonov:


bentrengrove:
This has already been fixed in #1561 but the docs only get generated with a new release so hasn't shown up yet.

MaximillianLeonov:
@bentrengrove,

Hi, Ben. Are these issues the same? You can see the issue below in the screenshot for which I provided the fix, which is caused by the missing four spaces indentation. Correct me if I am wrong, please.

**Actual:**
image

**Expected:**
image

bentrengrove:
I believe it is the same issue, if it doesn't work I will come back and merge yours

bentrengrove (approved):
Looks like it is still broken so will merge this in. Thank you for the fix!

Issue #1578, opened 2023-04-13
[Suggestion] Migrate gradle files from Groovy to Kotlin DSL


oas004:
Hey!

First of all, thanks for a great library! 

This is just a suggestion and I wanted to know what you guys thought about it. Feel free to close it, if it does not make sense.

I noticed that all the gradle files are in Groovy, what do you guys think about converting it to Kotlin DSL? Is there a reason for not migrating or any downside to migrating? :) 

I would love to help out and contribute if this is okay :)

One suggested approach to this as we did when migrating in the [horologist project](https://github.com/google/horologist) is the following:

1. Go module by module prepare the Groovy scripts following the steps described [here](https://docs.gradle.org/current/userguide/migrating_from_groovy_to_kotlin_dsl.html#prepare_your_groovy_scripts). Then make a PR for each module to keep it easy to review.
2.  Do the actual conversion of the file to `.kts` and check that it compiles and that everything works as expected. Then make a PR for each module, to make it easy to review.

One tool that might help migrate is [this](https://github.com/bernaferrari/GradleKotlinConverter) script to convert Groovy to Kotlin DSL.  I have not tried it myself though.

bentrengrove:
Thanks for the suggestion. We would definitely accept this as a contribution if you are open to doing it!

I agree with your plan to do it module by module.

oas004:
I would love to! :) I will try to prioritize this :)

bentrengrove:
I guess stalebot can't see all the PRs

oas004:
Just to give a status update here is that currently  22/27 groovy files are not in Kotlin DSL :) We have PRs up on the rest except the project grades. I will do that when everything else is in. When that is done my thoughts where to set the target and source java versions to 11 and to update the plugins to using version catalog correctly :) Then I think we can also bump the AGP version as of #1617

bentrengrove:
Thank you for all your effort!

PR #1579, opened 2023-04-16
[Adaptive] Migrate gradle module til Kotlin DSL

oas004:
This is the first PR in a series to migrate the gradle files to use Kotlin DSL instead of Groovy.  Since having Kotlin DSL and Groovy in the same project works fine, I thought migrating module by module was okay.

The related issue is #1578.

I have split up the PR into two commits. The first one prepares the file to be migrated by using steps described in [these docs](https://docs.gradle.org/current/userguide/migrating_from_groovy_to_kotlin_dsl.html#prepare_your_groovy_scripts).  The second one is where I actually convert the file to `.kts` file and updates the rest of the file to be compatible with Kotlin DSL.

I have added a line to suppress unstable API usage, since a lot of the APIs here are marked incubating. What do you guys think about that?

### DIFF
--- a/adaptive/build.gradle.kts
+++ b/adaptive/build.gradle.kts
@@ -1,5 +1,5 @@
 /*
- * Copyright 2022 The Android Open Source Project
+ * Copyright 2023 The Android Open Source Project
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -16,10 +16,11 @@
 @file:Suppress("UnstableApiUsage")
 
 plugins {
-    id("com.android.library")
-    id("kotlin-android")
-    id("org.jetbrains.dokka")
-    id("me.tylerbwong.gradle.metalava")
+    id(libs.plugins.android.library.get().pluginId)
+    id(libs.plugins.android.kotlin.get().pluginId)
+    id(libs.plugins.jetbrains.dokka.get().pluginId)
+    id(libs.plugins.gradle.metalava.get().pluginId)
+    id(libs.plugins.vanniktech.maven.publish.get().pluginId)
 }
 
 kotlin {
@@ -39,8 +40,8 @@ android {
     }
 
     compileOptions {
-        sourceCompatibility = JavaVersion.VERSION_1_8
-        targetCompatibility = JavaVersion.VERSION_1_8
+        sourceCompatibility = JavaVersion.VERSION_11
+        targetCompatibility = JavaVersion.VERSION_11
     }
 
     buildFeatures {
@@ -130,5 +131,3 @@ dependencies {
 
     testImplementation(libs.robolectric)
 }
-
-apply(plugin = "com.vanniktech.maven.publish")
--- a/adaptive/build.gradle.kts
+++ b/adaptive/build.gradle.kts
@@ -13,6 +13,7 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
+@file:Suppress("UnstableApiUsage")
 
 plugins {
     id("com.android.library")
@@ -28,12 +29,12 @@ kotlin {
 android {
     namespace = "com.google.accompanist.adaptive"
 
-    compileSdkVersion = 33
+    compileSdk = 33
 
     defaultConfig {
-        minSdkVersion 21
+        minSdk = 21
         // targetSdkVersion has no effect for libraries. This is only used for the test APK
-        targetSdkVersion 33
+        targetSdk = 33
         testInstrumentationRunner = "androidx.test.runner.AndroidJUnitRunner"
     }
 
@@ -51,9 +52,9 @@ android {
         kotlinCompilerExtensionVersion = libs.versions.composeCompiler.get()
     }
 
-    lintOptions {
+    lint {
         textReport = true
-        textOutput "stdout"
+        textOutput = File("stdout")
         // We run a full lint analysis as build part in CI, so skip vital checks for assemble tasks
         checkReleaseBuilds = false
     }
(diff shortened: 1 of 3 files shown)


adaptive/build.gradle.kts
```diff
@@ -0,0 +1,134 @@
...
+        compose = true
+    }
+
+    composeOptions {
+        kotlinCompilerExtensionVersion = libs.versions.composeCompiler.get()
+    }
+
+    lint {
+        textReport = true
+        textOutput = File("stdout")
```
oas004, line 57: I'm not quite sure about this, should this be set to the build folder maybe? 🤔

bentrengrove:
Thank you for these! Sorry I have been slow to review, I have been OOO this week. Just catching up but should be able to get to them soon.

oas004:
> Thank you for these! Sorry I have been slow to review, I have been OOO this week. Just catching up but should be able to get to them soon.

No worries at all! Take your time! :)

adaptive/build.gradle.kts
```diff
@@ -0,0 +1,134 @@
...
+
+    defaultConfig {
+        minSdk = 21
+        // targetSdkVersion has no effect for libraries. This is only used for the test APK
+        targetSdk = 33
+        testInstrumentationRunner = "androidx.test.runner.AndroidJUnitRunner"
+    }
+
+    compileOptions {
+        sourceCompatibility = JavaVersion.VERSION_1_8
```
borisf, line 42: Let's make it JavaVersion.Version_11, let me know if there is something we stuck on Java 8

adaptive/build.gradle.kts
```diff
@@ -0,0 +1,134 @@
...
+        targetCompatibility = JavaVersion.VERSION_1_8
```
borisf, line 43: Same, lets make Java 11

adaptive/build.gradle.kts
```diff
@@ -0,0 +1,134 @@
...
+    androidTestImplementation(libs.androidx.test.runner)
+    testImplementation(libs.androidx.test.runner)
+
+    androidTestImplementation(libs.androidx.window.testing)
+    testImplementation(libs.androidx.window.testing)
+
+    testImplementation(libs.robolectric)
+}
+
+apply(plugin = "com.vanniktech.maven.publish")
```
borisf, line 134: Lets use the plugins block https://developer.android.com/build/migrate-to-kotlin-dsl#migrate-buildscript

adaptive/build.gradle.kts
```diff
@@ -0,0 +1,134 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+@file:Suppress("UnstableApiUsage")
+
+plugins {
+    id("com.android.library")
```
borisf, line 19: Let's add plugins to catalog as well - https://github.com/google/accompanist/blob/main/gradle/libs.versions.toml, here is how https://developer.android.com/build/migrate-to-catalogs#migrate-plugins

adaptive/build.gradle.kts
```diff
@@ -0,0 +1,134 @@
+/*
+ * Copyright 2022 The Android Open Source Project
```
borisf, line 2: Shouldbn't it be 2023

borisf (changes requested):
Thanks a lot @oas004

adaptive/build.gradle.kts
oas004: I tried this, but since we are on agp version 7.4.1 I think we have this issue (https://github.com/gradle/gradle/issues/17968) since we have to maintain the class path version as well. I can do a workaround like this
`id(libs.plugins.android.library.get().pluginId)` would that be okay?

borisf: Yes please, as we continue to enjoy the catalogs, and we can change this code once the Gradle bug is fixed

borisf: On the same thought we should move to AGP 8.0 which is stable

adaptive/build.gradle.kts
oas004: Updated in 6bbdef1 :) Thanks!

adaptive/build.gradle.kts
oas004: Added this in 6bbdef1 :) Moving to AGP 8.0, should that maybe be in a separate PR if these are squashed in case it breaks something? Wdyt? :)

borisf: @bentrengrove call :) (my opinion makes sense to do in a separate PR)

adaptive/build.gradle.kts
oas004: Good idea! I think this went fine :) Upgraded it in 6bbdef1 :)

borisf (commented):
LGTM, @bentrengrove

adaptive/build.gradle.kts
bentrengrove: Yes, let's upgrade. Separate PR please, if you have time to do it that would be really great to unblock this. But if not, definitely no problem. It will just take me ~1 week to get to it.

bentrengrove (approved):

adaptive/build.gradle.kts
oas004: I can take a look into upgrading to AGP 8.0 parallel to this :)

oas004: Added an attempt on this here: https://github.com/google/accompanist/pull/1590

bentrengrove: Does this now solve the need for using pluginId?

oas004: I tried to rebase this locally and upgrading does not fix it. I think the fix is to migrate all the plugins at classpath to use the alias and then update this to use alias as well. The reason I think this is that the sync error is that it already has this plugin on the classpath with an unknown version, so the compatibility cannot be checked.

I can fix it if I add the `pluginManagment` block with `repositories` block to `settings.gradle` and then remove the classpath decleration and move the plugins to the `plugins` block using the `alias` there as well in the project gradle file. Should I do that here? Or wait until I migrate the project gradle and settings gradle file to Kotlin DSL? Or I can do it in a seperate PR if you want to seperate these two things.

bentrengrove: Let's wait until you migrate those other files and go with the current solution for now. I will merge this one in!

bentrengrove:
This just needs rebasing I think. Can you please rebase and reopen a PR when ready?

oas004:
Ah yes, I'm not quite sure why this build before we bumped AGP 8.0.0, but I had to revert the Java target version in this module to 1.8 to get it to compile. I think this is because we are setting 1.8 in project gradle. Maybe it makes sense to migrate everything there to 11 in bulk? Sorry about this, I should have rebased it before and checked.
Reopened a PR here https://github.com/google/accompanist/pull/1593

PR #1581, opened 2023-04-16
[DrawablePainter] Migrate drawablepainter module gradle file to Kotlin DSL

oas004:

Migrating the `drawablepainter` module to Kotlin DSL.

I'm a bit unsure about the lint lambda textOutout. Maybe that should be in the build file or test result folder or something. I will update it to what we decide in https://github.com/google/accompanist/pull/1579#discussion_r1167980394.

### DIFF
--- a/drawablepainter/build.gradle
+++ b/drawablepainter/build.gradle
@@ -15,10 +15,10 @@
  */
 
 plugins {
-    id 'com.android.library'
-    id 'kotlin-android'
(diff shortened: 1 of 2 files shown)


oas004:
Will wait on #1579. I moved some of the gradle plugin to toml, so will rebase this when that is merged so that I can use that :) 
Will apply the same comments as was mentioned in #1579

bentrengrove (approved):

PR #1590, opened 2023-04-26
[All] Update agp to 8.0.0 and distribution to 8.1

oas004:
Following up on https://github.com/google/accompanist/pull/1579#discussion_r1177240398

Update AGP to 8.0.0 and distribution to 8.1.  Making it a draft to see if this works through the pipeline. I'm unsure if I have to upgrade the Java version here as well.

To do this I went through the documentation steps here (https://docs.gradle.org/current/userguide/upgrading_version_7.html#changes_7.4).


bentrengrove (approved):
This looks great to me. Thank you so much!

PR #1593, opened 2023-04-28
[Adaptive] Migrate gradle module Adaptive to Kotlin DSL

oas004:
Referencing https://github.com/google/accompanist/pull/1579#issuecomment-1526576155 and https://github.com/google/accompanist/pull/1592

This was reverted in the latest commit because it did not build. Rebased and changed the Java target compatibility version as I could not get it to build with 11. I think we can do a build update to Java 11 parallel to these migrations.


adaptive/build.gradle.kts
```diff
@@ -0,0 +1,133 @@
...
+    defaultConfig {
+        minSdk = 21
+        // targetSdkVersion has no effect for libraries. This is only used for the test APK
+        targetSdk = 33
+        testInstrumentationRunner = "androidx.test.runner.AndroidJUnitRunner"
+    }
+
+    compileOptions {
+        sourceCompatibility = JavaVersion.VERSION_1_8
+        targetCompatibility = JavaVersion.VERSION_1_8
```
oas004, line 44: I had to revert this from #1579 Cc. @borisf and @bentrengrove

bentrengrove (approved):
Let's move them all to Java 11 in one go in parallel or at the end

PR #1594, opened 2023-04-28
[All] Bump vanniktechPublish plugin version

oas004:
After bumping to AGP 8.0 I think the publish task was broken.  Sorry about that. I could reproduce it locally with `gradle publishToMavenLocal`. Bumping this version fixed the issue locally. Would be nice to test this thoroughly before merging since it minor version bump. Is there a better way to test this than running `gradle publishToMavenLocal`?


bentrengrove (approved):
Thank you! It's fine to test this in main on merge, it just uploads a snapshot anyway so it won't actually roll anything out

oas004:
Nice! Seems like that made it work on main again at least :)

PR #1599, opened 2023-05-01
[Internal-testutils] Migrate gradle files to Kotlin DSL

oas004:

Migrating the `:internal-testutils` module to Kotlin DSL.

## Testing

The testing I have done for this is to check that it syncs and builds the sample application. Let me know if there is any other test than what the CI runs that you want me to run as well.


bentrengrove (approved):
Thank you for all of these! It is super appreciated!

PR #1614, opened 2023-05-11
[Permissions] Migrate build gradle to Kotlin DSL

oas004:

Migrating the `:permissions` module to Kotlin DSL.

### DIFF
--- a/permissions/build.gradle.kts
+++ b/permissions/build.gradle.kts
@@ -13,12 +13,14 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
+@file:Suppress("UnstableApiUsage")
 
 plugins {
-    id("com.android.library")
-    id("kotlin-android")
-    id("org.jetbrains.dokka")
-    id("me.tylerbwong.gradle.metalava")
+    id(libs.plugins.android.library.get().pluginId)
+    id(libs.plugins.android.kotlin.get().pluginId)
+    id(libs.plugins.jetbrains.dokka.get().pluginId)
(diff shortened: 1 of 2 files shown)


permissions/build.gradle.kts
```diff
@@ -0,0 +1,119 @@
...
+        minSdk = 21
+        // targetSdkVersion has no effect for libraries. This is only used for the test APK
+        targetSdk = 33
+
+        testInstrumentationRunner = "androidx.test.runner.AndroidJUnitRunner"
+
+        // The following argument makes the Android Test Orchestrator run its
+        // "pm clear" command after each test invocation. This command ensures
+        // that the app's state is completely cleared between tests.
+        testInstrumentationRunnerArguments["clearPackageData"] = "true"
```
oas004, line 45: I'm not 100% sure about this from Groovy, is there a good way to test this locally?

bentrengrove: @manuelvicnt any ideas?

manuelvicnt: IIRC, tests should fail without this, so that's a good way to check. If it doesn't fail, this isn't doing anything and we can remove it

oas004: Sorry for late reply, it was constitution day in Norway, so was out celebrating 🥳 Thank you for great tip! I tried locally to run the tests without this line and then 14 tests failed in the permissions module. So I think keeping this like this should be correct.

If you want, I can run this on CI as a separate commit, so we can check that these tests that failed locally are the same tests that you would expect to fail? :)

bentrengrove (approved):

Issue #1616, opened 2023-05-13
[Suggestion] Migrate AGP from 8.0.0 to 8.0.1


stevan-milovanovic:
Hey all, first of all, thanks for a great library!

I have a suggestion regarding AGP version and I wanted to know what you guys thought about it. Feel free to close it, if it does not make sense.

I noticed that project uses AGP 8.0.0 and version 8.0.1 is released.

I would love to help out and contribute if this is okay :)

stevan-milovanovic:
Then we can resolve this TODO from the project:

`// TODO: Remove when fixed https://github.com/gradle/gradle/issues/24037`

PR #1617, opened 2023-05-13
Migrate AGP from 8.0.0 to 8.0.1

stevan-milovanovic:
Following up https://github.com/google/accompanist/issues/1616

Migrating AGP from 8.0.0 to 8.0.1 and resolving TODO: Remove when fixed https://github.com/gradle/gradle/issues/24037

### DIFF
--- a/.idea/kotlinc.xml
+++ b/.idea/kotlinc.xml
@@ -4,6 +4,6 @@
     <option name="jvmTarget" value="1.8" />
   </component>
   <component name="KotlinJpsPluginSettings">
-    <option name="version" value="1.7.20" />
+    <option name="version" value="1.8.20" />
   </component>
 </project>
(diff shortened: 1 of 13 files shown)


bentrengrove:
Thanks for the contribution! We are currently in the process of migrating to kts gradle files as part of #1578 but I don't see why we can't merge this in the mean time.

Please just fix up that merge conflict

stevan-milovanovic:
You're welcome. Sounds good 👍 I fixed them up, should be good to go 😄

stevan-milovanovic:
@bentrengrove sorry for the delay, I updated the changes.

bentrengrove:
Sorry, as mentioned you are working at the same time as the migration to KTS so this now conflicts again. I might close this for now unfortunately as by the end of the migration the bump from 8.0.0 to 8.0.1 should be trivial

PR #1624, opened 2023-05-18
[Testharness] Migrate gradle module to Kotlin DSL

oas004:

Migrating the `:testharness` module to Kotlin DSL.

### DIFF
--- a/testharness/build.gradle.kts
+++ b/testharness/build.gradle.kts
@@ -13,27 +13,42 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
+@file:Suppress("UnstableApiUsage")
 
 plugins {
-    id("com.android.library")
-    id("kotlin-android")
-    id("org.jetbrains.dokka")
-    id("me.tylerbwong.gradle.metalava")
+    id(libs.plugins.android.library.get().pluginId)
+    id(libs.plugins.android.kotlin.get().pluginId)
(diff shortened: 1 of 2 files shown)


testharness/build.gradle.kts
```diff
@@ -0,0 +1,154 @@
...
+    id(libs.plugins.vanniktech.maven.publish.get().pluginId)
+    jacoco
+}
+
+kotlin {
+    explicitApi()
+}
+
+tasks.withType<Test> {
+    configure<JacocoTaskExtension> {
```
oas004, line 32: I'm not quite sure about this. Are we actually using Jacoco in Accompanist for anything? If I checkout main and run `./gradlew testharness:tasks` I cannot find any tasks relating to Jacoco test coverage. Also doing a search through the codebase only shows Jacoco being referenced in the `:testharness` module if I'm not mistaken. 🤔

bentrengrove: @alexvanyo any ideas?

alexvanyo: We checked it when initially creating the module in #1392 via `./gradlew :testharness:createDebugUnitTestCoverageReport`, but it's not monitored and it seems like it doesn't even work properly right now. Removing the configuration here sounds good to me.

oas004: Alright! Thanks for the answer! I will remove this configuration this :)

bentrengrove (approved):

PR #1632, opened 2023-05-19
[Navigation-Animation] Fix NavController receiving back events even while AnimatedNavHost is not in composition

Leedwon:
Hi folks, I think there's a bug in the library which casues navController to receive back events even if `AnimatedNavHost` is not in composition. Because of that each time `AnimatedNavHost` gets recomposed new navController will be registered for receiving back events which can lead to weird issues with back navigation. Please check the MRE  and detailed description in [this repo](https://github.com/Leedwon/AnimatedNavControllerBackPressIssues). 

Fix is based on `androidx.navigation.compose.NavHost` which has the same mechanism for disabling back events in `onDispose`.


ianhanniballake:
As per [this comment](https://issuetracker.google.com/issues/197140101#comment13), all of Accompanist Navigation Animation has already been merged back into Navigation Compose itself, so we won't be updating Accompanist anymore - you'll just be able to use Navigation `2.7.0-alpha01` once it comes out, which, as you've mentioned, already contains this fix.

Issue #1636, opened 2023-05-25
[Navigation Material] crash after upgrade  to compose 1.5.0-beta01


ti4n1993:
```console
java.lang.NoSuchMethodError: No static method ModalBottomSheetLayout-BzaUkTc(Lkotlin/jvm/functions/Function3;
Landroidx/compose/ui/Modifier;
Landroidx/compose/material/ModalBottomSheetState;
Landroidx/compose/ui/graphics/Shape;
FJJJLkotlin/jvm/functions/Function2;
Landroidx/compose/runtime/Composer;II)V in class Landroidx/compose/material/ModalBottomSheetKt;
 or its super classes (declaration of 'androidx.compose.material.ModalBottomSheetKt' appears in package-name
at com.google.accompanist.navigation.material.BottomSheetKt.ModalBottomSheetLayout-4erKP6g(BottomSheet.kt:48)
```

miduch:
+1

SpiritualForest:
I'm also affected

javichaques:
+1

hovadur:
+1

bentrengrove:
This doesn't seem to reproduce in the sample app just by upgrading to 1.5.0-beta01. Does anyone else have any info? Does this just happen as soon as you show a bottom sheet?

bentrengrove:
I assume this is just a binary incompatible change then, I will do a release with an upgraded Compose which should fix it

bentrengrove:
This has been released as v0.31.3-beta. Please let me know if the issue continues

SpiritualForest:
Just reporting that the latest version fixed it for me. Thanks @bentrengrove 👍

Palmer1985:
https://github.com/google/accompanist/issues/1713
Please help this issue 
How could I solve it.

PR #1637, opened 2023-05-25
chore: update compose version to 1.5.0-beta01

amanv8060:


bentrengrove:
We don't normally do this unless an Accompanist library is experiencing a bug. If you use Compose 1.5-beta01 in your project then it will automatically get upgraded transitively.

Is not targetting beta01 causing you an issue?

bentrengrove:
I'm going to close this for now as this happens as part of the release process but if not targetting beta01 causes you an issue then please file an issue.

bentrengrove:
Was this an attempt to fix #1636? Sorry i missed that first, thank you for trying to fix that. It really is appreciated. 

We wouldn't be able to merge this as is because of the bump to navigation to 2.7.0 and also the version number gets changed as part of the release process not through a PR like this. Normally I would leave comments for you to fix this up so we could merge it in but seeing as that crash is affecting a lot of people I am going to do this PR myself to speed up the process.

Thanks again and sorry I couldn't merge it in this time.

amanv8060:
> Was this an attempt to fix https://github.com/google/accompanist/issues/1636?

Yup, sample app didn't reproduce error after upgrading the compose, compiler and Kotlin versions, and copying the exact library into the app as a module also didn't reproduce crash.

Also, would the navigation version be upgraded as part of the deprecation process for AnimatedNavHost? Is there a process for this, or can I create a PR after this release?

bentrengrove:
Yes, navigation will be upgraded as part of the deprecation process which is planned. We will handle that one but thank you for the offer.

PR #1642, opened 2023-05-29
[navigation-material] expose sheetGesturesEnabled flag

### LINKED ISSUE #730: Provide a method to make the bottom sheet non-cancellable for some use-cases.
## Describe the bug
Provide a method to make the bottom sheet non-cancellable for some use-cases.

## Expected behaviour
When a user clicks the back button or the outside content of the bottom sheet, then it dismisses. However, there should be a way to control that behaviour based on a flag that the bottom sheet should not dismiss. It should only dismiss via calling ```navGraph.navigateBack()```.

## Environment:
 - Android OS version: [Android 11.0]
 - Device: [Samsung M20]
 - Accompanist version: [0.17.0]

amanv8060:
Fixes #730 

Expose sheetGesturesEnabled Flag

### DIFF
--- a/navigation-material/src/androidTest/java/com/google/accompanist/navigation.material/BottomSheetNavigatorTest.kt
+++ b/navigation-material/src/androidTest/java/com/google/accompanist/navigation.material/BottomSheetNavigatorTest.kt
@@ -33,6 +33,8 @@ import androidx.compose.material.ModalBottomSheetState
 import androidx.compose.material.ModalBottomSheetValue
 import androidx.compose.material.Text
 import androidx.compose.runtime.DisposableEffect
+import androidx.compose.runtime.State
+import androidx.compose.runtime.collectAsState
 import androidx.compose.runtime.getValue
 import androidx.compose.runtime.mutableStateOf
 import androidx.compose.runtime.remember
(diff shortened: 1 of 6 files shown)


ianhanniballake:
Similar to the [comment](https://github.com/google/accompanist/pull/1342#issuecomment-1250828716) on #1342, we'd want to make this available on a per destination level, not a global flag.

This may be easier said than done given the current state of `ModalBottomSheetLayout` (also as explained in that comment), but if it is possible, we'd be happy to merge a destination level flag in if you'd like to put up a new patch set.

amanv8060:
Will give it a try, although it doesn't look easily possible.

amanv8060:
@ianhanniballake, I have added a destination-level flag and updated the bottom sheet sample to demonstrate the same. I am a little unsure about the performance implications of this approach.

Can you take a look and let me know what you think?

If this looks good, I can try adding tests for this.

jossiwolf:
@amanv8060, thanks for the PR!

I'd be interested to see if tests for the following scenarios (see the comment Ian linked) would pass:

> The concerns are about recreating the state in time when the value changes as the destination changes, but maybe it'd work. We'd want some good test coverage that tests
> 
> Navigating from a skipHalfExpanded = false to a skipHalfExpanded = false destination
> Navigating from a skipHalfExpanded = false to a skipHalfExpanded = true destination
> Navigating from a skipHalfExpanded = true to a skipHalfExpanded = false destination
> and ensures that the sheet opens up to the correct state.

If yes, we can look at performance a bit more. What concerns do you have here?

amanv8060:
I will try adding tests for these cases.

amanv8060:
cc/ @jossiwolf icymi

VahidGarousi:
Would love to see this merged ✌️

ianhanniballake:

PR #1655, opened 2023-06-07
Add manual baseline profiles for each project

keyboardsurfer:
By creating a manual baseline profile for each of these libraries, all code will be part of the resulting baseline profile.
This is the most lightweight catch all approach to add accompanist code to baseline profiles.

We could add an explicit dependency on profileinstaller, but that's being brought in from Compose either way.
So adding the new dependency here would only add to maintenance overhead.

### DIFF
--- a/adaptive/src/main/baseline-prof.txt
+++ b/adaptive/src/main/baseline-prof.txt
@@ -1,0 +1,2 @@
+# include everything
+HSPLcom/google/accompanist/adaptive/**->**(**)**
--- a/appcompat-theme/src/main/baseline-prof.txt
+++ b/appcompat-theme/src/main/baseline-prof.txt
@@ -1,0 +1,2 @@
+# include everything
+HSPLcom/google/accompanist/appcompattheme/**->**(**)**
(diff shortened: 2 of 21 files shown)


Tolriq:
Would not that give a bad signal to all libraries to do that and in the end leads to a negative outcome?

keyboardsurfer:
I agree that not all libraries should do this. Each of the accompanist libraries is so small that it's acceptable to have a manual wildcard. In larger libraries, such as Compose we have manual rules and they are well maintained.
The recommendation for library developers is to use [the gradle plugin to generate a profile](https://developer.android.com/topic/performance/baselineprofiles/create-baselineprofile#baseline-profile-library)

Tolriq:
Yes I know the recommendation, just thinking out loud for the maintainers that blindly copy.
It took time to have quite a few libraries fix their too broad proguard rules, and I fear this will be the same for some time for baselines.
Maybe there's a way to put a comment in the commit or in each files?

It also took a long time to have R8 be able to easily exclude wrong rules provided by librairies and I do not think there's already a way to exclude wrong baseline profiles provided by libraries?

keyboardsurfer:
As discussed we won't move ahead with this PR.

PR #1672, opened 2023-07-05
[Placeholder] Deprecating accompanist placeholder

florina-muntenescu:


alirahimpour89:
Why deprecate this before having a good alternative? 😕 Will miss this one!

bentrengrove:
We are currently evaluating the future of some Accompanist libraries. We have decided that the placeholder functionality is easily reproduced and able to be customised via a fork or copy and we can't produce every variant that people would want so it doesn't make sense to upstream it to the main Compose libraries. It makes more sense for people to use it as a reference implementation and create their own that meets their needs. It will still be available here in a deprecated form and is safe to use, but we won't be making any changes to it going forwards unfortunately.

bentrengrove (approved):

PR #1680, opened 2023-07-16
[Permissions] Alter use of READ_EXTERNAL_STORAGE permission from permissions samples.

### LINKED ISSUE #1679: [Permissions] multiple permission sample does not work on API Level >= 33 due to obsolete permission constant
**Description**

I was investigating permissions feature and running the sample app, I noticed that the existing `RequestMultiplePermissionsSample` would not work as expected, because `READ_EXTERNAL_STORAGE` does not exist since API Level 33. I never got `"Camera and Read storage permissions Granted! Thank you!"` state because this permission is not going to be regarded as approved at `rememberMultiplePermissionsState()` whereas it will never be asked at `launchMultiplePermissionRequest()`.

**Steps to reproduce**

Launch the existing "Permissions: Request multiple permissions" sample, and approve ALL the *requested* permissions.

**Expected behavior** 

It should result in the "Thank you!" message above.

**Additional context**

I have a fix that simply replaces `READ_EXTERNAL_STORAGE` with `RECORD_AUDIO`, will be sent .

atsushieno:
fixes #1679.

RequestMultiplePermissionsSample was using READ_EXTERNAL_STORAGE, but this permission was removed at API Level 33 for more detailed items. It leads to the state where rememberMultiplePermissionsState() never count it as approved whereas launchMultiplePermissionRequest() will never ask the user for it as a valid permission, and the entire sample never gets to the "both approved" state.

Use another permission entry which is still valid at any API level.

### DIFF
--- a/sample/src/main/AndroidManifest.xml
+++ b/sample/src/main/AndroidManifest.xml
@@ -24,7 +24,7 @@
     <!-- Used for the permissions sample -->
     <uses-permission android:name="android.permission.CAMERA" />
     <uses-feature android:name="android.hardware.camera" android:required="false"/>
-    <uses-permission android:name="android.permission.READ_EXTERNAL_STORAGE" />
+    <uses-permission android:name="android.permission.RECORD_AUDIO" />
     <uses-permission android:name="android.permission.ACCESS_FINE_LOCATION" />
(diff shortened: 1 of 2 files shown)


bentrengrove (approved):

bentrengrove:
Thank you for this fix!

PR #1684, opened 2023-07-21
[Web]Remove the wrapper of webview to solve nestedscroll issue

### LINKED ISSUE #1652: [WebView] Nestedscroll Does not work in Compose WebView
**Description**
In this PR:https://github.com/google/accompanist/pull/1483, the webview is now wrapped with a Framelayout and then returned to AndroidViewHolder. However,  the AndroidViewHolder use view?.isNestedScrollingEnabled to determine whether the child view support nested scroll. It works well before this change because Webview supports nested scroll but failed now since FrameLayout does not.

**Steps to reproduce**
Just use the webview with a container that has some nested scroll logic, and you will see the abnormal behaviors.

**Expected behavior** 
The nested scroll should work well like the version before this pr

**Additional context**
Please let me know if there is any workaround for that problem.

KevinnZou:
In this [PR](https://github.com/google/accompanist/pull/1483), Ben wrapped the Webview with a Framelayout to workaround rare crash. However, this led to the problem that the nested scroll does not work now. Since the AndroidViewHolder use view?.isNestedScrollingEnabled to determine whether the child view support nested scroll. It works well before this change because Webview supports nested scroll but failed now since FrameLayout does not. I opened an [issue](https://github.com/google/accompanist/issues/1652) about this problem and Ben approved me to remove the Framelayout first to workaround this issue. Thus I finished it and submit this PR

Fixes #1652

### DIFF
--- a/web/src/main/java/com/google/accompanist/web/WebView.kt
+++ b/web/src/main/java/com/google/accompanist/web/WebView.kt
@@ -44,7 +44,6 @@ import androidx.compose.runtime.snapshotFlow
 import androidx.compose.runtime.snapshots.SnapshotStateList
 import androidx.compose.ui.Modifier
 import androidx.compose.ui.viewinterop.AndroidView
-import androidx.core.view.children
 import com.google.accompanist.web.LoadingState.Finished
 import com.google.accompanist.web.LoadingState.Loading
 import kotlinx.coroutines.CoroutineScope
@@ -225,7 +224,8 @@ fun WebView(
         },
         modifier = modifier,
         onRelease = { parentFrame ->
-            val wv = parentFrame.children.first() as WebView
+            // Since FrameLayout is removed,parentFrame is just the WebView
+            val wv = parentFrame
             onDispose(wv)
         }
     )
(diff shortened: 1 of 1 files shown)


web/src/main/java/com/google/accompanist/web/WebView.kt
```diff
@@ -222,19 +221,11 @@ fun WebView(
...
-            parentLayout.layoutParams = layoutParams
-            parentLayout.addView(childView)
-
-            parentLayout
         },
         modifier = modifier,
         onRelease = { parentFrame ->
-            val wv = parentFrame.children.first() as WebView
+            // Since FrameLayout is removed,parentFrame is just the WebView
+            val wv = parentFrame
```
bentrengrove, line 228: So the name `parentFrame` was for when it was wrapped in a `FrameLayout`. You can just remove most of this and just call `onDispose(it)`

bentrengrove (changes requested):
Looking good! Thanks for this, just one minor thing to fix up and then it will be good to merge

web/src/main/java/com/google/accompanist/web/WebView.kt
KevinnZou: Thanks for your comment! I have removed the unused code and resubmitted the pr.

bentrengrove (approved):
Thank you for the contribution!

PR #1685, opened 2023-07-26
[Insets] Remove library from Accompanist

bentrengrove:
Insets has been deprecated for multiple releases now so I am removing it completely from the repository.

I will leave the guide up for now in case people click the link, probably one more release.


.idea/migrations.xml
```diff
@@ -0,0 +1,10 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
alexvanyo, line 1: Nit: just checking if adding this was intentional?

README.md
```diff
@@ -67,9 +67,6 @@ A library providing a collection of utilities for adaptive layouts.
 ### 🗜 [Test Harness](./testharness/)
 Utilities for testing Compose layouts.
 
-### 📐 [Insets](./insets/) (Deprecated)
```
alexvanyo, line 70: It might be worth leaving this in the README as "Deprecated and Removed" with a link to the guide of developer.android.com.

permissions-lint/lint-report.html
alexvanyo, line 1: This probably shouldn't be included either

alexvanyo (commented):
And so the cycle for insets is (almost) complete! 🎉 

A couple minor requests, and can we remove the insets issue template as well?
https://github.com/google/accompanist/blob/main/.github/ISSUE_TEMPLATE/insets-bug-report.md

bentrengrove:
All fixed!

alexvanyo (approved):
Looks great!

PR #1709, opened 2023-08-25
[Docs] Replace corp URLs with publicly accessible ones

msfjarvis:
Replaces a couple instances of corp URLs with their publicly accessible versions.


alexvanyo (approved):
Thanks for catching that!

PR #1715, opened 2023-09-24
Upgrade Compose to 1.6.0-alpha06

### LINKED ISSUE #1714: 0.33.1-alpha is not compatible with compose 1.6.0-alpha06 or later
All libraries using `mutableFloatStateOf()` and built against 1.6.0 alphas before alpha06 are broken when used with newer alphas due to an accidental breaking ABI change that was fixed in alpha06. As such, the latest current alpha is not usable with compose alpha06 and beyond.

Ref: https://issuetracker.google.com/issues/298810665

bentrengrove:
Fixes #1714

### DIFF
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -1,7 +1,7 @@
 [versions]
 
-compose = "1.6.0-alpha02"
-composeCompiler = "1.5.1"
+compose = "1.6.0-alpha06"
+composeCompiler = "1.5.3"
 composeMaterial3 = "1.0.1"
 composesnapshot = "-" # a single character = no snapshot
 
@@ -12,7 +12,7 @@ gradlePlugin = "8.1.0"
 lintMinCompose = "30.0.0"
 
 ktlint = "0.45.2"
-kotlin = "1.9.0"
+kotlin = "1.9.10"
 coroutines = "1.6.4"
 okhttp = "3.12.13"
 coil = "1.3.2"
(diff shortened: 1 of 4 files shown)


gradle/libs.versions.toml
```diff
@@ -12,7 +12,7 @@ gradlePlugin = "8.1.0"
 lintMinCompose = "30.0.0"
 
 ktlint = "0.45.2"
-kotlin = "1.9.0"
```
alexvanyo, line 15: Not blocking - is bumping Kotlin and the compiler require from accompanist? Just wondering if the transitive Kotlin bump here is worth it (or if that bump would be transitive anyway from base Compose)

sample/src/main/java/com/google/accompanist/sample/adaptive/NavDrawerFoldAwareColumnSample.kt
```diff
@@ -68,7 +67,6 @@ class NavDrawerFoldAwareColumnSample : ComponentActivity() {
 @Composable
 fun NavDrawerExample(activity: Activity) {
     val icons = listOf(
-        Icons.Default.List,
```
alexvanyo, line 71: Nit: Was this removed? Or was this one of the `AutoMirrored` icons?

alexvanyo (approved):

sample/src/main/java/com/google/accompanist/sample/adaptive/NavDrawerFoldAwareColumnSample.kt
bentrengrove: It was automirrored but the label looks weird in the sample because of splitting at the first `.`. It seemed easier to just remove it rather than fix that

gradle/libs.versions.toml
bentrengrove: We always just release with the latest compiler to make sure we stay up to date with fixes there. I haven't heard an issue yet with an incompatible change

PR #1728, opened 2023-11-20
[Navigation Material] Add null check for retainedEntry in BackHandler

Vulama:
BackHandler sometimes crashes the application with the `NullPointerException`. 
Log from the FirebaseCrashlytics:

```
Fatal Exception: java.lang.NullPointerException:
       at com.google.accompanist.navigation.material.BottomSheetNavigator$sheetContent$1$2.invoke(BottomSheetNavigator.kt:224)
       at com.google.accompanist.navigation.material.BottomSheetNavigator$sheetContent$1$2.invoke(BottomSheetNavigator.kt:223)
```

This seems to be a random issue on low-end devices. 

This issue was noticed using `com.google.accompanist:accompanist-navigation-material:0.32.0` on a large-scale application.


bentrengrove:
Thanks for the fix! We have fixed this in a different way in #1736 so I will close this one out.

PR #1731, opened 2023-11-29
[Documentation] Fix warning box/text in docs

Monkey-Matt:
Some pages are currently incorrectly formatted so the warning boxes don't have text in them
Screenshot 2023-11-29 at 2 28 57 PM

This PR just changes the formatting so they show correctly.
The warning boxes (and caution boxes) should show with text inside it as you see on docs pages like [insets](https://google.github.io/accompanist/insets/)
Screenshot 2023-11-29 at 2 28 06 PM

I [previously made this fix](https://github.com/google/accompanist/pull/1724) for the placeholder page but didn't realise there is the same problem in other places. This PR address all the places I found


bentrengrove (approved):
Thank you!

PR #1742, opened 2024-01-24
Upgrade to Compose 1.6 stable

bentrengrove:


arriolac (approved):

IanGClifton:
@warting That bug should be resolved by depending on BOM 2024.02.00

PR #1743, opened 2024-01-25
[TestHarness] Deprecate testharness

alexvanyo:
Deprecates testharness with the replacement of `DeviceConfigurationOverride`.

This PR also replaces the internal usage in some other module's tests, and provides a migration guide to the new methods.

### DIFF
--- a/testharness/build.gradle.kts
+++ b/testharness/build.gradle.kts
@@ -110,6 +110,7 @@ dependencies {
     implementation(libs.androidx.core)
     testImplementation(libs.androidx.core)
     implementation(libs.kotlin.coroutines.android)
+    implementation(libs.compose.ui.test)
 
     // ======================
     // Test dependencies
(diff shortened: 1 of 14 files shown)


testharness/build.gradle.kts
```diff
@@ -110,6 +110,7 @@ dependencies {
     implementation(libs.androidx.core)
     testImplementation(libs.androidx.core)
     implementation(libs.kotlin.coroutines.android)
+    implementation(libs.compose.ui.test)
```
alexvanyo, line 113: Added for use in the `ReplaceWith` in the `@Deprecated` annotation.

testharness/src/main/java/com/google/accompanist/testharness/TestHarness.kt
```diff
@@ -70,6 +70,32 @@ import kotlin.math.floor
  * the [isScreenRound] will be left unchanged.
  */
 @Composable
+@Deprecated(
+    replaceWith = ReplaceWith(
```
alexvanyo, line 74: This isn't a perfect `ReplaceWith` - the resulting code will likely be far more verbose than necessary, with the assumption that usually only one or a couple parameters were specified, but it brings in the imports and can then be edited down to the important ones.

bentrengrove:
This looks good, let me branch off Compose 1.6 before we merge this in though

bentrengrove (approved):

Issue #1750, opened 2024-02-05
[Navigation Material] Crash in BottomSheetNavigator when user goes back


filip-misadin:
We have some crashes that, as far as I can see when looking at the logs, happen when a user goes back on any of our bottom modals. The version of Navigation Material library we currently use is `0.32.0`. I looked into the newest version (`0.34.0`) and by looking into the code I think we would have the same crashes with that version too, even though the code changed to some degree. 

I tried a lot of ways to reproduce this crash, but I didn't succeed in reproducing it. That's why I can't know for sure why this happens and what is the edge case that keeps popping up. This crash happens to around 1% of our users, so it's not that uncommon. 

This is the crash stack trace:
```
Fatal Exception: java.lang.NullPointerException
com.google.accompanist.navigation.material.BottomSheetNavigator$sheetContent$1$2.invoke (BottomSheetNavigator.kt:224)
com.google.accompanist.navigation.material.BottomSheetNavigator$sheetContent$1$2.invoke (BottomSheetNavigator.kt:223)
androidx.activity.compose.BackHandlerKt$BackHandler$backCallback$1$1.handleOnBackPressed (BackHandler.kt:89)
androidx.activity.OnBackPressedDispatcher.onBackPressed (OnBackPressedDispatcher.kt:276)
androidx.activity.ComponentActivity.onBackPressed (ComponentActivity.java:678)
android.app.Activity.onKeyUp (Activity.java:3976)
android.view.KeyEvent.dispatch (KeyEvent.java:2925)
android.app.Activity.dispatchKeyEvent (Activity.java:4329)
androidx.core.app.ComponentActivity.superDispatchKeyEvent (ComponentActivity.java:120)
androidx.core.view.KeyEventDispatcher.dispatchKeyEvent (KeyEventDispatcher.java:85)
androidx.core.app.ComponentActivity.dispatchKeyEvent (ComponentActivity.java:138)
com.android.internal.policy.DecorView.dispatchKeyEvent (DecorView.java:420)
android.view.ViewRootImpl$ViewPostImeInputStage.processKeyEvent (ViewRootImpl.java:6735)
android.view.ViewRootImpl$ViewPostImeInputStage.onProcess (ViewRootImpl.java:6601)
android.view.ViewRootImpl$InputStage.deliver (ViewRootImpl.java:6059)
android.view.ViewRootImpl$InputStage.onDeliverToNext (ViewRootImpl.java:6116)
[...]

Issue #1733, opened 2023-12-06
[BottomSheetNavigator] NullPointerException (race-condition?) when using back button


alixwar:
## Describe the bug
This is a crash reported from an analytics tool so I cannot give more details than the stacktrace and the devices where the issues were found.

## To Reproduce
```
Fatal Exception: java.lang.NullPointerException:
       at com.google.accompanist.navigation.material.BottomSheetNavigator$sheetContent$1$2.invoke(BottomSheetNavigator.kt:224)
       at com.google.accompanist.navigation.material.BottomSheetNavigator$sheetContent$1$2.invoke(BottomSheetNavigator.kt:223)
       at androidx.activity.compose.BackHandlerKt$BackHandler$backCallback$1$1.handleOnBackPressed(BackHandler.kt:89)
       at androidx.activity.OnBackPressedDispatcher.onBackPressed(OnBackPressedDispatcher.kt:276)
       at androidx.activity.OnBackPressedDispatcher$5.invoke(OnBackPressedDispatcher.kt:134)
       at androidx.activity.OnBackPressedDispatcher$5.invoke(OnBackPressedDispatcher.kt:134)
       at androidx.activity.OnBackPressedDispatcher$Api33Impl.createOnBackInvokedCallback$lambda$0(OnBackPressedDispatcher.kt:369)
       at android.view.ViewRootImpl$NativePreImeInputStage.onProcess(ViewRootImpl.java:7558)
       at android.view.ViewRootImpl$InputStage.deliver(ViewRootImpl.java:7213)
       at android.view.ViewRootImpl.deliverInputEvent(ViewRootImpl.java:10788)
       at android.view.ViewRootImpl.doProcessInputEvents(ViewRootImpl.java:10676)
       at android.view.ViewRootImpl.enqueueInputEvent(ViewRootImpl.java:10632)
       at android.view.ViewRootImpl$WindowInputEventReceiver.onInputEvent(ViewRootImpl.java:10926)
       at android.view.InputEventReceiver.dispatchInputEvent(InputEventReceiver.java:285)
       at android.os.MessageQueue.nativePollOnce(MessageQueue.java)
       at android.os.MessageQueue.next(MessageQueue.java:335)
       at android.os.Looper.loopOnce(Looper.java:186)
       at android.os.Looper.loop(Looper.java:313)
       at android.app.ActivityThread.main(ActivityThread.java:8757)
       at java.lang.reflect.Method.invoke(Method.java)
[...]

PR #1753, opened 2024-02-13
[navigation-material] Fixes crash in BottomSheetNavigator when user goes back

LluisFelip:
Attempting to address https://github.com/google/accompanist/issues/1750 and https://github.com/google/accompanist/issues/1733.

Although I couldn't reproduce the `NullPointerException` on `BottomSheetNavigator.kt:191`, by looking at the code I suggest avoiding the not-null assertion operator on `retainedEntry` with a `let` scope function.

### DIFF
--- a/navigation-material/src/main/java/com/google/accompanist/navigation/material/BottomSheetNavigator.kt
+++ b/navigation-material/src/main/java/com/google/accompanist/navigation/material/BottomSheetNavigator.kt
@@ -182,13 +182,13 @@ public class BottomSheetNavigator(
                 }
         }
 
-        if (retainedEntry != null) {
-            LaunchedEffect(retainedEntry) {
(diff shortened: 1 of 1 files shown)


jossiwolf:
Thanks for the PR! We will need a test for this, for which we might need steps to repro. Please let us know if you find any and are able to add a test!

LluisFelip:
Unfortunately I couldn't find any way of reproducing nor testing it, since it seems to be some sort of race-condition with that `retainedEntry` val inside that `sheetContent` Composable function 😞 

We're going to try it with real users by replacing this module with the fixed version of it with real users, so we'll be able to assess if it fixes the issue.

Any idea/suggestion on how to test this?

ianhanniballake:

PR #1755, opened 2024-02-18
[pager-indicators] Drop false-positive deprecation of pagerTabIndicatorOffset()

Philipp91:
Fixes #1744.

The deprecation was intended to apply only to the old overload above, but was presumably copy-pasted by accident.

PR #1757, opened 2024-02-26
[Pager] Remove false positive deprecation message from pagerTabIndicatorOffset

### LINKED ISSUE #1739: False positive deprecation message on pagerTabIndicatorOffset variant
The deprecation message for the method indicates that the `androidx.compose.foundation.pager.PagerState` version should be used.

However, this applies to both versions of the modifier, including the one already using `androidx.compose.foundation.pager.PagerState`. The message seems to suggest an alternative that is essentially the same as the current implementation.

https://github.com/google/accompanist/blob/3af68c89ca6686b4b611f41e8d5e8cffe5f4b56e/pager-indicators/src/main/java/com/google/accompanist/pager/PagerTab.kt#L63-L73

PhilipDukhov:
This method overload is using `androidx.compose.foundation.pager.PagerState` and shouldn't be marked as deprecated.

As doc [says](https://github.com/google/accompanist/blob/1645e59d53844e71fe7dbe9442309a6237f650a8/docs/pager.md?plain=1#L36):

> Implement it yourself, or still include and use `accompanist-pager-indicators`, it now supports `androidx.compose.foundation.pager.PagerState`

Closes #1739

### DIFF
--- a/pager-indicators/src/main/java/com/google/accompanist/pager/PagerTab.kt
+++ b/pager-indicators/src/main/java/com/google/accompanist/pager/PagerTab.kt
@@ -60,13 +60,6 @@ public fun Modifier.pagerTabIndicatorOffset(
  * [androidx.compose.foundation.pager.VerticalPager].
  */
 @OptIn(ExperimentalFoundationApi::class)
-@Deprecated(
-    """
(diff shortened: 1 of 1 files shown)


bentrengrove:
This looks like a dupe of #1755 so I will close this one as it is newer

PR #1769, opened 2024-04-29
Implemented DismissDispatcher to be able to intercept BottomSheet navigation dismissal

PavloZoria:
### Implemented DismissDispatcher: Intercepting BottomSheet Dismissal

I've implemented DismissDispatcher to provide more control over user-initiated dismissal of BottomSheets. This functionality mirrors the behavior of BackHandler for handling back button presses.

### What can you intercept?

- Back press: Dismissal triggered by pressing the back button.
- BottomSheet interactions:
   - Clicking on the scrim (background dimming area) outside the sheet content.
   - Swiping down on the sheet itself.

### How to use DismissDispatcher:

1. Call DismissHandler in your sheet's content.
2. Override the callback function. The usage is similar to BackHandler.

Code Example:

```Kotlin
@Composable
private fun BottomSheet(
  navigateUp: () -> Unit,
) {
  var showConfirmationDialog by remember {
    mutableStateOf(false)
  }

  if (showConfirmationDialog) {
    ConfirmationDialog(
      onCancel = { showConfirmationDialog = false },
      onApprove = { navigateUp() }
    )
  }

  DismissHandler {
    showConfirmationDialog = true
  }

  BottomSheetContent()
}
```

### Important Note:

This approach does not intercept dismissal initiated by the NavController. This allows you to easily close the BottomSheet programmatically using the NavController if needed.

### Attachments

https://github.com/google/accompanist/assets/38030130/2f7bd0d6-39e7-44d7-8778-b77da91c2613

### DIFF
--- a/navigation-material/api/current.api
+++ b/navigation-material/api/current.api
@@ -44,3 +44,14 @@ package com.google.accompanist.navigation.material {
 
 }
 
+package com.google.accompanist.navigation.material.dismiss {
+
+  public final class DismissDispatcherKt {
+  }
+
+  public final class DismissHandlerKt {
+    method @androidx.compose.runtime.Composable public static void DismissHandler(optional boolean isEnabled, kotlin.jvm.functions.Function0<kotlin.Unit> onDismiss);
+  }
+
+}
+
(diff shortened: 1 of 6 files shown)


bentrengrove:
Thank you for the contribution but unfortunately Navigation Material has been deprecated in Accompanist as the functionality is now available in the main Compose libraries.

Because of this I will have to close out this PR as we won't be making any further changes here.

PavloZoria:
@bentrengrove I'd like your thoughts on two options:

- Can we achieve this functionality by manually extending components from the Compose library? If so, could you provide feedback on a potential implementation approach?
- Do you think it would be appropriate to request this functionality as a feature in the Material library? What are the chances of it being included?

PR #1783, opened 2024-07-12
[Permissions] fix(gh-1781): check for `Denied(shouldShowRationale=false)` in `MutableMultiplePermissionsState.shouldShowRationale`

### LINKED ISSUE #1781: [Permissions] `MutableMultiplePermissionsState.shouldShowRationale` returns true despite some permissions being `Denied(shouldShowRationale=false)`
**Description**

The current implementation of `MutableMultiplePermissionsState.shouldShowRationale` does not consider cases where one or more permissions are `Denied(shouldShowRationale=false)`. This causes issues in statements such as

```kt
val permissions = rememberMultiplePermissionsState(listOf(ACCESS_FINE_LOCATION, ACCESS_BACKGROUND_LOCATION))

when {
  // Granted
  permissions.allPermissionsGranted -> Unit,
  // Denied, but I can ask the user
  permissions.shouldShowRationale -> 
    // ERROR: Does not trigger!
    permissions.launchMultiplePermissionRequest()
  // Denied and I may not ask the user.
  else -> Unit
}
```

The issue is located at

https://github.com/google/accompanist/blob/57907e1c9aadef1f70e51889c04048693e983f21/permissions/src/main/java/com/google/accompanist/permissions/MutableMultiplePermissionsState.kt#L130

Suggested fix:

```diff
    override val shouldShowRationale: Boolean by derivedStateOf {
-       permissions.any { it.status.shouldShowRationale }
+       permissions.any { it.status.shouldShowRationale } && permissions.none { !it.status.isGranted && !it.status.shouldShowRationale }
    }
```

FelixZY:
Adding the fix I proposed in #1781 

Fixes #1781

---

Additional complication:

I have indications that `ACCESS_BACKGROUND_LOCATION` may be `Denied(shouldShowRationale=false)` while `ACCESS_FINE_LOCATION` is `Denied(shouldShowRationale=true)`. It then seems like `ACCESS_BACKGROUND_LOCATION` shifts to `Denied(shouldShowRationale=true)` once `ACCESS_FINE_LOCATION` is granted.

I have not yet confirmed this but something seems to be acting strange with these two. If the above description is correct, the proposed solution in this PR is insufficient (though covering an unexpected gap that exists today).

I think someone should take a closer look and disprove the above theory before merging this (that'll probably be a few months out on my end - if even then).

### DIFF
--- a/permissions/src/main/java/com/google/accompanist/permissions/MutableMultiplePermissionsState.kt
+++ b/permissions/src/main/java/com/google/accompanist/permissions/MutableMultiplePermissionsState.kt
@@ -127,7 +127,8 @@ internal class MutableMultiplePermissionsState(
     }
 
     override val shouldShowRationale: Boolean by derivedStateOf {
-        permissions.any { it.status.shouldShowRationale }
+        permissions.any { it.status.shouldShowRationale } &&
+            permissions.none { !it.status.isGranted && !it.status.shouldShowRationale }
     }
 
     override fun launchMultiplePermissionRequest() {


bentrengrove:
Sorry for the slow response here, I am planning to get to this. Just trying to get the beta release out first

bentrengrove (approved):
Thanks for this and sorry for the slow review. Could you please rebase this to main just to kick the CI?

FelixZY:
Sure! It will be a day or two but I'll do so asap.

FelixZY:
@bentrengrove rebase completed :rocket:

lafamiliadedios1521-spec:
/

PR #1789, opened 2024-08-30
Fix markdown layout bug in navigation-material.md

SimonMarquis:
See https://google.github.io/accompanist/navigation-material/


alexvanyo (approved):
Looks good, thanks for the fix!

PR #1793, opened 2024-09-22
[Permissions] Permissions enhancements

eygraber:
Still need to update the tests and docs, but looking to get feedback.

### DIFF
--- a/permissions/src/main/java/com/google/accompanist/permissions/MutablePermissionState.kt
+++ b/permissions/src/main/java/com/google/accompanist/permissions/MutablePermissionState.kt
@@ -87,6 +87,7 @@ internal class MutablePermissionState(
     private val activity: Activity
 ) : PermissionState {
 
+    internal var isPreviousStatusNotRequested = true
     internal var isPermissionPostRequest = false
 
     override var status: PermissionStatus by mutableStateOf(getPermissionStatus())
@@ -114,9 +115,23 @@ internal class MutablePermissionState(
         } else {
             val shouldShowRationale = activity.shouldShowRationale(permission)
             when {
-                isPermissionPostRequest -> when {
-                    shouldShowRationale -> PermissionStatus.NotGranted.Denied
-                    else -> PermissionStatus.NotGranted.PermanentlyDenied
+                isPermissionPostRequest -> {
+                    when {
+                        shouldShowRationale -> PermissionStatus.NotGranted.Denied
+                        else -> when {
+                            // we can't go from NotRequested->PermanentlyDenied since there's
+                            // no way to determine if the permission was already PermanentlyDenied
+                            // or if the permission request was canceled.
+                            // isPreviousStatusNotRequested forces an extra step by inserting a
+                            // Denied status even though it should be PermanentlyDenied
+                            // since this will allow us to handle the case where the permission
+                            // request was canceled
+                            isPreviousStatusNotRequested -> PermissionStatus.NotGranted.Denied
+                            else -> PermissionStatus.NotGranted.PermanentlyDenied
+                        }
+                    }.also {
+                        isPreviousStatusNotRequested = false
+                    }
                 }
 
                 else -> when {
--- a/permissions/src/main/java/com/google/accompanist/permissions/MutablePermissionState.kt
+++ b/permissions/src/main/java/com/google/accompanist/permissions/MutablePermissionState.kt
@@ -86,7 +86,7 @@ internal class MutablePermissionState(
     private val activity: Activity
 ) : PermissionState {
 
-    override var status: PermissionStatus by mutableStateOf(getPermissionStatus())
+    override var status: PermissionStatus by mutableStateOf(getPermissionStatus(false))
 
     override fun launchPermissionRequest() {
         launcher?.launch(
@@ -94,18 +94,33 @@ internal class MutablePermissionState(
         ) ?: throw IllegalStateException("ActivityResultLauncher cannot be null")
     }
 
+    override fun openAppSettings() {
+        activity.openAppSettings(permission)
+    }
+
     internal var launcher: ActivityResultLauncher<String>? = null
 
     internal fun refreshPermissionStatus() {
-        status = getPermissionStatus()
+        status = getPermissionStatus(isPermissionPostRequest = true)
     }
 
-    private fun getPermissionStatus(): PermissionStatus {
+    private fun getPermissionStatus(isPermissionPostRequest: Boolean): PermissionStatus {
         val hasPermission = context.checkPermission(permission)
         return if (hasPermission) {
             PermissionStatus.Granted
         } else {
-            PermissionStatus.Denied(activity.shouldShowRationale(permission))
+            val shouldShowRationale = activity.shouldShowRationale(permission)
+            when {
+                isPermissionPostRequest -> when {
+                    shouldShowRationale -> PermissionStatus.NotGranted.Denied
+                    else -> PermissionStatus.NotGranted.PermanentlyDenied
+                }
+
+                else -> when {
+                    shouldShowRationale -> PermissionStatus.NotGranted.Denied
+                    else -> PermissionStatus.NotGranted.NotRequested
+                }
+            }
         }
     }
 }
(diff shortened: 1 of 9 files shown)


bentrengrove:
Haven't reviewed yet, but please summarise your API changes in the PR as well please. It helps to link back to from the release notes so people can see what changed and why.

permissions/src/main/java/com/google/accompanist/permissions/MutablePermissionState.kt
```diff
@@ -86,26 +86,41 @@ internal class MutablePermissionState(
...
 
+    override fun openAppSettings() {
+        activity.openAppSettings(permission)
+    }
+
     internal var launcher: ActivityResultLauncher<String>? = null
 
     internal fun refreshPermissionStatus() {
-        status = getPermissionStatus()
+        status = getPermissionStatus(isPermissionPostRequest = true)
```
alexvanyo, line 104: `refreshPermissionStatus()` is also called from `PermissionLifecycleCheckerEffect`, so right now the initial lifecycle event refreshing the status when the permission isn't granted will result in a status of `PermanentlyDenied`.

permissions/src/main/java/com/google/accompanist/permissions/MutablePermissionState.kt
```diff
@@ -86,26 +86,41 @@ internal class MutablePermissionState(
...
         val hasPermission = context.checkPermission(permission)
         return if (hasPermission) {
             PermissionStatus.Granted
         } else {
-            PermissionStatus.Denied(activity.shouldShowRationale(permission))
+            val shouldShowRationale = activity.shouldShowRationale(permission)
+            when {
+                isPermissionPostRequest -> when {
+                    shouldShowRationale -> PermissionStatus.NotGranted.Denied
+                    else -> PermissionStatus.NotGranted.PermanentlyDenied
```
alexvanyo, line 116: There's a case here where it's possible to have requested the permission and `shouldShowRationale` is false, but without the permission being permanently denied.

If a permission is requested for the first time, and the user uses the back button to exit the permission dialog without selecting any of the options, the permission won't be permanently denied. If the permission is requested again, then the dialog will be shown again normally. Also, since the user did not explicitly deny the permission, `shouldShowRationale` will still be false.

permissions/src/main/java/com/google/accompanist/permissions/MutablePermissionState.kt
eygraber: I can track `isPermissionPostRequest` as its own property and set it separately from `refreshPermissionStatus`, since once that is set to `true` there shouldn't be any scenario that would make it go back to `false`.

eygraber: Is there any ordering guarantees between `rememberLauncherForActivityResult` and `PermissionLifecycleCheckerEffect`?

eygraber:
With https://github.com/google/accompanist/pull/1793/commits/a12ee22482423496cd71bd52e6c85fe5813d9809 these states should be correct

```
NotRequested -> Grant in Settings -> Resume -> Granted
Not Requested -> Pause -> Resume -> !shouldShowRationale -> NotRequested

NotRequested -> Request -> Grant -> Resume -> Granted
NotRequested -> Request -> Deny -> Resume -> shouldShowRationale -> Denied
NotRequested -> Request -> NotShown -> Resume -> !shouldShowRationale -> PermanentlyDenied

Denied -> Grant in Settings -> Resume -> Granted
Denied -> Pause -> Resume -> shouldShowRationale -> Denied

PermanentlyDenied -> Grant in Settings -> Resume -> Granted
PermanentlyDenied -> Pause -> Resume -> !shouldShowRationale -> PermanentlyDenied
```

The only issue as @alexvanyo  pointed out, is if the user canceled the dialog:

```
NotRequested -> Request -> Cancel -> Resume -> !shouldShowRationale -> PermanentlyDenied
```

As a user, this doesn't happen to me often, but it does happen enough times (usually by mistake) that I can see it being a problem. Any ideas on how to handle that case?

alexvanyo:
Using the lifecycle effect, I wonder if it is possible to infer whether the permission dialog was actually shown (in the not requested state) versus no permission dialog shown at all (in the permanently denied state)?

This is even more of an edge case, but it would also be possible for the user to cancel the dialog repeatedly forever (request permission -> cancel -> request permission -> request permission -> request permission -> ...)

eygraber:
> Using the lifecycle effect, I wonder if it is possible to infer whether the permission dialog was actually shown

I think the OS (at least in later versions) protects against this and always pauses the requesting `Activity` for some amount of time so it isn't possible to infer.

> This is even more of an edge case, but it would also be possible for the user to cancel the dialog repeatedly forever

That might be the point where a line has to be drawn between supporting that case and supporting a permanently denied state :thinking: 

I can see situations where the repeated cancelling could come up in the real world, but they'd still have the escape hatch of granting the permission in settings. Maybe it could be configurable, so the dev could choose between handling cancelled requests or a permanently denied status?

bentrengrove:
I am actually remembering now, this is why we didn't support this in the first place. Unfortunately, unless we can work this out I don't think we could land this in Accompanist, which I'm sorry for because I told you it would be ok. I think it is probably fine in your library but we can't have inconsistent behaviour in an "official" library.

Let me see if I can find an internal expert

eygraber:
I went over it a few times today, and I think with the public API there's no way around that.

If there's a hidden API or some way to tap in to the OS to get that info, or if there's an expert that knows of a non obvious way to detect the difference between the dialog getting canceled and being permanently denied, that would be great!

bentrengrove:
So the answer I got is that the activity result contract should be returning an `emptyMap` and you could check for that to determine if a dialog was shown and the request was cancelled. Do you want to try that and see if it helps?

eygraber:
Interesting, I'll take a look at that.

bentrengrove:
What happens if you just always use `RequestMultiplePermissions` even if it's a single permission?

eygraber:
Same result as single:

  `resultCode == Activity.RESULT_OK`
  `intent.getStringArrayExtras(EXTRA_PERMISSIONS)` -> `["android.permission..."]`
  `intent.getIntArrayExtra(EXTRA_PERMISSION_GRANT_RESULTS)` -> `[-1]`

bentrengrove:
How are you testing? Just in the accompanist sample app? I will give it a try

eygraber:
Yes, in the sample app.

eygraber:
I'm AFK for the next few days. I can do it when I'm back, but the changes were pretty minimal (mostly just revoked -> denied IIRC)

eygraber:
@bentrengrove updated the branch so that the sample builds and runs

bentrengrove:
Thanks for that, I gave it a go and confirmed the same thing. I can't find any way to detect cancellation for permissions. Even the multiple permission request always returns -1 for me, never emptyMap(). I will go back and see what I can find out

eygraber:
@bentrengrove any update here?

bentrengrove:
Not yet sorry, still trying to find a solution but it's not looking good unfortunately

eygraber:
> Not yet sorry, still trying to find a solution but it's not looking good unfortunately

> Unfortunately, unless we can work this out I don't think we could land this in Accompanist...we can't have inconsistent behaviour in an "official" library

I think part of the frustration here is that the whole permission system seems inconsistent to developers. My understanding of that is that it's for user safety as well as discouraging certain UX paradigms (e.g. redirecting to app settings), and I'm sure there are other considerations as well.

However it leaves us in a situation where the underlying system is perceived as broken (e.g. not being able to differentiate between a permission getting permanently denied and cancelling the permission request).

I know that following the [permission workflow] would make it easier to operate with these constraints, but the honest truth is that I've never worked with a product/design team that found that workflow acceptable. That led me to look for and implement workarounds, which all ultimately ended up not working in all cases. My experience is not unique, and there are countless QAs about this topic, and just as many broken workarounds.

@bentrengrove @alexvanyo I'm not directing any of this at you, and some of these issues go back almost a decade. I'm hoping that you can surface this issue internally, and that hopefully there can be some way to address it.

[permission workflow]: https://developer.android.com/training/permissions/requesting#workflow_for_requesting_permissions

sindrenm:
> However it leaves us in a situation where the underlying system is perceived as broken (e.g. not being able to differentiate between a permission getting permanently denied and cancelling the permission request).

Pardon me for shooting in here, but I have always believed this to be an intentional decision made by the platform team to “protect” users. That not knowing whether the _user directly_ denied a permission or whether the _system denied on behalf of the user_ is actually a feature.

Of course, as a developer, I find this incredibly frustrating, because I don't know when to send users to Settings to enable stuff there if the permission has been permanently denied, so I would personally love to have this feature built-in.

bentrengrove:
There is a long turn around time for things like this unfortunately. I will close this out for now as it is not looking likely we can land this, but when we get a response we can reopen

PR #1802, opened 2024-11-26
Upgrade to Kotlin 2.0.20 and move to build-logic for Gradle

bentrengrove:
I decided to clean up the Gradle files for Accompanist while upgrading to Kotlin 2.0


build-logic/convention/src/main/kotlin/AndroidLibraryConventionPlugin.kt
```diff
@@ -0,0 +1,54 @@
...
+                configureKotlinAndroid(this)
+                defaultConfig.targetSdk = 34
+                defaultConfig.testInstrumentationRunner = "androidx.test.runner.AndroidJUnitRunner"
+
+                buildFeatures.buildConfig = false
+
+                testOptions.animationsDisabled = true
+                // The resource prefix is derived from the module name,
+                // so resources inside ":core:module1" must be prefixed with "core_module1_"
+                resourcePrefix = path.split("""\W""".toRegex()).drop(1).distinct().joinToString(separator = "_").lowercase() + "_"
```
mlykotom, line 45: Takes me some time to understand what does it do 😄

mlykotom (approved):
LGTM, but I think would be better if someone else also takes a look

gradle/libs.versions.toml
```diff
@@ -1,28 +1,27 @@
...
 
 # gradlePlugin and lint need to be updated together
-gradlePlugin = "8.5.2"
+androidTools = "31.7.2"
+gradlePlugin = "8.7.2"
 lintMinCompose = "30.0.0"
 
 ktlint = "0.45.2"
-kotlin = "1.9.22"
+kotlin = "2.0.20"
```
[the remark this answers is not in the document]
alexvanyo: If the Kotlin version here transitively forces a newer version of Kotlin to be used by consumers of the library, keeping it lower here would be safer.

internal-testutils/src/main/java/com/google/accompanist/internal/test/Assertions.kt
```diff
@@ -42,7 +42,7 @@ fun ImageBitmap.assertPixels(expected: Color, tolerance: Float = 0.001f) {
  * Run the [SemanticsNodeInteraction] provided by [block] repeatedly until either
  * the assertion succeeds, or the execution runs past [timeoutMillis].
  */
-fun SemanticsNodeInteraction.assertWithTimeout(
+public fun SemanticsNodeInteraction.assertWithTimeout(
```
alexvanyo, line 45: The change here is just updating visibilities, but I wonder if this utilities can be replaced with upstream replacements now.

build-logic/convention/src/main/kotlin/com/google/accompanist/KotlinAndroid.kt
```diff
@@ -0,0 +1,91 @@
...
+
+/**
+ * Configure base Kotlin options for JVM (non-Android)
+ */
+internal fun Project.configureKotlinJvm() {
+    extensions.configure<JavaPluginExtension> {
+        // Up to Java 11 APIs are available through desugaring
+        // https://developer.android.com/studio/write/java11-minimal-support-table
+        sourceCompatibility = JavaVersion.VERSION_11
+        targetCompatibility = JavaVersion.VERSION_11
```
alexvanyo, line 65: I think I would expect these to also be 1.8, for any libraries that are being published to run on Android consumers?

build.gradle
```diff
@@ -18,31 +20,31 @@ buildscript {
...
-        classpath libs.gradleMavenPublishPlugin
-
-        classpath libs.metalavaGradle
-
         classpath libs.affectedmoduledetector
     }
 }
 
 plugins {
     id "com.diffplug.spotless" version "6.5.2"
```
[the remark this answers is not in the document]
alexvanyo: I'd prefer not having an init-script if we don't need it.

I think it was needed due to configuration cache issues at one point, but I find having to run a separate `init-script` fairly annoying and easy to forget about.

build.gradle
```diff
@@ -1,3 +1,5 @@
+import org.jetbrains.dokka.gradle.DokkaMultiModuleTask
```
alexvanyo, line 1: Nit: this import snuck in before the license as well

sample/build.gradle.kts
```diff
@@ -35,24 +39,24 @@ android {
     }
 
     compileOptions {
-        sourceCompatibility = JavaVersion.VERSION_1_8
-        targetCompatibility = JavaVersion.VERSION_1_8
+        sourceCompatibility = JavaVersion.VERSION_17
+        targetCompatibility = JavaVersion.VERSION_17
```
alexvanyo, line 43: Is this safe if we don't add desugaring?

permissions-lint/build.gradle.kts
```diff
@@ -83,6 +83,6 @@ dependencies {
 }
 
 java {
-    sourceCompatibility = JavaVersion.VERSION_1_8
-    targetCompatibility = JavaVersion.VERSION_1_8
+    sourceCompatibility = JavaVersion.VERSION_17
+    targetCompatibility = JavaVersion.VERSION_17
```
alexvanyo, line 87: If I understand correctly, this will require consuming projects to build with at least JDK 17, since `permissions-lint` is consumed at Gradle build time.

That SGTM, given that AGP has required that level for a while now.

build.gradle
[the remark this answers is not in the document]
bentrengrove: Planning to tackle the root level gradle file in a different PR

gradle/libs.versions.toml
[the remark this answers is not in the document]
bentrengrove: Yep, bumping one minor release at a time

build-logic/convention/src/main/kotlin/com/google/accompanist/KotlinAndroid.kt
bentrengrove: This one is for jvm targets so not Android, the Android target is pointing to 1.8

permissions-lint/build.gradle.kts
bentrengrove: Yeah that's how I understand it as well but I wasn't 100% sure if it was right. Changing it to 1.8 though fails to build

build-logic/convention/src/main/kotlin/com/google/accompanist/KotlinAndroid.kt
bentrengrove: It actually wasn't used so I just removed it

build-logic/convention/src/main/kotlin/AndroidLibraryConventionPlugin.kt
bentrengrove: Yeah agreed but I just copy pasted it

build-logic/convention/src/main/kotlin/com/google/accompanist/KotlinAndroid.kt
```diff
@@ -0,0 +1,91 @@
...
+    commonExtension.apply {
+        compileSdk = 34
+
+        defaultConfig {
+            minSdk = 21
+        }
+
+        compileOptions {
+            sourceCompatibility = JavaVersion.VERSION_1_8
+            targetCompatibility = JavaVersion.VERSION_1_8
```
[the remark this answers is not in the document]
bentrengrove: 17 for code that runs on the computer e.g. lint. 1.8 for Android libraries. 11 was unused so I deleted

PR #1803, opened 2024-12-02
[Permissions] Support calling rememberPermissionState in a Compose preview

eygraber:
Spinning this out while the fate of https://github.com/google/accompanist/pull/1793 hangs in the balance :sweat_smile:

### DIFF
--- a/permissions/src/main/java/com/google/accompanist/permissions/MultiplePermissionsState.kt
+++ b/permissions/src/main/java/com/google/accompanist/permissions/MultiplePermissionsState.kt
@@ -17,7 +17,10 @@
 package com.google.accompanist.permissions
 
 import androidx.compose.runtime.Composable
+import androidx.compose.runtime.Immutable
 import androidx.compose.runtime.Stable
+import androidx.compose.ui.platform.LocalInspectionMode
+import androidx.compose.ui.util.fastMap
 
 /**
  * Creates a [MultiplePermissionsState] that is remembered across compositions.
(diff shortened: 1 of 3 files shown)


permissions/src/main/java/com/google/accompanist/permissions/MultiplePermissionsState.kt
```diff
@@ -36,10 +36,12 @@ import androidx.compose.ui.util.fastMap
 @Composable
 public fun rememberMultiplePermissionsState(
     permissions: List<String>,
-    onPermissionsResult: (Map<String, Boolean>) -> Unit = {}
+    onPermissionsResult: (Map<String, Boolean>) -> Unit = {},
+    permissionStatuses: Map<String, PermissionStatus> = emptyMap()
```
eygraber, line 41: If this would prevent a release because of the API change I can move it to a separate PR. Not sure what the policy is with default params, but `metalavaGenerateSignatureRelease` seems to treat it as an API change (also see next commit where I address this with `JvmOverloads`).

bentrengrove: Yeah, default params don't fix binary compat unfortunately. Just add a new function and then make this one call your new one and then it is fine

eygraber: Does the `JvmOverloads` work for that, or does there have to be an actual new function in source?

bentrengrove: Good question, I'm not 100% sure actually but on Compose we always do it with a new function in source so I assume there is a reason

eygraber: Ok I'll push that shortly.

bentrengrove (approved):
Thank you for this!

PR #1807, opened 2024-12-10
Convert sample app activity to Compose

bentrengrove:
I will convert the actual samples to M3 in a followup PR

Screenshot_20241211_090901

### DIFF
--- a/sample/src/main/AndroidManifest.xml
+++ b/sample/src/main/AndroidManifest.xml
@@ -34,7 +34,7 @@
         android:label="@string/app_name"
         android:roundIcon="@mipmap/ic_launcher_round"
         android:supportsRtl="true"
-        android:theme="@android:style/Theme.Material.Light.NoActionBar">
+        android:theme="@android:style/Theme.Material.NoActionBar">
 
         <profileable android:shell="true"
             tools:targetApi="q" />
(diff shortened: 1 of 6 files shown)


sample/src/main/AndroidManifest.xml
```diff
@@ -42,7 +42,8 @@
         <activity
             android:name="com.google.accompanist.sample.MainActivity"
             android:label="@string/app_name"
-            android:theme="@android:style/Theme.Material.Light"
+            android:theme="@android:style/Theme.Material.Light.NoActionBar"
```
alexvanyo, line 45: Nit: If the sample supports dark mode and light mode, should this theme not be the `Light` version?

sample/src/main/java/com/google/accompanist/sample/MainScreen.kt
```diff
@@ -0,0 +1,97 @@
...
+            TopAppBar(
+                title = { Text(stringResource(R.string.app_name)) },
+                scrollBehavior = scrollBehavior
+            )
+
+            ContentList(
+                listData,
+                onItemClick,
+                modifier = Modifier
+                    .fillMaxSize()
```
alexvanyo, line 65: Edge-to-edge nit: there should be horizontal padding applied here for any left or right system bar insets.

alexvanyo (approved):

sample/src/main/AndroidManifest.xml
bentrengrove: Good catch!

PR #1817, opened 2025-04-21
[Drawable Painter] Fix AnimatedImageDrawable rendered in wrong size for systems below Android 12

landicefu:
AnimatedImageDrawable doesn't respect bounds set to it when the Android version is below 12.

On Android version below 12, if we use the current way to render (setBounds), the size doesn't change with the container at all. It's always following intrinsic size which is wrong.

I've created a sample project for you to test it out.
[issue-demo-animated-image-drawable.zip](https://github.com/user-attachments/files/19833717/issue-demo-animated-image-drawable.zip)

Also, you can see from following videos to understand the issue.

Tested on Android 10
<video src="https://github.com/user-attachments/assets/6ece7847-34be-4b04-9d2d-10be02a77fec" />

Tested on Android 12
<video src="https://github.com/user-attachments/assets/b7757265-08ba-49ec-89ec-d75087e6f89e" />

Following is the commit in AOSP to add setBounds to AnimatedImageDrawable.
https://cs.android.com/android/_/android/platform/frameworks/base/+/be969efdb0234b7b39636ec0a586dbb708ec480d

This first appeared in android-12.0.0_r1, and this is the proof.
image

### DIFF
--- a/drawablepainter/src/main/java/com/google/accompanist/drawablepainter/DrawablePainter.kt
+++ b/drawablepainter/src/main/java/com/google/accompanist/drawablepainter/DrawablePainter.kt
@@ -17,6 +17,7 @@
 package com.google.accompanist.drawablepainter
 
 import android.graphics.drawable.Animatable
+import android.graphics.drawable.AnimatedImageDrawable
 import android.graphics.drawable.ColorDrawable
 import android.graphics.drawable.Drawable
 import android.os.Build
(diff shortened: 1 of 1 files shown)


landicefu:
@bentrengrove Hello. We are affected by this problem pretty seriously because we use a lot of animated webp images in our app. Coil is using DrawablePaint for Drawables.
I wonder is there a change this can be merged and release soon so we can ask coil to do an update?

landicefu:
@bentrengrove How about adding condition to apply this only for AnimatedImageDrawable on systems below Android 12? This is definitely wrong rendering behavior and not intentional. The root cause is in the framework base, but that's not something we can change now.

bentrengrove:
OK. Yes, I would accept a PR that only applied this workaround as you describe. On Android <12 and for AnimatedImageDrawable. Would you be willing to make those changes?

landicefu:
No problem. I will make that change today. Thank you very much for considering this!!!

landicefu:
I've made the changes. Let me know if anything should be further adjusted. Thanks

bentrengrove (approved):

bentrengrove:
Looks like you just have spotless errors

`./gradlew :drawablepainter:spotlessApply`

bentrengrove (approved):

landicefu:
Okay. It's done.

bentrengrove:
This has been released as 0.37.3

landicefu:
thank you!! ❤️

Anaxdi30 (approved):

jgranger1 (commented):
Not shore what this is about not shore what I need to do I'm I supposed to do something

ubutatu (approved):

ubutatu (approved):

